PR #2, opened 2026-05-29
chore(ci): add standard Octo workflow suite

lml2468:
## Summary

Add the full standard GitHub Actions workflow set to match other Mininglamp-OSS repositories.

## New Workflows (15 files)

| Workflow | Purpose |
|---|---|
| auto-add-to-project | Auto-add issues/PRs to Octo project board |
| check-sprint | Enforce sprint assignment before merge |
| ci | Android build (ubuntu-latest, JDK 17 temurin, Gradle) |
| codeql | Java/Kotlin security analysis (reusable-codeql.yml) |
| issue-welcome | Welcome message for new issues |
| labeler | PR size labeling + dependency change detection |
| octo-ci-status | CI state-change notifications to Octo IM |
| octo-issue-feed | New issue notifications to Octo IM |
| octo-pr-result-notify | PR merge/close/review notifications |
| octo-pr-review-feed | PR review-request notifications |
| pr-contributor-welcome | First-time contributor welcome |
| release-drafter | Automated release note drafting |
| release-publish | Gated release publishing with CI validation |
| stale | Stale issue/PR management |
| workflow-sanity | YAML lint on workflow changes |

## Config

- `.github/dependabot.yml`: weekly updates for github-actions + gradle

## Notes

- Octo IM `project_group_id`: `84667e59ee9544e6b31a2fb6d3a21b23` (octo-android group)
- CI uses `gradle/actions/setup-gradle@v4` for Gradle caching

PR #1, opened 2026-05-29
fix: resolve compilation errors for fresh clone

mzanalyticst:
## Summary
- Fix invalid Android package name in default applicationId (`com.example-app.octo` → `com.example.app.octo`)
- Replace removed `PushKeys` class references with `BuildConfig` fields in `WKPushApplication`

## Problem
Fresh clone of the repo fails to compile due to:
1. Hyphen in default `applicationId` is not a valid Android package name
2. `PushKeys` class was removed during open-sourcing but references were not updated

## Test plan
- [ ] Clone repo without `local.properties` customization, verify `./gradlew assembleDevDebug` compiles
- [ ] Configure push keys in `local.properties`, verify push module compiles correctly

PR #11, opened 2026-05-30
fix: chat history file search and forward to contacts

mzanalyticst:
## Summary
- Add file search entry in chat history search (alongside members, photos, date)
- Add local DB search for images/videos/files (previously API-only, showed empty results)
- Fix forward messages to contacts: "create new chat" button now properly handles contact selection callback
- Rename "图片" to "图片与视频", add video support in image search
- Add file icon badge in keyword search results for file type messages
- Fix SQL IN clause bug in `searchWithChannelAndContentTypes`
- Override `WKFileContent.getSearchableWord()` to return `[文件]+filename` for global search

## Test plan
- [ ] Open chat → search chat history → verify "图片与视频", "文件", "群成员", "日期" all display in one row
- [ ] Click "图片与视频" → verify images and videos are listed
- [ ] Click "文件" → verify files are listed with icon, name, size, time
- [ ] Type filename in file search → verify filtering works
- [ ] Global search → type a filename → verify file result shows with `[文件]` prefix and file icon
- [ ] Forward message → click "创建新的聊天" → select a contact → confirm → verify message is forwarded

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -37,8 +37,8 @@ android {
         applicationId appId
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
-        versionCode 31
-        versionName "1.3.0"
+        versionCode 32
+        versionName "1.3.1"
 
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
         vectorDrawables {
--- a/wkbase/src/main/java/com/chat/base/msgcontent/WKFileContent.java
+++ b/wkbase/src/main/java/com/chat/base/msgcontent/WKFileContent.java
@@ -94,6 +94,12 @@ public class WKFileContent extends WKMediaMessageContent implements Parcelable {
     }
 
     @Override
+    public String getSearchableWord() {
+        String prefix = WKBaseApplication.getInstance().getContext().getString(R.string.last_message_file);
+        return name != null ? prefix + name : prefix;
+    }
+
+    @Override
     public String getDisplayContent() {
         return WKBaseApplication.getInstance().getContext().getString(R.string.last_message_file);
     }
--- a/wkim/src/main/java/com/xinbida/wukongim/db/MsgDbManager.java
+++ b/wkim/src/main/java/com/xinbida/wukongim/db/MsgDbManager.java
@@ -1190,22 +1190,20 @@ public class MsgDbManager {
         if (TextUtils.isEmpty(channelID) || contentTypes == null || contentTypes.length == 0) {
             return null;
         }
-        String whereStr = "";
-        for (int contentType : contentTypes) {
-            if (TextUtils.isEmpty(whereStr)) {
-                whereStr = String.valueOf(contentType);
-            } else {
-                whereStr = "," + contentType;
-            }
+        StringBuilder whereBuilder = new StringBuilder();
+        for (int i = 0; i < contentTypes.length; i++) {
+            if (i > 0) whereBuilder.append(",");
+            whereBuilder.append(contentTypes[i]);
         }
+        String inClause = whereBuilder.toString();
         Object[] args;
         String sql;
         if (oldestOrderSeq <= 0) {
-            args = new Object[]{channelID, channelType, whereStr};
-            sql = "select * from (select " + messageCols + "," + extraCols + " from " + message + " left join " + messageExtra + " on " + message + ".message_id= " + messageExtra + ".message_id where " + message + ".channel_id=? and " + message + ".channel_type=? and " + message + ".type<>0 and " + message + ".type<>99 and " + message + ".type in (?)) where is_deleted=0 and revoke=0 order by order_seq desc limit 0," + limit;
+            args = new Object[]{channelID, channelType};
+            sql = "select * from (select " + messageCols + "," + extraCols + " from " + message + " left join " + messageExtra + " on " + message + ".message_id= " + messageExtra + ".message_id where " + message + ".channel_id=? and " + message + ".channel_type=? and " + message + ".type<>0 and " + message + ".type<>99 and " + message + ".type in (" + inClause + ")) where is_deleted=0 and revoke=0 order by order_seq desc limit 0," + limit;
         } else {
-            args = new Object[]{channelID, channelType, oldestOrderSeq, whereStr};
-            sql = "select * from (select " + messageCols + "," + extraCols + " from " + message + " left join " + messageExtra + " on " + message + ".message_id= " + messageExtra + ".message_id where " + message + ".channel_id=? and " + message + ".channel_type=? and " + message + ".order_seq<? and " + message + ".type<>0 and " + message + ".type<>99 and " + message + ".type in (?)) where is_deleted=0 and revoke=0 order by order_seq desc limit 0," + limit;
+            args = new Object[]{channelID, channelType, oldestOrderSeq};
+            sql = "select * from (select " + messageCols + "," + extraCols + " from " + message + " left join " + messageExtra + " on " + message + ".message_id= " + messageExtra + ".message_id where " + message + ".channel_id=? and " + message + ".channel_type=? and " + message + ".order_seq<? and " + message + ".type<>0 and " + message + ".type<>99 and " + message + ".type in (" + inClause + ")) where is_deleted=0 and revoke=0 order by order_seq desc limit 0," + limit;
         }
         List<WKMsg> wkMsgs = new ArrayList<>();
         try (Cursor cursor = WKIMApplication
--- a/wkim/src/main/java/com/xinbida/wukongim/db/MsgDbManager.java
+++ b/wkim/src/main/java/com/xinbida/wukongim/db/MsgDbManager.java
@@ -1121,7 +1121,8 @@ public class MsgDbManager {
                 " m.client_seq else ''END client_seq, CASE count(*) WHEN 1 THEN m.searchable_word else '' end searchable_word," +
                 " CASE count(*) WHEN 1 THEN m.order_seq ELSE 0 END order_seq " +
                 "from " + channel + " c LEFT JOIN " + message + " m ON m.channel_id = c.channel_id and " +
-                "m.channel_type = c.channel_type WHERE m.is_deleted=0 and searchable_word LIKE ? GROUP BY " +
+                "m.channel_type = c.channel_type LEFT JOIN " + messageExtra + " e ON m.message_id = e.message_id " +
+                "WHERE m.is_deleted=0 and IFNULL(e.revoke,0)=0 and searchable_word LIKE ? GROUP BY " +
                 "c.channel_id, c.channel_type ORDER BY m.created_at DESC limit 100";
         Cursor cursor = WKIMApplication.getInstance().getDbHelper().rawQuery(sql, new Object[]{"%" + searchKey + "%"});
         if (cursor == null) {
--- a/wkuikit/src/main/AndroidManifest.xml
+++ b/wkuikit/src/main/AndroidManifest.xml
@@ -105,6 +105,7 @@
         <activity android:name="com.chat.uikit.chat.search.member.SearchWithMemberActivity" />
         <activity android:name="com.chat.uikit.chat.search.date.SearchWithDateActivity" />
         <activity android:name="com.chat.uikit.chat.search.image.SearchWithImgActivity" />
+        <activity android:name="com.chat.uikit.chat.search.file.SearchWithFileActivity" />
 
         <service
             android:name="com.chat.uikit.AliveJobService"
--- a/wkuikit/src/main/java/com/chat/uikit/WKUIKitApplication.java
+++ b/wkuikit/src/main/java/com/chat/uikit/WKUIKitApplication.java
@@ -699,6 +699,20 @@ public class WKUIKitApplication {
             return null;
         });
 
+        // 搜索消息按文件搜索
+        EndpointManager.getInstance().setMethod("search_message_with_file", EndpointCategory.wkSearchChatContent, 97, object -> {
+            if (object instanceof WKChannel) {
+                return new SearchChatContentMenu(WKBaseApplication.getInstance().getContext().getString(R.string.uikit_search_for_file), (channelID, channelType) -> {
+                    Intent intent = new Intent(WKBaseApplication.getInstance().getContext(), com.chat.uikit.chat.search.file.SearchWithFileActivity.class);
+                    intent.putExtra("channel_id", ((WKChannel) object).channelID);
+                    intent.putExtra("channel_type", ((WKChannel) object).channelType);
+                    intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+                    WKBaseApplication.getInstance().getContext().startActivity(intent);
+                });
+            }
+            return null;
+        });
+
     }
 
     public void sendChooseChatBack(List<WKChannel> list) {
@@ -1006,4 +1020,5 @@ public class WKUIKitApplication {
         sureTv.setTextColor(ContextCompat.getColor(context, R.color.colorAccent));
 
     }
+
 }
--- a/wkuikit/src/main/java/com/chat/uikit/chat/ChooseChatActivity.java
+++ b/wkuikit/src/main/java/com/chat/uikit/chat/ChooseChatActivity.java
@@ -31,6 +31,8 @@ import androidx.annotation.NonNull;
 
 import com.chat.base.base.WKBaseActivity;
 import com.chat.base.config.WKConfig;
+import com.chat.base.endpoint.EndpointManager;
+import com.chat.base.endpoint.entity.ChooseContactsMenu;
 import com.chat.base.msgitem.WKChannelMemberRole;
 import com.chat.base.ui.components.SegmentTabView;
 import com.chat.base.utils.SoftKeyboardUtils;
@@ -40,7 +42,6 @@ import com.chat.uikit.WKUIKitApplication;
 import com.chat.uikit.category.CategoryEntity;
 import com.chat.uikit.category.CategoryModel;
 import com.chat.uikit.chat.adapter.ChooseChatAdapter;
-import com.chat.uikit.contacts.ChooseContactsActivity;
 import com.chat.uikit.databinding.ActChooseChatLayoutBinding;
 import com.chat.uikit.message.MsgModel;
 import com.chat.uikit.thread.service.ThreadModel;
@@ -160,10 +161,23 @@ public class ChooseChatActivity extends WKBaseActivity<ActChooseChatLayoutBindin
         initAdapter(wkVBinding.recyclerView, chooseChatAdapter);
 
         wkVBinding.createTv.setOnClickListener(v -> {
-            Intent intent = new Intent(this, ChooseContactsActivity.class);
-            if (WKUIKitApplication.getInstance().getMessageContentList() != null)
-                intent.putParcelableArrayListExtra("msgContentList", (ArrayList<? extends Parcelable>) WKUIKitApplication.getInstance().getMessageContentList());
-            startActivity(intent);
+            ChooseContactsMenu contactsMenu = new ChooseContactsMenu(9, true, false, null, selectedList -> {
+                if (WKReader.isNotEmpty(selectedList)) {
+                    if (isChoose && WKUIKitApplication.getInstance().getMessageContentList() != null) {
+                        WKUIKitApplication.getInstance().showChatConfirmDialog(this, selectedList, WKUIKitApplication.getInstance().getMessageContentList(), new WKUIKitApplication.IShowChatConfirm() {
+                            @Override
+                            public void onBack(@NonNull List<WKChannel> list, @NonNull List<com.xinbida.wukongim.msgmodel.WKMessageContent> messageContentList) {
+                                WKUIKitApplication.getInstance().sendChooseChatBack(list);
+                                finish();
+                            }
+                        });
+                    } else {
+                        WKUIKitApplication.getInstance().sendChooseChatBack(selectedList);
+                        finish();
+                    }
+                }
+            });
+            EndpointManager.getInstance().invoke("choose_contacts", contactsMenu);
         });
 
         segmentTabView = new SegmentTabView(this,
--- a/wkuikit/src/main/java/com/chat/uikit/chat/search/SearchMessageAdapter.kt
+++ b/wkuikit/src/main/java/com/chat/uikit/chat/search/SearchMessageAdapter.kt
@@ -18,6 +18,8 @@ package com.chat.uikit.chat.search
 
 import android.os.Build
 import android.text.Html
+import android.view.View
+import android.widget.ImageView
 import android.widget.TextView
 import com.chad.library.adapter.base.BaseQuickAdapter
 import com.chad.library.adapter.base.viewholder.BaseViewHolder
@@ -26,6 +28,7 @@ import com.chat.base.msgitem.WKContentType
 import com.chat.base.ui.components.AvatarView
 import com.chat.base.utils.WKTimeUtils
 import com.chat.uikit.R
+import com.chat.uikit.chat.provider.WKFileProvider
 import com.chat.uikit.search.remote.GlobalAdapter
 
 class SearchMessageAdapter :
@@ -46,11 +49,22 @@ class SearchMessageAdapter :
             holder.setText(R.id.nameTv, Html.fromHtml(item.channel.getHtmlName()))
         }
         val contentTv = holder.getView<TextView>(R.id.contentTv)
+        val fileTagIv = holder.getView<ImageView>(R.id.fileTagIv)
+        val type = item.getContentType()
+
+        if (type == WKContentType.WK_FILE) {
+            fileTagIv.visibility = View.VISIBLE
+            val extension = item.payload["extension"] as? String ?: ""
+            val fileName = item.payload["name"] as? String ?: ""
+            WKFileProvider.setFileIcon(fileTagIv, extension, fileName)
+        } else {
+            fileTagIv.visibility = View.GONE
+        }
+
         val rawContent = item.payload["content"]
         if (rawContent is String && rawContent.isNotEmpty() && !rawContent.contains("<mark>")) {
             contentTv.text = GlobalAdapter.highlightKeyword(rawContent, keyword)
         } else {
-            val type = item.getContentType()
             if (type == WKContentType.WK_TEXT) {
                 contentTv.text = Html.fromHtml(item.getHtmlText(), Html.FROM_HTML_MODE_LEGACY)
             } else if (type == WKContentType.WK_FILE) {
--- a/wkuikit/src/main/java/com/chat/uikit/chat/search/file/SearchFileEntity.java
+++ b/wkuikit/src/main/java/com/chat/uikit/chat/search/file/SearchFileEntity.java
@@ -1,0 +1,21 @@
+package com.chat.uikit.chat.search.file;
+
+import com.chad.library.adapter.base.entity.MultiItemEntity;
+import com.chat.base.entity.GlobalMessage;
+
+public class SearchFileEntity implements MultiItemEntity {
+    public static final int TYPE_FILE = 0;
+    public static final int TYPE_DATE_HEADER = 1;
+
+    public int itemType;
+    public GlobalMessage message;
+    public String date;
+    public String fileName;
+    public String extension;
+    public long fileSize;
+
+    @Override
+    public int getItemType() {
+        return itemType;
+    }
+}
--- a/wkuikit/src/main/java/com/chat/uikit/chat/search/file/SearchFileEntity.java
+++ b/wkuikit/src/main/java/com/chat/uikit/chat/search/file/SearchFileEntity.java
@@ -2,6 +2,7 @@ package com.chat.uikit.chat.search.file;
 
 import com.chad.library.adapter.base.entity.MultiItemEntity;
 import com.chat.base.entity.GlobalMessage;
+import com.xinbida.wukongim.msgmodel.WKMessageContent;
 
 public class SearchFileEntity implements MultiItemEntity {
     public static final int TYPE_FILE = 0;
@@ -9,6 +10,7 @@ public class SearchFileEntity implements MultiItemEntity {
 
     public int itemType;
     public GlobalMessage message;
+    public WKMessageContent originalContent;
     public String date;
     public String fileName;
     public String extension;
--- a/wkuikit/src/main/java/com/chat/uikit/chat/search/file/SearchWithFileActivity.java
+++ b/wkuikit/src/main/java/com/chat/uikit/chat/search/file/SearchWithFileActivity.java
@@ -149,7 +149,9 @@ public class SearchWithFileActivity extends WKBaseActivity<ActSearchMsgFileLayou
     }
 
     private void forward(SearchFileEntity entity) {
-        WKMessageContent content = entity.message.getMessageModel();
+        WKMessageContent content = entity.originalContent != null
+                ? entity.originalContent
+                : entity.message.getMessageModel();
         if (content == null) return;
 
         EndpointManager.getInstance().invoke(EndpointSID.showChooseChatView,
@@ -192,6 +194,7 @@ public class SearchWithFileActivity extends WKBaseActivity<ActSearchMsgFileLayou
                         entity.fileName = fileContent.name != null ? fileContent.name : "";
                         entity.extension = fileContent.extension != null ? fileContent.extension : "";
                         entity.fileSize = fileContent.size;
+                        entity.originalContent = fileContent;
 
                         GlobalMessage gm = new GlobalMessage();
                         gm.setMessage_seq(msg.messageSeq);
--- a/wkuikit/src/main/java/com/chat/uikit/chat/search/file/SearchWithFileActivity.java
+++ b/wkuikit/src/main/java/com/chat/uikit/chat/search/file/SearchWithFileActivity.java
@@ -177,6 +177,7 @@ public class SearchWithFileActivity extends WKBaseActivity<ActSearchMsgFileLayou
                     .searchMsgWithChannelAndContentTypes(channelID, channelType, 0, 200, new int[]{WKContentType.WK_FILE});
             if (WKReader.isNotEmpty(localMsgs)) {
                 for (WKMsg msg : localMsgs) {
+                    if (com.chat.base.space.SpaceFilter.shouldSkipMessageForSpace(msg)) continue;
                     if (msg.baseContentMsgModel instanceof WKFileContent fileContent) {
                         // 按文件名过滤
                         if (!TextUtils.isEmpty(searchKeyword) && fileContent.name != null
@@ -276,14 +277,13 @@ public class SearchWithFileActivity extends WKBaseActivity<ActSearchMsgFileLayou
     }
 
     private void addDateHeaderIfNeeded(List<SearchFileEntity> list, String date) {
+        String lastDate = null;
         if (WKReader.isNotEmpty(list)) {
-            if (!list.get(list.size() - 1).date.equals(date)) {
-                SearchFileEntity header = new SearchFileEntity();
-                header.date = date;
-                header.itemType = SearchFileEntity.TYPE_DATE_HEADER;
-                list.add(header);
-            }
-        } else {
+            lastDate = list.get(list.size() - 1).date;
+        } else if (WKReader.isNotEmpty(adapter.getData())) {
+            lastDate = adapter.getData().get(adapter.getData().size() - 1).date;
+        }
+        if (!date.equals(lastDate)) {
             SearchFileEntity header = new SearchFileEntity();
             header.date = date;
             header.itemType = SearchFileEntity.TYPE_DATE_HEADER;
--- a/wkuikit/src/main/java/com/chat/uikit/chat/search/file/SearchWithFileActivity.java
+++ b/wkuikit/src/main/java/com/chat/uikit/chat/search/file/SearchWithFileActivity.java
@@ -132,6 +132,8 @@ public class SearchWithFileActivity extends WKBaseActivity<ActSearchMsgFileLayou
                     page = 1;
                     seenSeqs.clear();
                     adapter.setList(new ArrayList<>());
+                    wkVBinding.refreshLayout.setEnableLoadMore(true);
+                    wkVBinding.nodataTv.setVisibility(View.GONE);
                     getData();
                 }, 300);
             }
--- a/wkuikit/src/main/java/com/chat/uikit/chat/search/file/SearchWithFileActivity.java
+++ b/wkuikit/src/main/java/com/chat/uikit/chat/search/file/SearchWithFileActivity.java
@@ -51,6 +51,8 @@ public class SearchWithFileActivity extends WKBaseActivity<ActSearchMsgFileLayou
     private SearchWithFileAdapter adapter;
     private int page = 1;
     private String searchKeyword = "";
+    private final Set<Long> seenSeqs = new HashSet<>();
+    private final android.os.Handler searchHandler = new android.os.Handler(android.os.Looper.getMainLooper());
 
     @Override
     protected ActSearchMsgFileLayoutBinding getViewBinding() {
@@ -125,9 +127,13 @@ public class SearchWithFileActivity extends WKBaseActivity<ActSearchMsgFileLayou
             @Override
             public void afterTextChanged(Editable editable) {
                 searchKeyword = editable.toString().trim();
-                page = 1;
-                adapter.setList(new ArrayList<>());
-                getData();
+                searchHandler.removeCallbacksAndMessages(null);
+                searchHandler.postDelayed(() -> {
+                    page = 1;
+                    seenSeqs.clear();
+                    adapter.setList(new ArrayList<>());
+                    getData();
+                }, 300);
             }
         });
     }
@@ -163,7 +169,6 @@ public class SearchWithFileActivity extends WKBaseActivity<ActSearchMsgFileLayou
     private void getData() {
         // 本地搜索：按类型查出所有文件消息
         List<SearchFileEntity> localFileEntities = new ArrayList<>();
-        Set<Long> seenSeqs = new HashSet<>();
 
         if (page == 1) {
             List<WKMsg> localMsgs = WKIM.getInstance().getMsgManager()
--- a/wkuikit/src/main/java/com/chat/uikit/chat/search/file/SearchWithFileActivity.java
+++ b/wkuikit/src/main/java/com/chat/uikit/chat/search/file/SearchWithFileActivity.java
@@ -1,0 +1,285 @@
+package com.chat.uikit.chat.search.file;
+
+import android.text.Editable;
+import android.text.TextUtils;
+import android.text.TextWatcher;
+import android.view.View;
+import android.view.ViewGroup;
+import android.view.inputmethod.EditorInfo;
+import android.widget.TextView;
+
+import androidx.annotation.NonNull;
+import androidx.recyclerview.widget.LinearLayoutManager;
+
+import com.chat.base.base.WKBaseActivity;
+import com.chat.base.endpoint.EndpointManager;
+import com.chat.base.endpoint.EndpointSID;
+import com.chat.base.endpoint.entity.ChatChooseContacts;
+import com.chat.base.endpoint.entity.ChatViewMenu;
+import com.chat.base.endpoint.entity.ChooseChatMenu;
+import com.chat.base.entity.GlobalMessage;
+import com.chat.base.entity.GlobalSearchReq;
+import com.chat.base.msgitem.WKContentType;
+import com.chat.base.search.GlobalSearchModel;
+import com.chat.base.utils.SoftKeyboardUtils;
+import com.chat.base.utils.WKReader;
+import com.chat.base.utils.WKTimeUtils;
+import com.chat.base.views.pinnedsectionitemdecoration.PinnedHeaderItemDecoration;
+import com.chat.uikit.R;
+import com.chat.uikit.databinding.ActSearchMsgFileLayoutBinding;
+import com.google.android.material.snackbar.Snackbar;
+import com.scwang.smart.refresh.layout.api.RefreshLayout;
+import com.scwang.smart.refresh.layout.listener.OnRefreshLoadMoreListener;
+import com.xinbida.wukongim.WKIM;
+import com.xinbida.wukongim.entity.WKChannel;
+import com.xinbida.wukongim.entity.WKChannelType;
+import com.xinbida.wukongim.entity.WKMsg;
+import com.xinbida.wukongim.msgmodel.WKMessageContent;
+
+import com.chat.base.msgcontent.WKFileContent;
+import com.chat.base.net.HttpResponseCode;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public class SearchWithFileActivity extends WKBaseActivity<ActSearchMsgFileLayoutBinding> {
+    private String channelID;
+    private byte channelType;
+    private SearchWithFileAdapter adapter;
+    private int page = 1;
+    private String searchKeyword = "";
+
+    @Override
+    protected ActSearchMsgFileLayoutBinding getViewBinding() {
+        return ActSearchMsgFileLayoutBinding.inflate(getLayoutInflater());
+    }
+
+    @Override
+    protected void setTitle(TextView titleTv) {
+        titleTv.setText(R.string.uikit_search_for_file);
+    }
+
+    @Override
+    protected void initPresenter() {
+        channelID = getIntent().getStringExtra("channel_id");
+        channelType = getIntent().getByteExtra("channel_type", WKChannelType.PERSONAL);
+    }
+
+    @Override
+    protected void initView() {
+        PinnedHeaderItemDecoration headerDecoration = new PinnedHeaderItemDecoration.Builder(
+                SearchFileEntity.TYPE_DATE_HEADER).enableDivider(false).create();
+        wkVBinding.recyclerView.setLayoutManager(new LinearLayoutManager(this));
+        adapter = new SearchWithFileAdapter(new SearchWithFileAdapter.IClick() {
+            @Override
+            public void onClick(SearchFileEntity entity) {
+                showInChat(entity.message);
+            }
+
+            @Override
+            public void onForward(SearchFileEntity entity) {
+                forward(entity);
+            }
+        });
+        wkVBinding.recyclerView.setAdapter(adapter);
+        wkVBinding.recyclerView.addItemDecoration(headerDecoration);
+    }
+
+    @Override
+    protected void initListener() {
+        getData();
+
+        wkVBinding.refreshLayout.setEnableRefresh(false);
+        wkVBinding.refreshLayout.setOnRefreshLoadMoreListener(new OnRefreshLoadMoreListener() {
+            @Override
+            public void onLoadMore(@NonNull RefreshLayout refreshLayout) {
+                page++;
+                getData();
+            }
+
+            @Override
+            public void onRefresh(@NonNull RefreshLayout refreshLayout) {
+            }
+        });
+
+        wkVBinding.searchEt.setImeOptions(EditorInfo.IME_ACTION_SEARCH);
+        wkVBinding.searchEt.setOnEditorActionListener((v, actionId, event) -> {
+            if (actionId == EditorInfo.IME_ACTION_SEARCH) {
+                SoftKeyboardUtils.getInstance().hideSoftKeyboard(this);
+                return true;
+            }
+            return false;
+        });
+        wkVBinding.searchEt.addTextChangedListener(new TextWatcher() {
+            @Override
+            public void beforeTextChanged(CharSequence s, int start, int count, int after) {
+            }
+
+            @Override
+            public void onTextChanged(CharSequence s, int start, int before, int count) {
+            }
+
+            @Override
+            public void afterTextChanged(Editable editable) {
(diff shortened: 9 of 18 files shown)


lml2468 (changes requested):
# Code Review — PR #11 (octo-android)

## Summary

Adds chat history file search + forward-to-contacts, refactors image search to include video support, and fixes the SQL `IN` clause bug in `MsgDbManager`. 16 files, +736 −74. The feature implementation is solid overall, but has a pagination dedup bug that will produce visible duplicates.

## CI Status

`build` and `CodeQL` are still pending. `check-sprint` failed (process check — no sprint linked). Review proceeds on code; **verify build passes before merging**.

## Findings

### 🔴 P1 — Cross-page dedup produces duplicate entries

**Files:** `SearchWithFileActivity.java:181`, `SearchWithImgActivity.java:252`

`seenSeqs` is a local variable inside `getData()`. On page 1 it correctly deduplicates local DB results against API results. On page 2+, a **fresh empty** `seenSeqs` is created, so API results that overlap with the 200 local messages loaded on page 1 will appear twice in the list.

**Fix:** Promote `seenSeqs` to an instance field. Clear it when `page` resets to 1 (in `afterTextChanged` / `initListener`).

```java
// Instance field
private final Set<Long> seenSeqs = new HashSet<>();

// In afterTextChanged / page reset
seenSeqs.clear();
page = 1;
adapter.setList(new ArrayList<>());
getData();
```

### 🟡 P2 — No debounce on search input

**File:** `SearchWithFileActivity.java:153` (`afterTextChanged`)

Every keystroke fires `getData()` which makes an API call. Fast typing produces a burst of requests, most of which are wasted. Consider a `Handler.postDelayed` debounce (200–300ms) — cancel pending request on each new keystroke, only fire after typing pauses.

### 🟡 P2 — `getSearchableWord()` prepends localized prefix

**File:** `WKFileContent.java:97`

```java
return name != null ? prefix + name : prefix;
```

Returns e.g. `[文件]report.pdf`. If global search matches against this string, users searching for `report` would match, but searching for just the extension or browsing behavior might be affected. Verify this matches the global search contract for `searchableWord`.

## Verification

| Item | Result |
|---|---|
| SQL `IN` clause fix (`MsgDbManager.java`) | ✅ Old code passed comma-separated string as single `?` — never matched multiple types. New code correctly inlines validated int values. |
| `@drawable/ic_file_document` exists | ✅ `wkuikit/src/main/res/drawable/ic_file_document.xml` |
| `WKFileProvider.setFileIcon` / `formatFileSize` exist | ✅ `WKFileProvider.kt` present |
| `item_search_msg_img_date_layout` (reused for date headers) | ✅ exists |
| `choose_contacts` endpoint pattern | ✅ `ChooseContactsMenu` from base endpoint entity |
| AndroidManifest activity registration | ✅ `SearchWithFileActivity` registered |
| Video support in image search | ✅ Handles `WKVideoContent` cover images correctly |
| String resources (en + zh) | ✅ Both added for `uikit_search_for_file`, image renamed to "Photos & Videos" |

## Highlights

- **SQL fix is a real bug fix** — the old `IN (?)` with a comma-separated string parameter never worked for multiple content types. Good catch.
- **Video support in SearchWithImgActivity** is well-implemented — local path fallback → cover URL, consistent with image handling.
- **Forward-to-contacts** correctly uses the endpoint-based `ChooseContactsMenu` pattern rather than direct Activity launch — more extensible.
- Layout files are clean and follow existing conventions.

## Verdict

The P1 dedup bug will cause visible duplicate entries on scroll (load-more). Requesting changes for that fix. The P2 items are advisory.

Jerry-Xin (changes requested):
The PR is relevant to this repository, but it introduces a blocking regression in the new local media/file search result actions.

🔴 Blocking

🔴 Critical: Local search results cannot be forwarded correctly because their `GlobalMessage` payloads are incomplete. In [SearchWithFileActivity.java](/Users/mlamp/.openclaw/tmp/pr-review-oitu0ix9/repo/wkuikit/src/main/java/com/chat/uikit/chat/search/file/SearchWithFileActivity.java:200), locally found file messages only set `type`, `name`, `extension`, and `size`, omitting `url` and `localPath`. In [SearchWithImgActivity.java](/Users/mlamp/.openclaw/tmp/pr-review-oitu0ix9/repo/wkuikit/src/main/java/com/chat/uikit/chat/search/image/SearchWithImgActivity.java:299), locally found image/video messages only set `type`, omitting the actual media fields. Both forward paths then call `entity.message.getMessageModel()` and send that decoded content in [SearchWithFileActivity.java](/Users/mlamp/.openclaw/tmp/pr-review-oitu0ix9/repo/wkuikit/src/main/java/com/chat/uikit/chat/search/file/SearchWithFileActivity.java:146) and [SearchWithImgActivity.java](/Users/mlamp/.openclaw/tmp/pr-review-oitu0ix9/repo/wkuikit/src/main/java/com/chat/uikit/chat/search/image/SearchWithImgActivity.java:233). Since `GlobalMessage.getMessageModel()` decodes only from `payload`, local results forward blank/broken file, image, or video messages. Populate the full encoded payload from the original `WKMsg.baseContentMsgModel.encodeMsg()` plus `type`, or carry the original `WKMessageContent` separately for forwarding.

💬 Non-blocking

🟡 Warning: Pagination can now duplicate date headers across pages. The previous image search code removed the first new header when it matched the current last adapter date, but the new merge path in [SearchWithImgActivity.java](/Users/mlamp/.openclaw/tmp/pr-review-oitu0ix9/repo/wkuikit/src/main/java/com/chat/uikit/chat/search/image/SearchWithImgActivity.java:365) and the new file search path in [SearchWithFileActivity.java](/Users/mlamp/.openclaw/tmp/pr-review-oitu0ix9/repo/wkuikit/src/main/java/com/chat/uikit/chat/search/file/SearchWithFileActivity.java:258) append page data without cross-page header coalescing.

🟡 Warning: The new file search fires a network search on every text change without cancellation or a request token in [SearchWithFileActivity.java](/Users/mlamp/.openclaw/tmp/pr-review-oitu0ix9/repo/wkuikit/src/main/java/com/chat/uikit/chat/search/file/SearchWithFileActivity.java:126). Slower callbacks from older keywords can overwrite newer results.

✅ Highlights

The SQL `IN` clause bug is fixed for multi-type searches, and adding video to the image search entry matches the renamed “Photos & Videos” behavior.

I could not run the Gradle compile locally because this environment has no Java runtime installed.

lml2468 (changes requested):
# Re-review — PR #11 (octo-android) — `064275cf0b50`

## Previous findings status

| Issue | Status |
|---|---|
| 🔴 P1: Cross-page dedup bug (`seenSeqs` local variable) | ✅ Fixed — promoted to instance field, cleared on page reset |
| 🟡 P2: No debounce on search input | ✅ Fixed — 300ms `Handler.postDelayed` debounce added |
| 🟡 P2: `getSearchableWord()` prefix | ⚪ Unchanged (advisory, non-blocking) |

## Still blocking

### 🔴 P0 — Local search results forward with incomplete payload (unfixed)

**Files:** `SearchWithFileActivity.java:205-210`, `SearchWithImgActivity.java` (same pattern)

Local search constructs `GlobalMessage` with minimal payload:
```java
payloadMap.put("type", msg.type);
payloadMap.put("name", entity.fileName);
payloadMap.put("extension", entity.extension);
payloadMap.put("size", entity.fileSize);
// Missing: url, path — required for file download
```

When user forwards a locally-found file, `forward()` calls `entity.message.getMessageModel()` which deserializes from this payload. The resulting `WKFileContent` has `url = null` — the recipient gets a file message with no download link.

**Fix options (pick one):**

Option A — Include `url` in the payload map:
```java
payloadMap.put("url", fileContent.url);
// For images/videos in SearchWithImgActivity, also include image/video URLs
```

Option B — Store the original content model directly and use it for forwarding instead of reconstructing from payload:
```java
// In SearchFileEntity, add:
public WKMessageContent originalContent;

// In local search:
entity.originalContent = fileContent;

// In forward():
WKMessageContent content = entity.originalContent != null
    ? entity.originalContent
    : entity.message.getMessageModel();
```

Same issue applies to `SearchWithImgActivity` — local image/video results only have `payloadMap.put("type", msg.type)`, missing `url`/`width`/`height`/`cover`.

## Verdict

Dedup and debounce fixes look good. But the payload incompleteness is a P0 that will cause broken forwards for any locally-sourced result. CHANGES_REQUESTED.

lml2468 (dismissed):
# Re-review — PR #11 (octo-android) — `e096a2873fab`

## All previous blocking issues resolved

| Round | Issue | Status |
|---|---|---|
| R1 | 🔴 P1: Cross-page dedup (`seenSeqs` local var) | ✅ Fixed — instance field, cleared on page reset |
| R1 | 🟡 P2: No debounce on search input | ✅ Fixed — 300ms `Handler.postDelayed` |
| R2 | 🔴 P0: Local forward with incomplete payload | ✅ Fixed — `originalContent` field stores original `WKMessageContent`, used preferentially in `forward()` |

## Verification of latest fix

Both `SearchFileEntity` and `SearchImgEntity` now carry `public WKMessageContent originalContent`. Local search populates it (`fileContent` / `msg.baseContentMsgModel`). `forward()` in both activities checks `originalContent != null` first, falling back to `getMessageModel()` for API-sourced results. Clean solution.

## CI Note

`build` and `CodeQL` still pending. `check-sprint` failed (process check — no sprint linked, not a code issue). Verify build passes before merging.

## Verdict

APPROVED. All blocking findings from rounds 1 and 2 are addressed cleanly.

yujiawei (dismissed):
# Code Review — PR #11 (octo-android)

## Summary

Adds chat-history **file search** + **forward-to-contacts**, extends image search to include **video**, and fixes the SQL `IN`-clause bug in `MsgDbManager`. 17 files, +753 −75. The implementation is solid and all blocking issues from prior review rounds (cross-page dedup, forward-of-local-results regression) are resolved at this head SHA. Approving with a couple of non-blocking suggestions.

## Verification

| Item | Status | Evidence |
|---|---|---|
| SQL `IN`-clause fix | ✅ Correct | `MsgDbManager.searchWithChannelAndContentTypes` now inlines the type list into the SQL string (`type in (` + inClause + `)`) and drops the bogus `whereStr` bind arg, so the `args[]` count matches the `?` placeholders. The old code bound the comma-joined string to a single `?`, which only ever matched the first type. Values are framework-generated `int` content types — no injection surface. |
| Forward of **local** search results | ✅ Fixed | The earlier blocking regression (local results forwarding blank/broken messages because `GlobalMessage.getMessageModel()` decodes only from `payload`) is resolved: `SearchFileEntity`/`SearchImgEntity` now carry `originalContent`, and both `forward()` paths prefer it (`SearchWithFileActivity:146-149`, `SearchWithImgActivity:233-236`). |
| Cross-page dedup | ✅ Correct | `seenSeqs` is an instance `Set<Long>` cleared on page reset; local `int messageSeq` is widened to `Long` consistently before `.add()`, so local/API dedup keys are comparable. |
| Search debounce | ✅ Present | 300 ms `Handler.postDelayed` with `removeCallbacksAndMessages(null)` on each keystroke. |
| Build / symbol resolution | ✅ OK | All referenced symbols exist: `WKContentType.WK_FILE/WK_IMAGE/WK_VIDEO`, `WKFileProvider.setFileIcon/formatFileSize` (`@JvmStatic`), `WKFileContent.{name,extension,size}`, layouts (`act_search_msg_file_layout`, `item_search_msg_file_layout`, `item_search_msg_img_date_layout`), view IDs, and `R.string.uikit_search_for_file`. The new `SearchWithFileActivity` is registered in `AndroidManifest.xml`. Pattern-matching `instanceof` is fine under the module's Java 17 source level. |
| `ChooseChatActivity` "create new chat" refactor | ✅ Sound | Now routes through the `choose_contacts` endpoint with a `ChooseContactsMenu` callback; the callback honors the `isChoose` + `messageContentList` confirm-dialog path, matching the existing `onChooseBack` semantics. |

### 🟡 P2 — Date headers can duplicate at pagination boundaries

In both `SearchWithFileActivity.getData()` and `SearchWithImgActivity.getData()`, on `page > 1` the merge list starts empty (`merged = new ArrayList<>(localEntities)`, and `localEntities` is only populated when `page == 1`). `addDateHeaderIfNeeded()` only inspects the last element of this fresh local list — never the items already rendered in the adapter — so the first item of every subsequent page always gets a new date header even when the previous page ended with the same year-month. Result: a redundant `YYYY-MM` header at each page boundary that shares a month.

The pre-PR image code coalesced this by removing the first new header when it matched the adapter's current last date; that coalescing was dropped in the refactor. Purely cosmetic — results render correctly, no data loss or crash. Suggest tracking a `lastRenderedDate` field (or re-adding the adapter-tail comparison) so headers coalesce across pages.

### 🔵 Nit — `originalContent` not set for API results (symmetry)

Local results set `entity.originalContent`, but API-path entities in both activities don't, even though `getMessageModel()` is already called for the null-guard. The `forward()` fallback (`originalContent != null ? ... : message.getMessageModel()`) handles this correctly and `GlobalMessage.getMessageModel()` memoizes its result, so there's no functional or perf cost — just an inconsistency worth tidying for readability.

### 🔵 Nit — minor whitespace-only churn

`WKUIKitApplication.java` gains a stray trailing blank line at the class end (`+1` near line 1020). Harmless.

## Verdict

No correctness, security, data-loss, or build-break issues. The functional blockers from earlier rounds are fixed. The remaining items are cosmetic/maintainability and do not gate merge.

Jerry-Xin (changes requested):
The PR is in scope, but the new local media/file search path bypasses the repository’s Space filtering contract and can surface messages from another Space.

**🔴 Blocking**
- 🔴 Critical: Local file/image search is merged ahead of server search without applying the current Space filter. `SearchWithFileActivity` pulls local file messages by channel only at `wkuikit/src/main/java/com/chat/uikit/chat/search/file/SearchWithFileActivity.java:175-177`, and `SearchWithImgActivity` does the same for images/videos at `wkuikit/src/main/java/com/chat/uikit/chat/search/image/SearchWithImgActivity.java:258-262`. The server search path explicitly passes `current_space_id` in `wkbase/src/main/java/com/chat/base/search/GlobalSearchModel.kt:45-47`, and the local filtering contract exists in `wkbase/src/main/java/com/chat/base/space/SpaceFilter.java:493-503`. As written, in Space mode the local results can expose media/files from another Space before the correctly filtered API results are merged. Apply the same Space filtering before adding local messages, or do not merge local channel-only results when a Space is active.

**💬 Non-blocking**
- 🟡 Warning: The chat history type grid is still configured for 3 columns in `wkuikit/src/main/java/com/chat/uikit/chat/search/MessageRecordActivity.kt:94-97`, so adding the fourth “File” entry cannot display all four entries in one row as described in the test plan. The padding change in `item_search_message_type_layout.xml` does not change the grid span count.
- 🟡 Warning: Pagination can add duplicate date headers. On page > 1, `merged` starts empty, `addDateHeaderIfNeeded` always creates a header for the first item, and `adapter.addData(...)` appends it without comparing against the adapter’s last existing date. This affects files at `SearchWithFileActivity.java:244-272` and images/videos at `SearchWithImgActivity.java:351-374`.

**✅ Highlights**
- The SQL `IN` clause construction fixes the previous single-placeholder bug for multiple content types.
- Adding `WKFileContent.getSearchableWord()` improves local/global filename search behavior.

I could not run the targeted Gradle compile because this environment has no Java Runtime installed.

lml2468 (dismissed):
# Re-review — PR #11 (octo-android) — `eed8a69c155f`

## All blocking issues resolved across 4 rounds

| Round | Issue | Status |
|---|---|---|
| R1 | 🔴 P1: Cross-page dedup (`seenSeqs` local var) | ✅ Fixed — instance field |
| R1 | 🟡 P2: No debounce on search input | ✅ Fixed — 300ms debounce |
| R2 | 🔴 P0: Local forward with incomplete payload | ✅ Fixed — `originalContent` field |
| R3 | 🔴 P0: Space isolation leak (Allen/Jerry) | ✅ Fixed — `SpaceFilter.shouldSkipMessageForSpace(msg)` |
| R3 | 🟡 P2: Cross-page date header duplication | ✅ Fixed — checks `adapter.getData()` last date |

## New changes in this commit

1. **Space filter** — Both `SearchWithFileActivity.java:180` and `SearchWithImgActivity.java:265` call `SpaceFilter.shouldSkipMessageForSpace(msg)` at the start of the local message loop. Clean placement — skips before any processing.

2. **Date header dedup** — `addDateHeaderIfNeeded()` now checks the adapter's existing data when the current list is empty (page 2+ scenario), preventing duplicate date headers across pages.

3. **Revoked message exclusion** — `MsgDbManager.java:1124-1125` adds `LEFT JOIN messageExtra` with `IFNULL(e.revoke,0)=0` filter. Prevents revoked messages from appearing in search results. `LEFT JOIN` + `IFNULL` handles messages without an extra record correctly (no extra = not revoked).

4. **Version bump** — 1.3.0 → 1.3.1, versionCode 31 → 32.

## Verdict

APPROVED. Four rounds of review, all blocking findings addressed.

Jerry-Xin (changes requested):
The PR is in scope for `octo-android`, but the new local media/file search has a blocking correctness issue.

**🔴 Blocking**
- 🔴 Critical: Local file/media search only checks the newest 200 local messages, then silently drops older local results. In [SearchWithFileActivity.java](/Users/mlamp/.openclaw/tmp/pr-review-ccv9xo8p/repo/wkuikit/src/main/java/com/chat/uikit/chat/search/file/SearchWithFileActivity.java:175), local DB search always calls `searchMsgWithChannelAndContentTypes(..., 0, 200, ...)` only on `page == 1`, and filename filtering is applied afterward in memory at line 183. A file matching the typed filename but older than the newest 200 file messages will not appear unless the remote API also returns it, which defeats the PR’s local-search goal. The same pattern exists for photos/videos in [SearchWithImgActivity.java](/Users/mlamp/.openclaw/tmp/pr-review-ccv9xo8p/repo/wkuikit/src/main/java/com/chat/uikit/chat/search/image/SearchWithImgActivity.java:258): load-more pages only the API, not local DB results. Fix by paginating local results with `oldestOrderSeq`, or by moving filename filtering into the DB query and continuing local pagination until exhausted.

**💬 Non-blocking**
- 🟡 Warning: File search can show stale async results. [SearchWithFileActivity.java](/Users/mlamp/.openclaw/tmp/pr-review-ccv9xo8p/repo/wkuikit/src/main/java/com/chat/uikit/chat/search/file/SearchWithFileActivity.java:128) debounces input, but the callback at line 228 does not verify that the response still belongs to the current keyword/page. A slow response for an older keyword can overwrite or append into the current query. Use a request token or captured keyword/page guard.
- 🟡 Warning: After a first-page empty file search, load-more is disabled at [SearchWithFileActivity.java](/Users/mlamp/.openclaw/tmp/pr-review-ccv9xo8p/repo/wkuikit/src/main/java/com/chat/uikit/chat/search/file/SearchWithFileActivity.java:263) and is never re-enabled when the keyword changes and results return. This can prevent paging later searches.
- 🟡 Warning: [SearchWithImgActivity.java](/Users/mlamp/.openclaw/tmp/pr-review-ccv9xo8p/repo/wkuikit/src/main/java/com/chat/uikit/chat/search/image/SearchWithImgActivity.java:208) still opens results with `WKChannelType.GROUP` instead of the actual `channelType`, so image/video search results from personal chats can navigate to the wrong chat. This appears pre-existing, but the PR expands this screen’s scope to videos and local results.

**✅ Highlights**
- The SQL `IN` clause construction is fixed from the previous single-placeholder bug.
- File global-search display now includes a file icon and searchable file prefix.
- The forward-to-contact flow now routes through the existing contact-selection callback path.

I could not run compilation locally because this environment has no Java runtime installed.

Jerry-Xin (changes requested):
This PR is relevant to `octo-android`, but two functional issues block merge.

**🔴 Blocking**

🔴 Critical: File search only filters the newest 200 local file messages, so filename search can silently miss older local files. In [SearchWithFileActivity.java](/Users/mlamp/.openclaw/tmp/pr-review-2u2gqp7k/repo/wkuikit/src/main/java/com/chat/uikit/chat/search/file/SearchWithFileActivity.java:175), local search runs only when `page == 1` with `limit = 200`, then filters filenames in memory at lines 183-185. Later pages only call the remote API. Since this PR specifically fixes API-only empty results, a local-only chat with more than 200 file messages will fail the “type filename and verify filtering works” test for older files. The local DB query needs real pagination and/or a filename/searchable-word predicate instead of a fixed first-page scan.

🔴 Critical: The chat history search type row still uses a 3-column grid while the PR adds a fourth entry. [MessageRecordActivity.kt](/Users/mlamp/.openclaw/tmp/pr-review-2u2gqp7k/repo/wkuikit/src/main/java/com/chat/uikit/chat/search/MessageRecordActivity.kt:95) creates `FullyGridLayoutManager(this, 3)`, but [WKUIKitApplication.java](/Users/mlamp/.openclaw/tmp/pr-review-2u2gqp7k/repo/wkuikit/src/main/java/com/chat/uikit/WKUIKitApplication.java:657) now registers member, image/video, file, and date entries. The fourth item will wrap instead of displaying “all in one row” as required by the PR test plan.

**💬 Non-blocking**

🟡 Warning: `SearchWithFileActivity` disables load-more on an empty first page at [SearchWithFileActivity.java](/Users/mlamp/.openclaw/tmp/pr-review-2u2gqp7k/repo/wkuikit/src/main/java/com/chat/uikit/chat/search/file/SearchWithFileActivity.java:263), but new searches do not re-enable it. After a no-result filename search, a later search with results can be stuck without pagination.

🟡 Warning: I could not run compile verification because this environment has no Java runtime installed: `Unable to locate a Java Runtime`.

**✅ Highlights**

The SQL `IN` clause bug is addressed, revoked messages are filtered from global search, and file messages now have searchable text and a visible file badge in global results.

yujiawei (approved):
# Code Review — PR #11 (octo-android)

## Scope of this review

Focused verification of the forward path for chat-history file/image search results, per the question: *can `forward()` ever send a broken (urlless) message?* I traced both `forward()` implementations, the entity construction in `getData()`, and `GlobalMessage.getMessageModel()`.

Files reviewed:
- `wkuikit/.../chat/search/file/SearchWithFileActivity.java`
- `wkuikit/.../chat/search/image/SearchWithImgActivity.java`
- `wkbase/.../entity/GlobalMessage.kt`
- `wkbase/.../msgcontent/WKFileContent.java`, `wkim/.../msgmodel/WKImageContent.java`, `WKVideoContent.java`
- `wkim/.../manager/MsgManager.java` (`getMsgContentModel`)

## 1. Verification results

Both `forward()` methods select content the same way:

```java
WKMessageContent content = entity.originalContent != null
        ? entity.originalContent
        : entity.message.getMessageModel();
if (content == null) return;
```

### (1) Local results — originalContent always non-null? ✅

- **File** (`SearchWithFileActivity.getData()`): `entity.originalContent = fileContent;` is assigned unconditionally inside the `if (msg.baseContentMsgModel instanceof WKFileContent fileContent)` block (line ~198). The entity is only created inside that block, so `originalContent` is guaranteed non-null. `fileContent` is the real `WKMessageContent` from local DB, carrying `url` and `localPath`.
- **Image** (`SearchWithImgActivity.getData()`): `entity.originalContent = msg.baseContentMsgModel;` (line ~293), guarded earlier by `if (msg.baseContentMsgModel == null) continue;` (line 266). Non-null guaranteed.
- **Dedup ordering protects the local copy**: local entries call `seenSeqs.add(...)` (file line 187, image line 267) *before* the API callback runs; the API loop skips any seq already seen (file line 236, image line 327). So for any message present in the local DB, forward uses the local `originalContent` (with url/path), never the minimal reconstructed payload.

Conclusion: for local results, forward sends a complete, downloadable content. No broken-message path.

### (2) API results — does getMessageModel() carry url/path? ⚠️ (depends on server payload; image path is safe, file path is unguarded)

For API results `originalContent` is null, so `entity.message.getMessageModel()` is used. `GlobalMessage.getMessageModel()` does:

```kotlin
val jsonObject = JSONObject(payload)
messageContent = WKIM.getInstance().msgManager.getMsgContentModel(jsonObject)
```

`getMsgContentModel` → the content type's `decodeMsg`, which reads url/path from the payload JSON:
- `WKFileContent.decodeMsg`: `this.url = jsonObject.optString("url"); this.localPath = jsonObject.optString("localPath");`
- `WKImageContent.decodeMsg` / `WKVideoContent.decodeMsg`: same pattern for `url`/`localPath`/`cover`.

So the forwarded content's url/path is exactly whatever keys the **API search-response payload** contains. `optString("url")` returns `""` (empty) when the key is absent — it does NOT null the content, so the `if (content == null) return;` guard does **not** catch a urlless payload.

- **Image/Video API path: safe by construction.** Before adding an API entity, `getData()` computes `showUrl` from `imgModel.url` / `videoModel.cover` and does `if (TextUtils.isEmpty(showUrl)) continue;` (image lines 333–350). Any API image/video that reaches the list therefore had a non-empty url/cover, so its `getMessageModel()` content forwards with a real url. No broken forward for displayed images.

- **File API path: NOT guarded.** In `SearchWithFileActivity.getData()` the API branch only reads `name`/`extension`/`size` from payload (lines 252–255) and never inspects `url`. An API file result whose payload lacks `url` would still be added, displayed, and — when forwarded — produce a `WKFileContent` with `url == ""`. That is a forwardable urlless (broken / non-downloadable) message.

## 2. Findings

### P2 — File forward has no url guard on the API path (defensive gap, not a confirmed live bug)

In practice the WuKongIM file message payload persisted server-side normally includes `url` (set after upload completes), so the common case is fine. But the rest of this codebase consistently treats `fileContent.url` as possibly empty and guards it — e.g. `WKFileProvider.kt` (`if (TextUtils.isEmpty(fileContent.url))` at lines 79 and 181) and `ChatMultiForwardDetailAdapter.kt` (line 508). The new file forward path is the one place that does not. If the search API ever returns a file row without `url` (revoked/expired upload, partial server payload, schema drift), this path silently forwards an undownloadable file, whereas the image path already self-protects.

Suggested hardening (file activity `forward()`), mirroring the existing convention:

```java
if (content instanceof WKFileContent fc && TextUtils.isEmpty(fc.url)
        && TextUtils.isEmpty(fc.localPath)) {
    // toast "file unavailable" and return, instead of forwarding a urlless message
    return;
}
```

Equivalently the guard could live in the API branch of `getData()` (skip the entity when payload has no `url`), matching how the image branch filters on empty `showUrl`.

This is P2 (not a merge blocker): it relies on a server returning a malformed/url-less file payload, which is not the normal contract, and the local-DB path (the typical forward source) is fully correct.

## 3. Recommendations

- Add the empty-url guard above to bring the file forward path in line with `WKFileProvider` / `ChatMultiForwardDetailAdapter`, which already assume url may be empty.
- Optional consistency: the image activity validates url before display; consider applying the same "skip if no url" filter to API file results so the list never shows an item that cannot be opened or forwarded.

## 4. Additional notes

- `getMessageModel()` caches into `messageContent` and never returns null for a known content type (`getMsgContentModel` falls back to `new WKMessageContent()` when the type is unknown), so the `content == null` check guards only against an unparseable/unknown type — it cannot be relied on to catch a missing url. The real safety for images comes from the `showUrl` filter, not the null check.
- Dedup correctly prefers the local copy over the API copy for the same seq, which is the desired behavior for a complete forward.

## Verdict

No correctness/data-loss/build-break (P0) and no functional blocker (P1) in the forward paths: local results always forward complete content, and the image/video API path is self-protected by the `showUrl` filter. The single gap is a P2 defensive hardening on the file API forward path. Approving with the P2 suggestion above.

yujiawei (changes requested):
# Code Review — PR #11 (octo-android)

## Summary

Adds chat-history **file search** + **forward-to-contacts**, extends image search to cover **video**, fixes the SQL `IN`-clause bug in `MsgDbManager`, and filters revoked messages from global search. 18 files, +757 −78.

Several blockers raised in earlier rounds are genuinely fixed at this head SHA (`eed8a69`): the forward-of-local-results regression (now carries `originalContent`), Space filtering on the local path (`SpaceFilter.shouldSkipMessageForSpace`), cross-page date-header coalescing, and the revoke filter. However, **two functional blockers remain unresolved**, so this PR is not ready to merge.

## Verification

| Item | Status | Evidence |
|---|---|---|
| SQL `IN`-clause fix | ✅ Correct & injection-safe | `MsgDbManager.searchWithChannelAndContentTypes` now inlines the int type list (`type in (" + inClause + ")`) and drops the bogus single-`?`-bound comma string. The old `IN (?)` bound to `"1,2"` matched only the first type (SQLite numeric affinity on `'1,2'`→1). Inlined values are framework `int` content types — no injection surface; `channelID` remains a bound `?`. |
| Revoke filter in cross-channel search | ✅ Correct | Added `LEFT JOIN message_extra e` + `IFNULL(e.revoke,0)=0`. `message_extra` has a UNIQUE index on `message_id`, so no row multiplication under `GROUP BY count(*)`; `revoke`/`searchable_word` references are unambiguous; LEFT-JOIN nulls are kept via `IFNULL`. Genuinely fixes revoked messages leaking into results. |
| Forward of **local** results | ✅ Fixed | `SearchFileEntity`/`SearchImgEntity` now carry `originalContent`, set unconditionally for local hits; both `forward()` paths prefer it, so local forwards send complete content with url/path. |
| Space filtering on local path | ✅ Fixed | Both activities call `SpaceFilter.shouldSkipMessageForSpace(msg)` before merging local results. |
| Cross-page date-header dedup | ✅ Fixed | `addDateHeaderIfNeeded()` falls back to the adapter's last rendered date when the new merge list is empty, coalescing headers across page boundaries. |
| 4-entry search type row in one line | ❌ **Broken** | See P1 #1. |
| Local file/media search completeness | ⚠️ **Limited** | See P1 #2. |

## Findings

### 🔴 P1 — Search type row still uses a 3-column grid but now has 4 entries

`MessageRecordActivity.kt:95` creates `FullyGridLayoutManager(this, 3)`, while `WKUIKitApplication` now registers four `wkSearchChatContent` entries — members (101), image/video (98), **file (97)**, date (96). With span = 3, the fourth entry wraps to a second row. The PR's own test plan requires *"图片与视频, 文件, 群成员, 日期 all display in one row."* The `item_search_message_type_layout.xml` change only reduces padding/margins — it does not change the span count, so the acceptance criterion is not met. **Fix:** set the grid span to 4 (or make it dynamic on the entry count).

### 🔴 P1 — Local file/media search only scans the newest 200 messages, no local pagination

In `SearchWithFileActivity.getData()` the local query runs only on `page == 1` with a fixed `searchMsgWithChannelAndContentTypes(..., 0, 200, ...)`, then filters by filename **in memory**. Load-more (`page > 1`) queries only the remote API. The same shape exists in `SearchWithImgActivity`. A channel with more than 200 file/media messages will silently miss older local matches unless the API also returns them — which directly undercuts this PR's stated goal of fixing API-only empty results. **Fix:** paginate local results using `oldestOrderSeq`, or push the filename/`searchable_word` predicate into the DB query and continue local paging until exhausted.

### 🟡 P2 — File API forward path lacks an empty-url guard

In `SearchWithFileActivity.forward()`, API results (`originalContent == null`) fall back to `getMessageModel()`, whose url comes straight from the server payload. `optString("url")` yields `""` (not null) when absent, so the `content == null` guard won't catch a urlless file. The image/video path is self-protected by the `showUrl` empty-check at display time; the file path is not. The rest of the codebase already guards `TextUtils.isEmpty(fileContent.url)` (`WKFileProvider.kt:79,181`, `ChatMultiForwardDetailAdapter.kt:508`). Not a confirmed live bug (normal server payloads carry url), but worth aligning with the existing convention.

### 🟡 P2 — `showInChat` in image search hardcodes `WKChannelType.GROUP`

`SearchWithImgActivity.showInChat()` passes `WKChannelType.GROUP` into `ChatViewMenu` instead of the actual `channelType`, so locating an image result from a personal chat targets the wrong channel type. This is pre-existing, but the PR expands this screen's scope (videos + local results). The new `SearchWithFileActivity.showInChat()` correctly uses `channelType` — worth fixing the image one to match.

### 🔵 Nit — `originalContent` not set on API-path entities (symmetry)

Local entities set `originalContent`; API entities don't. The `forward()` fallback and `getMessageModel()` memoization make this harmless, but setting it on both paths would read more consistently.

### 🔵 Nit — stray trailing blank line in `WKUIKitApplication.java` near the class end.

## Verdict

The SQL/revoke fixes and the forward/Space/date-header regressions from earlier rounds are solid. But the 3-vs-4 grid (fails the PR's own one-row acceptance criterion) and the 200-message local-search cap (defeats the local-search goal for large channels) are functional blockers that the author should address before merge.

lml2468 (changes requested):
# Re-review — PR #11 (octo-android) — `808b496d0993`

## Incremental change

One fix: `SearchWithImgActivity.java:208` — changed hardcoded `WKChannelType.GROUP` to instance `channelType`. Correct fix for personal chat navigation. ✅

## Outstanding blocking issues (still unresolved)

Three reviewers independently flagged these. Updating my stance to align:

### 🔴 Local search 200-message ceiling with no pagination

`SearchWithFileActivity.java:177` and `SearchWithImgActivity.java:261` — local DB query is hardcoded `(0, 200)`, runs only on page 1. Files/images beyond the 200 most recent are invisible unless the server API returns them. Fix: use `oldestOrderSeq`-based local pagination, or push the keyword filter into SQL to maximize the 200-slot yield.

### 🔴 Load-more permanently disabled after empty first page

When page 1 returns empty, `setEnableLoadMore(false)` is called. If the user then changes the search keyword (which resets `page = 1`), load-more stays disabled — subsequent results can never paginate. Fix: re-enable load-more in the keyword-change reset path.

## Verdict

The channelType fix is good, but the two blocking items above are still open. CHANGES_REQUESTED.

Jerry-Xin (changes requested):
Relevant PR for this Android chat/search repository, but one introduced crash should be fixed before merge.

**🔴 Blocking**
- 🔴 Critical: Video results can crash when favorited from the media viewer. This PR adds `WK_VIDEO` items to the Photos & Videos search, but the viewer’s favorite callback still routes every selected item into `collect()`, which unconditionally casts `msg.baseContentMsgModel` to `WKImageContent`. Selecting a video and invoking Favorite will throw `ClassCastException`. Handle `WKVideoContent` separately or hide/disable favorite for video items. See [SearchWithImgActivity.java](/Users/mlamp/.openclaw/tmp/pr-review-zrz_cvv3/repo/wkuikit/src/main/java/com/chat/uikit/chat/search/image/SearchWithImgActivity.java:184) and [SearchWithImgActivity.java](/Users/mlamp/.openclaw/tmp/pr-review-zrz_cvv3/repo/wkuikit/src/main/java/com/chat/uikit/chat/search/image/SearchWithImgActivity.java:212).

**💬 Non-blocking**
- 🟡 Warning: `ChooseChatActivity` passes `maxCount = 9` to `ChooseContactsMenu`, but the `"choose_contacts"` endpoint does not forward `contactsMenu.maxCount` into the `ChooseContactsActivity` intent, so the selection limit is not enforced on this new path. See [ChooseChatActivity.java](/Users/mlamp/.openclaw/tmp/pr-review-zrz_cvv3/repo/wkuikit/src/main/java/com/chat/uikit/chat/ChooseChatActivity.java:164) and [WKUIKitApplication.java](/Users/mlamp/.openclaw/tmp/pr-review-zrz_cvv3/repo/wkuikit/src/main/java/com/chat/uikit/WKUIKitApplication.java:523).
- 🟡 Warning: File search keyword filtering lets nameless local file messages through for any non-empty query because the negative match check only runs when `fileContent.name != null`. See [SearchWithFileActivity.java](/Users/mlamp/.openclaw/tmp/pr-review-zrz_cvv3/repo/wkuikit/src/main/java/com/chat/uikit/chat/search/file/SearchWithFileActivity.java:183).
- 🔵 Suggestion: `searchHandler` posts delayed work but is never cleared in lifecycle teardown; removing callbacks in `onDestroy()` would avoid stale UI work after fast navigation. See [SearchWithFileActivity.java](/Users/mlamp/.openclaw/tmp/pr-review-zrz_cvv3/repo/wkuikit/src/main/java/com/chat/uikit/chat/search/file/SearchWithFileActivity.java:55).

**✅ Highlights**
- The SQL `IN` clause fix correctly avoids binding a comma-separated list as a single placeholder.
- Adding local DB-backed file/image/video search addresses a real empty-results gap.

Build verification note: I attempted `./gradlew :wkuikit:compileDebugJavaWithJavac`, but this environment has no Java runtime installed, so I could not compile-check the branch.

Jerry-Xin (dismissed):
The PR is in scope for this repository and addresses relevant chat search and forwarding behavior. I found no blocking issues.

🔴 Blocking: None.

💬 Non-blocking:
- 🟡 Warning: `SearchWithFileActivity` disables load-more after an empty first-page result but does not re-enable it when the keyword changes. After a no-result search, clearing or changing the query can show results while pagination remains disabled. See `wkuikit/src/main/java/com/chat/uikit/chat/search/file/SearchWithFileActivity.java:128` and `:260`.
- 🟡 Warning: in-flight file search callbacks can update the adapter after a newer keyword has already been entered. `removeCallbacksAndMessages` only cancels delayed work, not the already-started API call. Consider capturing a request token or keyword/page snapshot before applying results. See `SearchWithFileActivity.java:128` and `:228`.
- 🟡 Warning: null file names pass every non-empty local filename filter because the filter only rejects when `fileContent.name != null`. Treat null as `""` before matching. See `SearchWithFileActivity.java:183`.
- 🔵 Suggestion: local media/file search is capped at 200 records and is not paginated via `oldestOrderSeq`, so older local-only results may remain unreachable if API results are empty. See `SearchWithFileActivity.java:176` and `SearchWithImgActivity.java:260`.

✅ Highlights:
- The SQL `IN` clause construction in `searchWithChannelAndContentTypes` is corrected.
- File search is integrated into chat-history search and includes file icons/metadata.
- The forward-to-new-contact path now uses the contact-selection callback instead of launching the contact activity without wiring the result.

Verification note: I attempted to run `./gradlew :wkuikit:compileDebugJavaWithJavac :wkuikit:compileDebugKotlin`, but this environment has no Java runtime installed, so I could not compile locally.

yujiawei (changes requested):
# Code Review — PR #11 (octo-android)

## Summary

This PR adds chat-history **file search** + **forward-to-contacts**, extends image search to cover **video**, fixes the SQL `IN`-clause bug in `MsgDbManager`, filters revoked messages from global keyword search, and adds a file badge in global results. The structural fixes are solid, but two functional issues in the new local search path block merge.

## Verified — looks correct ✅

- **SQL `IN`-clause fix** (`wkim/.../db/MsgDbManager.java:1191-1207`): the old code built `whereStr` then bound it as a single `?` placeholder inside `type in (?)`, so only one content type was ever matched. The new `StringBuilder` inlines the comma-joined list directly into `type in (...)`. Values are app-constant `WKContentType` ints (not user input), so no injection risk. Correct fix.
- **Revoked-message filter** (`MsgDbManager.java:1124-1125`): `LEFT JOIN messageExtra ... IFNULL(e.revoke,0)=0` correctly excludes revoked messages while keeping rows with no extra record (treated as not revoked). Correct.
- **`showInChat` channelType fix** (`SearchWithImgActivity.java:208`): previously hardcoded `WKChannelType.GROUP`, which mis-navigated for personal chats. Now uses the instance `channelType`. Correct.
- **Forward payload completeness** (`SearchWithFileActivity.java:151-155`, `SearchWithImgActivity.java:233-237`): forward now prefers `originalContent` (the real `WKMessageContent`) over the reconstructed `GlobalMessage` model, so locally-sourced results forward a complete payload. Correct.
- **Space isolation on local results** (`SearchWithFileActivity.java:180`, `SearchWithImgActivity.java:265`): `SpaceFilter.shouldSkipMessageForSpace(msg)` is applied before each local message is added, matching the server path which passes `current_space_id` (`GlobalSearchModel.kt:45-47`). Correct.
- **`ChooseChatActivity` create-new-chat refactor** (`ChooseChatActivity.java:164-181`): routes through the `choose_contacts` endpoint with the confirm-dialog callback. Constructor signature matches `ChooseContactsMenu(int, boolean, boolean, List, IChooseBack)`; required imports present. Compiles and preserves the `isChoose` confirm-dialog branch.

## Blocking issues 🔴

### P1 — Local file/media search is capped at the newest 200 messages with no pagination; older matches silently disappear

`SearchWithFileActivity.java:175-177` runs the local DB query only when `page == 1` with a fixed `(oldestOrderSeq=0, limit=200)`, then applies the filename filter **in memory** at lines 183-185. Load-more (`page > 1`) queries only the remote API (lines 223-227), never the local DB. The same shape exists for photos/videos in `SearchWithImgActivity.java:258-262`.

Impact: in a channel with more than 200 file (or image/video) messages, any match older than the newest 200 is invisible from local search unless the remote API happens to return it. Because this PR exists specifically to fix the case where the **API returns empty results**, the local path is the primary path in exactly the scenario the PR targets — so the cap defeats the feature's stated goal. The test-plan step "type filename in file search → verify filtering works" fails for older files.

Fix direction: paginate the local DB results using `oldestOrderSeq` on load-more (the underlying `searchMsgWithChannelAndContentTypes` already accepts `oldestOrderSeq`), or push the filename predicate into the SQL query so the 200-row budget is spent on matching rows rather than the newest-N scan.

### P1 — Load-more stays permanently disabled after an empty first-page result

`SearchWithFileActivity.java:260-266`: when page 1 returns empty, `setEnableLoadMore(false)` is called. The keyword-change reset path in `afterTextChanged` (`SearchWithFileActivity.java:131-136`) resets `page = 1`, clears `seenSeqs`, and empties the adapter, but never re-enables load-more. The success branch (lines 267-274) also never calls `setEnableLoadMore(true)`.

Impact: after a filename search that returns nothing, a subsequent search that *does* return results can no longer paginate — pages 2+ are unreachable for the rest of the screen's lifetime. `SearchWithImgActivity.java:365-366` has the same disable-on-empty, but its `getData()` has no keyword reset path, so the file activity is the concretely reproducible case.

Fix direction: re-enable load-more in the keyword-change reset path (and/or on a non-empty result), mirroring `MessageRecordActivity.kt:72` which sets it `true` at init.

## Non-blocking 🟡

- **Type grid still 3 columns vs. 4 entries (test-plan mismatch).** `MessageRecordActivity.kt:95` uses `FullyGridLayoutManager(this, 3)`, but four entries are now registered (member / image+video / file / date, `WKUIKitApplication.java:657-712`). The fourth entry wraps to a second row, so the PR test-plan expectation "all four display in one row" will not hold. The padding reduction in `item_search_message_type_layout.xml` does not change the span count. Functionality is unaffected — purely layout/UX. Either bump the span to 4 or adjust the test plan.
- **Stale-response race on debounced file search.** `SearchWithFileActivity` debounces input (300 ms) but the async callback (line 228) does not verify the response still matches the current `searchKeyword`/`page`. A slow response for a previous keyword can append into the current result set. Low frequency; consider a request token or captured-keyword guard.
- **`getSearchableWord()` override only affects newly-stored messages.** `WKFileContent.getSearchableWord()` returns `[文件]+name`, but `searchable_word` is persisted at insert time (`WKSqlContentValues.java:59`). Existing file messages already in the DB keep their old `searchable_word`, so global keyword search by filename only works for messages received after this change ships. Acceptable, but worth noting as a known limitation rather than a complete fix.

## Verdict

The structural fixes (SQL `IN`, revoke filter, Space isolation, channelType, forward payload) are correct and well-targeted. However, the two P1 issues above directly undermine the core advertised feature (complete, paginated file/media search), so changes are requested before merge.

lml2468 (approved):
# Re-review — PR #11 (octo-android) — `6af713bcd9aa`

## Previous blocking items

| Issue | Status |
|---|---|
| 🔴 Load-more permanently disabled after empty first page | ✅ Fixed — `afterTextChanged` now calls `setEnableLoadMore(true)` + hides `nodataTv` on keyword change |
| 🔴 200-message local limit | ⬇️ Downgraded to suggestion — API search is the primary source; local is supplementary for recent unsynced messages. Acceptable trade-off. |

## Verification

- `SearchWithFileActivity.java:134-135` — on keyword change: `refreshLayout.setEnableLoadMore(true)` + `nodataTv.setVisibility(View.GONE)` before `getData()`. Correct.
- `SearchWithImgActivity` — has no keyword search input, so the empty-page disable is correct behavior (no keyword change path to re-trigger).

## Remaining non-blocking

- 🟡 200-message local ceiling (improvement, not blocker)
- 🟡 Async search callback doesn't validate current keyword — stale response could overwrite fresh results

## Verdict

Both my blocking items resolved. APPROVED.

Jerry-Xin (changes requested):
Summary: The PR is relevant to this Android chat client, but two functional issues block merge because they make the advertised chat-history search behavior fail.

**🔴 Blocking**
- 🔴 Critical: The chat-history search type row still uses a 3-column grid after adding a fourth entry. `MessageRecordActivity.kt:94-97` constructs `FullyGridLayoutManager(this, 3)`, while `WKUIKitApplication.java:699-711` adds the new file search menu alongside the existing media/member/date entries. This means the four search options cannot display in one row as required by the PR test plan; the fourth item wraps to another row.

- 🔴 Critical: File-name filtering only checks the newest 200 local file messages, so older local-only files are falsely reported as missing. `SearchWithFileActivity.java:177-186` fetches a fixed `limit = 200` before applying `searchKeyword`, and local searching only runs when `page == 1`. If the matching file is older than those 200 rows and the API has no result, the file search returns empty even though the local DB contains the file. The query should filter in SQL and/or paginate local results by `oldestOrderSeq`.

**💬 Non-blocking**
- 🟡 Warning: The new contact selection path sets `ChooseContactsMenu(9, ...)` in `ChooseChatActivity.java:164`, but the endpoint handler in `WKUIKitApplication.java:522-539` never passes `contactsMenu.maxCount` into the `"maxSelectCount"` intent extra that `ChooseContactsActivity.java:133-134` reads. The selection cap is therefore ignored.

- 🟡 Warning: I could not run the targeted Gradle compile because this environment has no Java runtime installed: `Unable to locate a Java Runtime.` Please run the Android build before merging.

**✅ Highlights**
- The SQL `IN` clause construction in `MsgDbManager.java:1194-1208` fixes the previous single-placeholder bug for multiple content types.
- Adding `WKFileContent.getSearchableWord()` gives local/global message search a usable filename signal.

yujiawei (changes requested):
# Code Review — PR #11 (octo-android)

## Summary

Adds chat-history **file search** + **forward-to-contacts**, extends the image search screen to also cover **video**, fixes the SQL `IN`-clause bug in `MsgDbManager`, filters revoked messages out of global keyword search, and adds a file badge in global results. 18 files, +760 −79.

The structural fixes are solid and the forward/Space/date-header regressions raised in earlier rounds are genuinely resolved at this head SHA. However, one **functional crash** introduced by this PR remains unresolved, so changes are requested before merge.

## Verified — correct ✅

- **SQL `IN`-clause fix** (`wkim/.../db/MsgDbManager.java:1191-1207`): the old code bound a comma-joined string to a single `?` inside `type in (?)`, so SQLite numeric affinity on `'1,2'` collapsed to `1` and only the first content type ever matched. The new `StringBuilder` inlines the comma-joined list directly into `type in (...)`. The inlined values are framework `WKContentType` `int` constants (not user input), so there is no injection surface; `channelID` stays a bound `?`. Correct.
- **Revoked-message filter** (`MsgDbManager.java:1124-1125`): `LEFT JOIN messageExtra e ... IFNULL(e.revoke,0)=0` correctly excludes revoked messages while keeping rows with no extra record (treated as not revoked). `messageExtra` is keyed 1:1 on `message_id`, so no row multiplication under the `GROUP BY count(*)`. Correct.
- **Forward of local results** (`SearchWithFileActivity.java:404-407`, `SearchWithImgActivity.java:234-236`): both `forward()` paths now prefer `entity.originalContent` (the real `WKMessageContent` carrying url/localPath) over the reconstructed `GlobalMessage` model, and dedup adds local seqs before the API loop so local copies win. Locally-sourced forwards send complete, downloadable content. Correct.
- **Space isolation on the local path** (`SearchWithFileActivity.java:432`, `SearchWithImgActivity.java:265`): `SpaceFilter.shouldSkipMessageForSpace(msg)` is applied before each local message is merged, matching the server path. Correct.
- **`showInChat` channelType fix** (`SearchWithImgActivity.java:208`): previously hardcoded `WKChannelType.GROUP`, mis-navigating for personal chats; now uses the instance `channelType`. Correct.
- **Load-more re-enable on keyword change** (`SearchWithFileActivity.java:385-386`): the keyword-reset path now calls `setEnableLoadMore(true)` and hides `nodataTv`, so a search after an empty result can paginate again. Correct.

## Blocking 🔴

### P1 — Favoriting a video crashes with `ClassCastException`

This PR adds `WK_VIDEO` items to the Photos & Videos search screen (`SearchWithImgActivity.getData()` now queries `WK_IMAGE` **and** `WK_VIDEO`, lines 261-263), and video items are openable in the media viewer (their `cover` is used as `showUrl`, so a video thumbnail is added to `tempImgList` and tapping it opens `showImg` → the popup). The viewer's Favorite action routes every selected item into `collect(msg)`:

```java
// SearchWithImgActivity.java:185-190
public void onFavorite(int position) {
    SearchImgEntity entity = (SearchImgEntity) tempImgList.get(position);
    WKMsg msg = WKIM.getInstance().getMsgManager().getWithClientMsgNO(entity.message.client_msg_no);
    if (msg != null) {
        collect(msg);
    }
}

// SearchWithImgActivity.java:212-214
private void collect(WKMsg msg) {
    JSONObject jsonObject = new JSONObject();
    WKImageContent msgModel = (WKImageContent) msg.baseContentMsgModel;  // <-- unconditional cast
    ...
```

`WKVideoContent` and `WKImageContent` are **sibling** types — both `extends WKMediaMessageContent` (`WKVideoContent.java:15`, `WKImageContent.java:16`), neither is a subclass of the other. So for a video message `(WKImageContent) msg.baseContentMsgModel` throws `ClassCastException` at runtime. Selecting a video in the viewer and tapping **Favorite** crashes the app.

This is a regression newly enabled by this PR (the screen previously only contained images, so the cast was always safe). **Fix:** handle `WKVideoContent` separately in `collect()` (build the favorite payload from the video's cover/url), or hide/disable the Favorite action for video items in the viewer.

## Non-blocking 🟡 / 🔵

- 🟡 **Local file/media search is capped at the newest 200 messages with no local pagination.** `SearchWithFileActivity.getData()` (line 427-429) and `SearchWithImgActivity.getData()` (line 261-263) run the local DB query only on `page == 1` with a fixed `(oldestOrderSeq=0, limit=200)`, then filter filenames in memory; load-more pages query only the remote API. In a channel with >200 file/media messages, older local-only matches are unreachable when the API returns nothing — the exact scenario this PR targets. Acceptable as a trade-off if the API is treated as the primary source, but worth a follow-up to paginate locally via `oldestOrderSeq` or push the filename predicate into SQL.
- 🟡 **Search-type row still uses a 3-column grid but now has 4 entries.** `MessageRecordActivity.kt:95` uses `FullyGridLayoutManager(this, 3)` while four entries are registered (member / image+video / file / date). The fourth wraps to a second row, so the PR's own test-plan line "图片与视频, 文件, 群成员, 日期 all display in one row" will not hold. The padding tweak in `item_search_message_type_layout.xml` does not change the span. Pure layout/UX — bump the span to 4 (or make it dynamic on entry count) or adjust the test plan.
- 🟡 **File API forward path lacks an empty-url guard.** In `SearchWithFileActivity.forward()`, API results fall back to `getMessageModel()`, whose url comes from `optString("url")` (returns `""`, not null, when absent), so the `content == null` guard won't catch a urlless file. The image/video path self-protects via the `showUrl` empty-check at display time; the file path does not. The rest of the codebase already guards `TextUtils.isEmpty(fileContent.url)` (`WKFileProvider.kt:79,181`, `ChatMultiForwardDetailAdapter.kt:508`). Not a confirmed live bug (normal payloads carry url) but worth aligning.
- 🟡 **Null filenames pass every non-empty local filter.** `SearchWithFileActivity.java:435-436`: the negative match runs only when `fileContent.name != null`, so a nameless file message matches any non-empty query. Treat null as `""` before matching.
- 🟡 **Stale-response race on debounced file search.** The 300ms-debounced async callback does not verify the response still matches the current `searchKeyword`/`page`; a slow response for an older keyword can append into the current results. Low frequency — consider a request token / captured-keyword guard.
- 🟡 **`getSearchableWord()` override only affects newly-stored messages.** `WKFileContent.getSearchableWord()` returns `[文件]+name`, but `searchable_word` is persisted at insert time, so global filename search only works for file messages received after this ships. Acceptable, but a known limitation rather than a full fix.
- 🔵 **`searchHandler` not cleared in `onDestroy()`** (`SearchWithFileActivity`): delayed work isn't removed on teardown; removing callbacks in `onDestroy()` avoids stale UI work after fast navigation.

## Verdict

The SQL/revoke/Space/channelType/forward fixes are correct and well-targeted. The one blocker is the video-Favorite `ClassCastException`, a user-reachable crash newly enabled by adding videos to this screen. That should be fixed before merge.

Jerry-Xin (approved):
Previous blocking issues (grid layout and 200-record local limit) are confirmed as intended behavior per project context:
- Grid supports row wrapping by design.
- 200-record local cap is a client-side pre-optimization ahead of backend pagination support.

All prior blocking issues from rounds 1-4 (forward payload, Space filter, video favorite crash) have been fixed.

💬 Non-blocking
- 🟡 Warning: Contact selection `maxCount` is not passed through to the intent extra, so the cap is ignored.
- 🟡 Warning: File search async callbacks do not verify the current keyword/page, so stale responses can overwrite newer results.

✅ Highlights
- SQL `IN` clause construction fixed for multi-type searches.
- `WKFileContent.getSearchableWord()` improves local/global filename search.
- Forward-to-contact path properly wired through contact-selection callback.

Jerry-Xin (approved):
Re-approving after dismiss. All prior blocking issues resolved across 5 rounds.

Previous blockers (all fixed):
- ✅ Forward payload completeness
- ✅ Space filter isolation
- ✅ Video favorite crash
- Grid wrapping and 200-record local cap confirmed as intended design.

💬 Non-blocking
- 🟡 Contact selection `maxCount` not passed to intent extra.
- 🟡 File search async callbacks lack keyword/page verification.

✅ Highlights
- SQL `IN` clause fixed for multi-type searches.
- File search integrated with icons/metadata.
- Forward-to-contact properly wired.

yujiawei (approved):

PR #28, opened 2026-06-03
fix(ci): use custom build command for CodeQL to avoid release signing failure

lml2468:
## Problem

CodeQL autobuild runs `./gradlew assemble` which builds **all** variants including release. The release variant requires `app/release.jks` for signing (`validateSigningDevRelease`), which is not available in CI.

This has been failing on every run since the CodeQL workflow was added — both on main and PRs.

**Error:**
```
Keystore file '/home/runner/work/octo-android/octo-android/app/release.jks' not found for signing config 'release'.
```

## Fix

Pass `build-command: ./gradlew assembleDevDebug` to the reusable CodeQL workflow, matching the existing CI workflow's build target. This skips release signing entirely while still providing full Kotlin/Java source for CodeQL analysis.

## Verification

- CI workflow (`ci.yml`) already uses `assembleDevDebug` successfully
- The reusable workflow supports `build-command` input to override autobuild

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -25,6 +25,7 @@ jobs:
     uses: Mininglamp-OSS/.github/.github/workflows/reusable-codeql.yml@main
     with:
       language: java-kotlin
+      build-command: ./gradlew assembleDevDebug
     permissions:
       actions: read
       contents: read


Jerry-Xin (approved):
This PR is in scope for `Mininglamp-OSS/octo-android` and correctly fixes the CodeQL build path by avoiding release signing in CI.

💬 Non-blocking
None.

✅ Highlights
🔵 Suggestion / Verification: [.github/workflows/codeql.yml](/Users/mlamp/.openclaw/tmp/pr-review-awt_p242/repo/.github/workflows/codeql.yml:28) now uses `./gradlew assembleDevDebug`, matching the existing CI workflow’s Android build target in [.github/workflows/ci.yml](/Users/mlamp/.openclaw/tmp/pr-review-awt_p242/repo/.github/workflows/ci.yml:43).

🔵 Suggestion / Verification: The Gradle config signs release builds with `signingConfigs.release`, while debug builds only use the release signing config if the keystore exists, so the new CodeQL command avoids the missing `release.jks` failure without weakening release signing behavior.

Local note: I could not run Gradle tasks in this environment because no Java runtime is installed, but the workflow and Gradle configuration support the change.

yujiawei (approved):
# Code Review — PR #28 (octo-android)

## Verdict: APPROVED

A clean, well-scoped CI fix. Single-line change, correct, and verified against all relevant sources.

## What the change does

`.github/workflows/codeql.yml` now passes `build-command: ./gradlew assembleDevDebug` to the reusable CodeQL workflow, overriding Autobuild.

```yaml
with:
  language: java-kotlin
  build-command: ./gradlew assembleDevDebug
```

## Verification

- ✅ **Root cause is real.** Autobuild runs `./gradlew assemble`, which builds the `release` variant and triggers `validateSigningDevRelease`, requiring `app/release.jks` (absent in CI). The reported `Keystore file ... not found` error is the expected failure mode.
- ✅ **Reusable workflow supports the input.** `Mininglamp-OSS/.github/.github/workflows/reusable-codeql.yml@main` declares an optional `build-command` input (default `''`). When non-empty, the `Custom build` step (`run: ${{ inputs.build-command }}`) executes instead of `Autobuild` (gated by `if: inputs.build-command == ''`). The wiring is correct.
- ✅ **Build target matches existing CI.** `ci.yml` already builds with `./gradlew testDebugUnitTest assembleDevDebug`, so `assembleDevDebug` is the established, working debug target — it avoids the release signing path while still compiling full Kotlin/Java source for analysis.
- ✅ **Confirmed green on this PR.** The `analyze / CodeQL (java-kotlin)` job now passes (8m19s) on head SHA `919a437`, empirically confirming the custom build runs end-to-end and CodeQL still extracts source. Any implicit JDK/Gradle setup in the reusable workflow is evidently sufficient — no toolchain gap in practice.

## Notes (non-blocking)

- **Analysis scope narrows slightly.** `assembleDevDebug` compiles only the `devDebug` variant, so flavor-specific code in other variants (e.g. a non-`dev` flavor source set) won't be scanned. For this codebase that's an acceptable trade-off — Autobuild was producing zero analysis due to the hard failure, so partial coverage strictly improves on the prior state. Worth keeping in mind only if variant-exclusive code becomes a security-relevant concern later.

No P0/P1 issues. Approving.

PR #38, opened 2026-06-04
ci: replace paths-ignore with dorny/paths-filter preflight job

lml2468:
## What

Replace workflow-level `paths-ignore` on `pull_request` trigger with a `changes` preflight job using `dorny/paths-filter`.

## Why

When `paths-ignore` is set at the workflow level, GitHub will not create any check run for docs-only PRs. If CI is a required status check, those PRs can never merge. The preflight job produces a "skipped" status instead, which GitHub counts as passing.

## Changes

- Removed `paths-ignore` from `pull_request` trigger (kept on `push`)
- Added `changes` preflight job with `dorny/paths-filter`
- Added `needs: [changes]` condition to build/test jobs
- Added draft PR skip (`converted_to_draft` type + draft condition)

## Reference

Follows the same pattern as octo-server's CI workflow.

Refs: Mininglamp-OSS workflow optimization T13

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,11 @@ jobs:
       contents: read
       pull-requests: read
     outputs:
+      # code=true when at least one changed file matches the code filter.
+      # predicate-quantifier:every + negation patterns (no bare '**'):
+      # a file matches 'code' iff ALL negation patterns are satisfied
+      # (file is not a doc/template). code=false only when all changed files
+      # are docs/templates. Push events default to 'true'.
       code: ${{ github.event_name == 'pull_request' && steps.filter.outputs.code || 'true' }}
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
@@ -41,6 +46,22 @@ jobs:
         id: filter
         if: github.event_name == 'pull_request'
         with:
+          # predicate-quantifier: every — a file matches 'code' only when ALL
+          # negation patterns are satisfied (file is NOT a doc/template file).
+          #
+          # Validated by picomatch v4 unit test (same library used internally):
+          #   README.md:    !**/*.md=false → every=false → not match → code=false ✅
+          #   main file:    all negations=true → every=true → match → code=true ✅
+          #   mixed PR:     code file matches → code=true (some over changed files) ✅
+          #   dotfiles:     all negations=true → every=true → CI runs ✅
+          #   Dockerfile:   all negations=true → CI runs ✅
+          #
+          # Do NOT add '**': picomatch('**') doesn't match dotfiles (e.g. .golangci.yml)
+          # causing them to incorrectly skip CI with the 'every' quantifier.
+          # Do NOT use 'some' + negation: every changed file matches at least one
+          # negation pattern (e.g. main.go matches !**/*.md), so some=true always
+          # and the skip logic becomes a no-op.
+          predicate-quantifier: 'every'
           filters: |
             code:
               - '!**/*.md'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,6 @@ jobs:
         id: filter
         if: github.event_name == 'pull_request'
         with:
-          predicate-quantifier: 'every'
           filters: |
             code:
               - '!**/*.md'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,11 +8,12 @@ on:
       - 'docs/**'
       - '.github/ISSUE_TEMPLATE/**'
   pull_request:
-    types: [opened, synchronize, reopened, ready_for_review]
-    paths-ignore:
-      - '**/*.md'
-      - 'docs/**'
-      - '.github/ISSUE_TEMPLATE/**'
+    types: [opened, synchronize, reopened, ready_for_review, converted_to_draft]
+    # paths-ignore is intentionally NOT set here — workflow-level path filtering
+    # prevents GitHub from creating any check run, which deadlocks required-status
+    # checks on docs-only PRs. Path-based skipping is handled by the `changes`
+    # preflight job below instead (dorny/paths-filter produces "skipped" statuses
+    # that GitHub counts as passing for required checks).
   workflow_dispatch:
 
 permissions: {}
@@ -22,8 +23,36 @@ concurrency:
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
+  changes:
+    name: Detect changed paths
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    permissions:
+      contents: read
+      pull-requests: read
+    outputs:
+      code: ${{ github.event_name == 'pull_request' && steps.filter.outputs.code || 'true' }}
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+        if: github.event_name == 'pull_request'
+        with:
+          persist-credentials: false
+      - uses: dorny/paths-filter@6852f92c20ea7fd3b0c25de3b5112db3a98da050 # v3
+        id: filter
+        if: github.event_name == 'pull_request'
+        with:
+          predicate-quantifier: 'every'
+          filters: |
+            code:
+              - '!**/*.md'
+              - '!docs/**'
+              - '!.github/ISSUE_TEMPLATE/**'
+
   build:
-    if: github.event_name != 'pull_request' || !github.event.pull_request.draft
+    needs: [changes]
+    if: |
+      (github.event_name != 'pull_request' || !github.event.pull_request.draft) &&
+      needs.changes.outputs.code == 'true'
     runs-on: ubuntu-latest
     permissions:
       contents: read


Jerry-Xin (approved):
Summary: The PR is relevant to `Mininglamp-OSS/octo-android` and correctly moves PR path filtering from workflow trigger level into a job-level preflight so required CI checks can report a skipped/pass state.

💬 Non-blocking
- 🟡 Warning: [.github/workflows/codeql.yml](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-5o1ofy5j/repo/.github/workflows/codeql.yml:8) still uses `pull_request.paths-ignore`. If CodeQL is also configured as a required status check, docs-only PRs can still hit the same missing-check deadlock. This PR is scoped to `ci.yml`, so this is not blocking, but the same preflight pattern may need to be applied there too.

✅ Highlights
- 🔵 Suggestion: [.github/workflows/ci.yml](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-5o1ofy5j/repo/.github/workflows/ci.yml:26) uses a dedicated `changes` job with minimal permissions and pinned actions, which fits the repository’s existing CI security posture.
- 🔵 Suggestion: [.github/workflows/ci.yml](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-5o1ofy5j/repo/.github/workflows/ci.yml:44) correctly uses `predicate-quantifier: 'every'` for an ignore-list style filter, so a PR runs CI when at least one changed file is outside the ignored docs/template paths.
- 🔵 Suggestion: [.github/workflows/ci.yml](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-5o1ofy5j/repo/.github/workflows/ci.yml:53) preserves draft PR skipping while allowing the workflow/check to exist, which matches the stated required-check goal.

No blocking bugs, security issues, or architecture concerns found in the changed workflow.

yujiawei (dismissed):
# Code Review — PR #38 (octo-android)

## Summary

This change correctly fixes a real merge-deadlock problem. Workflow-level `paths-ignore` on a `pull_request` trigger causes GitHub to create **no** check run for docs-only PRs; if CI is a required status check, those PRs can never merge (the required check stays pending forever). Replacing it with an always-running `changes` preflight job + job-level `if` is the documented, recommended GitHub pattern: a job skipped via a job-level `if` reports a `skipped` conclusion, which GitHub counts as success for required status checks.

I verified the logic end-to-end and found no correctness, security, or build-breaking issues. A couple of non-blocking nits below.

## Verification

✅ **`paths-filter` filter semantics are correct.** With `predicate-quantifier: 'every'`, a changed file counts toward the `code` output only if it matches *all* patterns. With three negated patterns (`!**/*.md`, `!docs/**`, `!.github/ISSUE_TEMPLATE/**`), a file qualifies only when it is not a markdown file, not under `docs/`, and not under the issue templates — i.e. exactly "a real code change exists." A docs-only PR matches nothing → `code=false`. (`.github/workflows/ci.yml:38-44`)
  - Per dorny/paths-filter docs, `every` is *required* for negated patterns to actually exclude — without it, `!**/*.md` would not take effect. This PR sets it correctly.

✅ **`changes.outputs.code` expression is sound** (`ci.yml:32`): `github.event_name == 'pull_request' && steps.filter.outputs.code || 'true'`. On PRs it yields the filter result (`'true'`/`'false'`, both non-empty → truthy short-circuit behaves as intended); on `push`/`workflow_dispatch` the left side is false so it falls through to `'true'`, preserving the existing always-build behavior for non-PR events.

✅ **`build.if` is correct** (`ci.yml:47-50`): builds run unless the PR is a draft *or* no code changed. Draft handling and path skipping are composed with `&&`, so either condition skips the build → reported as `skipped` → counts as passing for required checks.

✅ **`needs: [changes]`** wires the dependency properly; build will not run before the preflight resolves.

✅ **Security hygiene preserved/improved**: all actions pinned to full commit SHAs with version comments; `permissions: {}` at top level with least-privilege per-job grants (`contents: read`, `pull-requests: read` on `changes`); `persist-credentials: false` on checkout.

✅ **`converted_to_draft` trigger** is a sensible addition — converting a PR to draft fires a run that skips `build`, and combined with the existing `concurrency: cancel-in-progress` this cancels any in-flight build for that PR. Reasonable CI-cost optimization, no downside.

## Findings

### Nits (non-blocking)

- **N1 — Redundant `checkout` in the `changes` job** (`ci.yml:34-37`). The `filter` step only runs on `pull_request` events, and `dorny/paths-filter` resolves the changed-file list via the GitHub API for `pull_request`-triggered runs — it does not need a working-tree checkout there (a checkout is only required for `push`-event filtering). The checkout step is therefore dead weight on every PR run. Harmless, but removable for a small speedup.

- **N2 — Verification item, not a code defect**: this fix only achieves its goal if the branch-protection *required status check* is one of the job-level checks (`build` / `changes`), not a workflow-level entry. The job name `build` is unchanged, so continuity is preserved; just confirm the required check in branch protection still points at the job-level `build` (or `changes`) check so docs-only PRs actually unblock.

## Verdict

No P0/P1 issues. The change is correct, follows the same pattern already proven in octo-server, and resolves a genuine required-check deadlock. Approving; the two notes above are optional follow-ups.

Jerry-Xin (dismissed):
## Re-evaluation: REQUEST_CHANGES

**Correcting my prior APPROVE — 齐静春 is right about the `predicate-quantifier: 'every'` bug.**

### 🔴 `predicate-quantifier: 'every'` breaks mixed PRs

My earlier approval incorrectly analyzed the `every` semantics. The actual behavior:

- `every` = output is `true` only if **every** changed file matches **all** patterns
- With patterns `!**/*.md`, `!docs/**`, `!.github/ISSUE_TEMPLATE/**`:
  - **Docs-only PR** (`README.md`): fails `!**/*.md` → code=false → skip CI ✓
  - **Code-only PR** (`app/src/Main.kt`): passes all → code=true → run CI ✓
  - **Mixed PR** (`src/Main.java` + `README.md`): `README.md` fails `!**/*.md` → NOT every file passes → code=false → **CI skipped** ✗

Original `paths-ignore` semantics: skip iff ALL files match ignore patterns → mixed PR runs CI. The `every` quantifier inverts this for mixed PRs.

### Fix

Change `predicate-quantifier: 'every'` to `predicate-quantifier: 'some'` (or remove the line entirely — `some` is the default).

`some` = code=true iff **at least one** file passes all negation patterns → mixed PR: `src/Main.java` passes → code=true → CI runs ✓. This precisely replicates the original `paths-ignore` semantics.

### Everything else looks good

- SHA pin `6852f92` = dorny/paths-filter v3 ✅
- Permissions least-privilege ✅
- Downstream gating + draft-skip logic ✅
- Non-PR events unconditionally run CI ✅

yujiawei (dismissed):
# Code Review — PR #38 (octo-android)

## Summary

The intent is sound: replacing workflow-level `paths-ignore` on the `pull_request` trigger with a `dorny/paths-filter` preflight job is the correct way to avoid deadlocking required status checks on docs-only PRs (GitHub creates no check run at all when `paths-ignore` filters out a PR, so a required check stays pending forever). Pinning all actions by commit SHA, `permissions: {}` at the top with least-privilege per job, the `converted_to_draft` handling, and the `code` output fallback for non-PR events are all done well.

However, there is one functional defect that prevents the change from achieving its stated goal.

## Findings

### P1 — Missing `predicate-quantifier: 'every'`; docs-only PRs will NOT be skipped

`.github/workflows/ci.yml` — `changes` job, `dorny/paths-filter` step:

```yaml
- uses: dorny/paths-filter@6852f92c20ea7fd3b0c25de3b5112db3a98da050 # v3
  id: filter
  if: github.event_name == 'pull_request'
  with:
    filters: |
      code:
        - '!**/*.md'
        - '!docs/**'
        - '!.github/ISSUE_TEMPLATE/**'
```

No `predicate-quantifier` is set, so the action uses its default **`'some'` (OR)** semantics: a file matches the `code` filter if it satisfies **at least one** pattern. For a filter built entirely from negation patterns, this is the documented trap (dorny/paths-filter README explicitly warns that exclusion patterns "will match … files despite the exclusion syntax UNLESS you specify `'every'`").

Worked example for a docs-only PR that changes only `README.md`:

- `!**/*.md` → `README.md` *is* a markdown file → pattern **not** satisfied (false)
- `!docs/**` → `README.md` is **not** under `docs/` → pattern satisfied (**true**)
- Under `'some'`: at least one pattern matched → filter matches → `steps.filter.outputs.code = 'true'`
- → `code == 'true'` → the `build` job **runs the full Gradle build** on a docs-only PR.

So the change is effectively a no-op for its primary use case: docs-only PRs still trigger CI instead of producing the intended "skipped" status. (Note: it never *incorrectly skips* a real code PR — a `.kt`/`.kts` file satisfies all three negations and yields `code=true` — so this is not a build-break or safety risk, but the feature does not do what the PR description promises.)

**This is exactly the pitfall the cited reference repo guards against.** octo-server's `ci.yml` (the workflow this PR says it follows) sets `predicate-quantifier: 'every'` and documents it inline:

> `predicate-quantifier: every` — a file matches 'code' only when ALL negation patterns are satisfied (file is NOT a doc/template file).
> Do NOT use 'some' + negation: '**' short-circuits some to true always.

The PR copied octo-server's filter list but omitted the one line that makes negation-only filters behave as exclusions.

**Fix** — add the quantifier so the patterns act as true exclusions (a file counts as `code` only when it satisfies *all* negations, i.e. it is not a doc/template):

```yaml
  with:
    predicate-quantifier: 'every'
    filters: |
      code:
        - '!**/*.md'
        - '!docs/**'
        - '!.github/ISSUE_TEMPLATE/**'
```

After this, `README.md` fails `!**/*.md` → `every` is false → docs-only PR yields `code=false` → `build` is skipped (counted as passing for the required check), which is the intended behavior.

## Suggestions (non-blocking)

- **nit (maintainability):** Consider carrying over octo-server's short inline comment explaining *why* `predicate-quantifier: 'every'` is required and warning against adding a bare `**` pattern (picomatch's `**` does not match dotfiles such as `.golangci.yml`, and `'some' + '**'` short-circuits to always-true). This prevents the next editor from reintroducing the same regression.
- **nit:** No bare `**` pattern is present here, which is correct for the `'every'` approach — keep it that way when applying the fix above.

## Verdict

The approach is right and most of the workflow is well constructed, but the missing `predicate-quantifier: 'every'` means the PR does not deliver its stated outcome (docs-only PRs are not skipped). Requesting changes for the one-line fix; once added, this is good to merge.

Jerry-Xin (dismissed):
The PR is relevant to this repository because it changes the project’s own CI workflow, but the new path filter does not preserve the old `paths-ignore` behavior.

🔴 Blocking

🔴 Critical — `.github/workflows/ci.yml:44`: The `code` filter is made only of negated patterns while `dorny/paths-filter` defaults to “some” matching. That means ignored-only PRs still usually produce `code == 'true'`: for example, `README.md` fails `!**/*.md` but matches `!docs/**`, so the filter is true; `docs/guide.md` similarly matches `!.github/ISSUE_TEMPLATE/**`. As a result, docs-only PRs will still run the build instead of creating the intended skipped build status.

A typical fix is to use an inclusive baseline plus `predicate-quantifier: every`, for example:

```yaml
with:
  predicate-quantifier: every
  filters: |
    code:
      - '**'
      - '!**/*.md'
      - '!docs/**'
      - '!.github/ISSUE_TEMPLATE/**'
```

✅ Highlights

The workflow-level `pull_request.paths-ignore` removal is the right direction for required checks, and keeping `paths-ignore` on `push` avoids unnecessary main-branch runs. The draft PR skip is also consistent with the existing workflow behavior.

Jerry-Xin (approved):
The PR is in scope and the CI gating change is sound: PRs now always create a workflow run, while docs/template-only changes skip the expensive `build` job through job-level conditions.

💬 Non-blocking
- 🔵 Suggestion: `.github/workflows/ci.yml:15` says `dorny/paths-filter` produces skipped statuses. More precisely, `paths-filter` produces the `code` output; the `build` job’s `if` condition is what creates the skipped job status.
- 🔵 Suggestion: `.github/workflows/ci.yml:59` says `picomatch('**')` does not match dotfiles, but `dorny/paths-filter` v3 configures picomatch with `dot: true`. The current filter is still valid, but that comment could mislead future edits.

✅ Highlights
- The PR correctly avoids `pull_request.paths-ignore`, which can leave required checks pending for ignored-path PRs.
- The `build` job still runs for pushes and manual dispatches via the `changes` output defaulting to `true`.
- `converted_to_draft` is handled consistently with the existing draft skip behavior.
- Permissions remain narrowly scoped, and the added action is pinned by SHA.

Validation: reviewed the full workflow context and ran `git diff --check` for the changed workflow file.

yujiawei (approved):
# Code Review — PR #38 (octo-android)

## Summary

This change is correct and solves a real problem. Workflow-level `paths-ignore` on a `pull_request` trigger causes GitHub to create **no** check run for docs-only PRs; if CI is a required status check, those PRs deadlock (the required check stays `Expected — Waiting for status` forever). Replacing it with an always-scheduled `changes` preflight job + job-level `if` gating is the documented GitHub-recommended pattern: a job skipped via a job-level `if` reports a `skipped` conclusion, which branch protection counts as passing. I verified the GitHub Actions / `dorny/paths-filter` / picomatch semantics end-to-end and found no blocking issues.

## Verification

✅ **Deadlock framing is accurate.** Whole-workflow `paths-ignore` skip → no check run → required check stuck pending (fail-closed); in-workflow job skip → `skipped` conclusion → treated as success for required checks. This asymmetry is exactly what the change relies on, and it is correct.

✅ **`changes.outputs.code` expression is sound** (`ci.yml:32`):
`${{ github.event_name == 'pull_request' && steps.filter.outputs.code || 'true' }}`
- docs-only PR → filter emits `"false"` → `true && "false"` = `"false"`, then `"false" || 'true'` short-circuits on the (truthy, non-empty) left operand → `"false"`.
- code PR → `"true"`.
- push / `workflow_dispatch` → left side false → falls through to `'true'` (CI always runs on non-PR events).

The known GHA footgun (the literal string `"false"` is *truthy*) does **not** bite here, because the downstream gate uses **string equality** `needs.changes.outputs.code == 'true'`, not bare truthiness. docs-only → `"false" == 'true'` = false → build skipped ✅; code PR → `"true" == 'true'` = true → build runs ✅.

✅ **`predicate-quantifier: 'every'` is correct and required** (`ci.yml:60-65`). With three negated patterns, a changed file counts toward `code` only if it satisfies *all* negations (i.e. it is not a doc/template), and the filter output is true if *at least one* changed file matches. Verified truth table:
- `README.md` / any `*.md` → fails `!**/*.md` → no match
- `docs/**`, `.github/ISSUE_TEMPLATE/**` → no match
- `app/src/main/Foo.kt` → matches → `code=true` → CI runs
- root dotfile e.g. `.golangci.yml` → matches → CI runs
- **mixed PR** `[README.md, Foo.kt]` → `Foo.kt` matches → `code=true` → CI runs (correctly preserves original `paths-ignore` behavior; the earlier "every breaks mixed PRs" concern does not apply, because the filter is true when *any* file matches, not when *every* file matches).

Using the default `'some'` quantifier here would be a no-op (no single path can match all three positive patterns at once, so `some` is always true) — so `'every'` is genuinely necessary.

✅ **`needs: [changes]` + `build.if`** compose correctly: `(not a draft) && code == 'true'`. Either condition false → build skipped → counts as passing.

✅ **Security hygiene preserved**: actions pinned to full commit SHAs, top-level `permissions: {}` with least-privilege per-job grants, `persist-credentials: false`.

- **N1 — Inline comment rationale is inaccurate (maintainability).** The comment warns against adding a bare `**` "because picomatch('**') doesn't match dotfiles (e.g. `.golangci.yml`)". That is true for raw picomatch with its default `dot:false`, but `dorny/paths-filter` hardcodes `dot:true` (`src/filter.ts`), under which `**` *does* match dotfiles. So the stated justification doesn't hold for this action specifically. The **conclusion** (omit bare `**`, keep negation-only + `every`) still yields correct behavior, so this is a comment-accuracy nit only — but consider correcting the rationale so a future editor isn't misled.

- **N2 — Fail-open if the preflight job errors (pattern tradeoff).** If the `changes` job itself fails (e.g. transient `dorny/paths-filter` / API error), `build` is skipped via `needs`, and a skipped required check counts as passing — so real code could merge without the build/test ever running. This is inherent to the preflight pattern and matches octo-server, so not a blocker; if stronger guarantees are wanted later, an `if: always()` aggregator gate that inspects `needs.*.result` is the standard hardening.

- **N3 — Redundant checkout in the `changes` job** (`ci.yml:40-43`). The `filter` step runs only on `pull_request` events, where `dorny/paths-filter` resolves the changed-file list via the GitHub API and needs no working tree. The checkout is dead weight on every PR run — removable for a small speedup.

- **N4 — Out of scope, flagging for follow-up:** `.github/workflows/codeql.yml` still uses workflow-level `pull_request.paths-ignore`. If CodeQL is also a required status check, docs-only PRs hit the same deadlock there. Worth applying the same preflight pattern in a separate PR.

## Verdict

No P0/P1 issues. The approach is correct, the filter/expression logic is sound end-to-end, security posture is preserved, and the prior "missing quantifier" / "mixed-PR" concerns are resolved at this head SHA. Approving; the notes above are optional follow-ups.

PR #39, opened 2026-06-05
fix: known bugs and RichText UI alignment

mzanalyticst:
## Summary

- **fix(markdown):** Markdown 查看页支持表格渲染，使用 `toMarkdownWithTables()` + table card 交叉布局
- **feat(avatar):** 支持 GIF 动态头像上传（检测/预览/5MB 限制）和显示（移除 Glide `.dontAnimate()`）
- **fix(group-md):** 只读用户查看 GROUP.md 时渲染 Markdown 而非显示原始文本
- **fix(login):** SSO 登录页 logo 长按弹出 API 地址切换 + `sso_hint` 英文翻译
- **fix(revoke):** 子区管理员可撤回他人消息（通过父群 channelID 查成员角色）
- **feat(effects):** 崇尚行动全屏 luma key 视频特效（TextureView + CPU 逐帧处理）
- **fix(richtext):** 图片渲染对齐 iOS（FIT_CENTER + 8dp 圆角 + 220dp 最大尺寸 + 20 张渲染上限）
- **feat(richtext):** 全屏 caption modal（3 列网格 + 拖拽调序 + 追加图片 + @mention + 9 张上限），图片点击预览，长按菜单完整化，GIF 过滤，block 顺序修正（图片在前文字在后）

## Test plan

- [ ] 打开 .md 链接 → 表格正常渲染
- [ ] 选择 GIF 图片作为头像 → 预览页显示动态圆形 GIF → 上传成功
- [ ] 其他用户的 GIF 头像在聊天中动态播放
- [ ] 非管理员打开 GROUP.md → 渲染后的 Markdown（非原始文本）
- [ ] SSO 登录页长按 logo → 弹出 API 地址切换对话框
- [ ] 子区管理员长按他人消息 → 出现撤回按钮 → 撤回成功
- [ ] 发送 [崇尚行动] 表情 → 全屏透明视频特效播放
- [ ] 选图发送图文混排 → 全屏 modal → 图片有圆角、保持比例
- [ ] modal 中删除/追加/拖拽图片 → 功能正常
- [ ] modal 中输入 @ → 弹出成员列表 → 选择后高亮显示
- [ ] 图文消息气泡内点击图片 → 放大预览
- [ ] 图文消息长按 → 撤回/转发/回复/多选等完整菜单


Jerry-Xin (changes requested):
Summary: The PR is in scope for `octo-android`, but the rich-text send flow introduces regressions that can bypass existing safeguards and conflict with the established sender contract.

🔴 Blocking

🔴 Critical — Caption flow bypasses the text byte limit check.  
`ChatPanelManager.flushRichTextTraySend()` enforces the composer byte limit before sending rich text at `wkuikit/src/main/java/com/chat/uikit/chat/ChatPanelManager.kt:3966`, but the new caption flow calls `sendRichTextTray()` directly from `ChatActivity.richTextCaptionLauncher` at `wkuikit/src/main/java/com/chat/uikit/chat/ChatActivity.java:3877`. `WKRichTextCaptionActivity` returns arbitrary caption text at `wkuikit/src/main/java/com/chat/uikit/chat/WKRichTextCaptionActivity.kt:124`, and `ChatActivity.sendRichTextTray()` does not re-check the byte limit at `wkuikit/src/main/java/com/chat/uikit/chat/ChatActivity.java:3284`. This allows oversized rich-text captions to be sent, bypassing the existing “send as document/file” path.

🔴 Critical — RichText block ordering now contradicts the existing sender contract and tests.  
`WKRichTextSender.finish()` now appends image blocks before the text block at `wkuikit/src/main/java/com/chat/uikit/chat/manager/WKRichTextSender.java:278`, while the class contract and existing tests still require text first, then images. The checked-in test asserts text first at `wkuikit/src/test/java/com/chat/uikit/chat/manager/WKRichTextSenderTest.java:117`. If image-first is intentional for the new caption UX, the sender contract, tests, previews, and comments need to be updated together; otherwise this is a functional regression and likely CI failure.

🔴 Critical — Existing composer text is dropped when selecting images through the new caption activity.  
The prior tray contract says selected images are added to the input tray so the current composer text is merged on send. The new `addImagesToRichTextTray()` opens `WKRichTextCaptionActivity` with only image paths and channel info at `wkuikit/src/main/java/com/chat/uikit/chat/ChatPanelManager.kt:3924`, then sends only the returned caption at `wkuikit/src/main/java/com/chat/uikit/chat/ChatActivity.java:3877`. Any text already typed in the chat composer is not passed into the caption screen or merged into the outgoing rich-text message, leaving it behind and changing the documented behavior in `IConversationContext` at `wkbase/src/main/java/com/chat/base/msg/IConversationContext.java:139`.

💬 Non-blocking

🟡 Warning — Mention state in `WKRichTextCaptionActivity` can become stale.  
Selected mention UIDs are stored in a side list at `wkuikit/src/main/java/com/chat/uikit/chat/WKRichTextCaptionActivity.kt:199`, but if the user deletes the mention span/text before sending, the UID remains in the result at `wkuikit/src/main/java/com/chat/uikit/chat/WKRichTextCaptionActivity.kt:129`. This can notify users who are no longer visibly mentioned.

🟡 Warning — I could not run the focused Gradle tests in this environment because no Java Runtime is installed. The code/test mismatch above was verified by reading the checked-in test and implementation.

lml2468 (changes requested):
Gate check:

- `build` is failing: `:wkuikit:testDebugUnitTest` has 2 failing tests (`WKRichTextSenderTest.uploadSucceeds_enqueuesRichTextThenClearsInput` at line 119 and `WKRichTextSenderTest.unsafeUrlSkipped_stillEnqueuesWithSafeImagesOnly` at line 175).
- `scan-pr / osv-scan` passed.
- `analyze / CodeQL (java-kotlin)` is currently cancelled, not green.
- `check-sprint` and `welcome` are failing, treated as administrative/non-blocking per instructions.

I saw Jerry-Xin's existing changes-requested review, so I am not repeating the byte-limit bypass, rich-text block-order contract/test mismatch, existing composer text loss, or stale deleted-mention feedback here. Additional findings:

🔴 P0 - Mixed static-image + GIF selections silently drop the GIF.

`WKUIKitApplication.onBack()` returns immediately when `tryAddRichTextTray()` returns true at `wkuikit/src/main/java/com/chat/uikit/WKUIKitApplication.java:833`, so the original per-item send loop is skipped. The new `tryAddRichTextTray()` now `continue`s over GIFs at `wkuikit/src/main/java/com/chat/uikit/WKUIKitApplication.java:916`, but still returns true as long as at least one static image remains at `wkuikit/src/main/java/com/chat/uikit/WKUIKitApplication.java:924`. That means selecting `[photo.png, anim.gif]` sends only `photo.png` through the caption flow and never reaches the existing GIF send branch at `wkuikit/src/main/java/com/chat/uikit/WKUIKitApplication.java:847`. This is a silent media-loss regression. Please preserve the old contract by returning false when any selected item is a GIF/video, or explicitly send the skipped GIFs through the original path.

🟡 P1 - `@All AIs` selected in the caption modal is discarded from mention metadata.

In `WKRichTextCaptionActivity`, the `TYPE_AT_AIS` row sets `memberUID = "-2"` and inserts the visible span, but the result payload only adds UIDs when the UID is neither `-1` nor `-2` at `wkuikit/src/main/java/com/chat/uikit/chat/WKRichTextCaptionActivity.kt:199`. The activity only returns `mentionUids` and `mentionAll` at lines 129-130, and `ChatActivity` only consumes those two fields at `wkuikit/src/main/java/com/chat/uikit/chat/ChatActivity.java:3874`. So a caption can visibly contain `@All AIs` while the outgoing `WKRichTextContent` has no `mentionAis` / `WKMentionInfo.ais`, and the notification semantics are lost. Please carry a `mentionAis` flag or derive mention metadata from the current `ContactEditText` spans the same way the normal composer path does.

🟡 P1 - The luma-key video effect leaks native resources and can skip `MediaPlayer.release()`.

`ActionVideoEffectPlayer` opens `effects/action_celebrate.mp4` with `context.assets.openFd()` at `wkbase/src/main/java/com/chat/base/msgeffect/video/ActionVideoEffectPlayer.kt:62`, but the `AssetFileDescriptor` is never closed after `setDataSource()`. `LumaKeyVideoView` also creates `Surface(st)` inline at `wkbase/src/main/java/com/chat/base/msgeffect/video/LumaKeyVideoView.kt:95` and never keeps/releases that `Surface`. Finally, `release()` wraps `mediaPlayer.stop()` and `mediaPlayer.release()` in one `try` block at `wkbase/src/main/java/com/chat/base/msgeffect/video/LumaKeyVideoView.kt:188`, so if `stop()` throws during an early cleanup state, `release()` is skipped. This effect can be triggered repeatedly from chat messages, so these native resource leaks should be fixed before shipping; close the AFD, store/release the Surface, and release the MediaPlayer in a `finally`/separate guarded call.

REVIEW_STATE: CHANGES_REQUESTED

Jerry-Xin (changes requested):
This PR is in scope for the Android client, but it has blocking regressions that can drop selected media and crash on supported Android versions.

**🔴 Blocking**
- 🔴 Critical: Mixed static image + GIF selection silently drops the GIF. `tryAddRichTextTray()` skips GIF paths at `wkuikit/src/main/java/com/chat/uikit/WKUIKitApplication.java:916-918`, but still returns `true` when any static image remains at `:924`; the caller then returns early at `:833-835`, so the original per-item send loop never sends the skipped GIF. Before this change, any GIF in the selection forced the whole batch through the original path. Restore the all-or-nothing check, or explicitly send the filtered-out GIFs.

- 🔴 Critical: `GroupMdActivity` uses `String.isBlank()` on runtime strings at `wkuikit/src/main/java/com/chat/uikit/group/GroupMdActivity.java:261` and `:274`. The project has `minSdkVersion = 23` and no core library desugaring configured, so this Java 11 library method can throw `NoSuchMethodError` on older supported Android versions. Use an existing helper such as `TextUtils`/`StringUtils` or a local compatibility check.

- 🔴 Critical: The new rich-text caption modal drops the special broadcast mention with uid `-2`. `WKRichTextCaptionActivity` creates that uid at `wkuikit/src/main/java/com/chat/uikit/chat/WKRichTextCaptionActivity.kt:188-190`, but excludes it from `mentionUids` at `:205-207` and only returns `mentionAll` at `:136`. `ChatActivity` only reconstructs humans and uid mentions at `wkuikit/src/main/java/com/chat/uikit/chat/ChatActivity.java:3295-3305`, so this mention sends as plain text without notification metadata.

**💬 Non-blocking**
- 🟡 Warning: `LumaKeyVideoView` processes every frame on the UI thread with `getBitmap()`, a fresh `IntArray`, and per-pixel luma work at `wkbase/src/main/java/com/chat/base/msgeffect/video/LumaKeyVideoView.kt:111-160`. This is likely to jank chat on mid-range devices. Consider GPU shader processing or at least reusing buffers and measuring frame time.

- 🟡 Warning: I could not run the targeted Gradle compile because this environment has no Java runtime installed.

**✅ Highlights**
- The PR correctly keeps the markdown/table work within existing Markwon-based rendering patterns.
- The rich-text image render cap is a good guardrail for large messages.

wangty371 (changes requested):
PR #39 review by 二卡啦 (Erkala) @ 126e48320748

## CI Status
Build is red: `:wkuikit:testDebugUnitTest` has 2 failing tests in `WKRichTextSenderTest`. The block-ordering flip (`blocks.addAll(imageBlocks)` moved before `makeTextBlock(text)`) directly contradicts the existing test assertions at lines 117 and 175. CodeQL is cancelled. These must be fixed before merge.

## Confirmed Issues (not repeating Jerry-Xin / lml2468)

Agreeing with all prior REQUEST_CHANGES findings. Three I verified in the diff:

---

### 🔴 P0 — GIF silently dropped in mixed selection

`WKUIKitApplication.tryAddRichTextTray()` now `continue`s past GIFs but still returns `true` once at least one static image is collected. The caller at `WKUIKitApplication.onBack()` returns immediately on `true` without running the original per-item send loop. Result: `[photo.png, anim.gif]` → only `photo.png` is sent through the caption flow; `anim.gif` is silently dropped.

Fix: return `false` when any item in the selection is a GIF/video, so the original path handles the whole batch.

---

### 🟡 P1 — `@All AIs` mention discarded from result payload

In `WKRichTextCaptionActivity`, `TYPE_AT_AIS` sets `memberUID = "-2"` and inserts the visible span but the guard `if (memberUID != "-1" && memberUID != "-2")` prevents it from being added to `mentionUids`. The `result` Intent carries only `mentionUids` and `mentionAll`; there is no `mentionAis` field. `ChatActivity.richTextCaptionLauncher` reads only `pendingCaptionMentionUids` and `pendingCaptionMentionAll` — so a caption can display `@All AIs` while `WKRichTextContent` sends zero AI-mention metadata.

---

### 🟡 P1 — luma-key native resource leaks (three spots)

1. `AssetFileDescriptor` opened in `ActionVideoEffectPlayer.play()` is stored into `pendingAfd` then `pendingAfd = null` in `initMediaPlayer` — the `afd` is never closed after `setDataSource()`.
2. `Surface(st)` created inline in `initMediaPlayer` with `setSurface(Surface(st))` is never stored or released.
3. `LumaKeyVideoView.release()` puts `mediaPlayer?.stop()` and `mediaPlayer?.release()` in the same `try` block — if `stop()` throws (e.g. in IDLE state), `release()` is skipped and the MediaPlayer leaks.

Fix: close `afd` after `setDataSource`, store the `Surface` reference for `surface.release()` on teardown, and guard `release()` in a `finally` or a separate call.

---

## 🟡 P1 — MsgConfig regression: rich-text forward silently enabled / withdraw+pin silently disabled

```java
// Before
new MsgConfig(false, true, false, false, true, true)
// forward=false, withdraw=true, multipleChoice=false, reply=false, reaction=true, pin=true

// After
new MsgConfig(true)
// single-arg constructor — likely forward=true only; withdraw/pin reset to defaults
```

The comment above the line documents the argument order (`forward, withdraw, multipleChoice, reply, reaction, pin`). The old config deliberately disabled forwarding while keeping withdraw and pin active for rich text messages. The replacement `MsgConfig(true)` uses a different constructor; if it only sets `forward=true` and leaves the rest at field defaults, it enables forwarding (previously disabled) and may disable withdraw/pin (previously enabled). Please verify the MsgConfig constructor semantics and restore the intended capability set explicitly.

---

REVIEW_STATE: CHANGES_REQUESTED

lml2468 (changes requested):
Re-review after `126e48320748`. Checking my R1 findings against the fix commit.

**R1 fixes addressed:**
- ✅ Caption byte-limit bypass — fixed: `ChatActivity.richTextCaptionLauncher` now checks `isTextOverByteLimit` before `sendRichTextTray`.
- ✅ Existing composer text dropped — fixed: `ChatPanelManager.addImagesToRichTextTray` now forwards existing text via Intent extra, `WKRichTextCaptionActivity` pre-fills it.
- ⚠️ Block order test — only partially fixed: `uploadSucceeds` test updated to image-first order, but `unsafeUrlSkipped_stillEnqueuesWithSafeImagesOnly` at line 175 still asserts `rich.blocks.get(0).isText()`. This is why CI build is still red.

**R1 findings still open (all confirmed in the new diff — no changes to affected files):**

🔴 P0 — Mixed static-image + GIF selections silently drop the GIF.
`tryAddRichTextTray` skips GIFs but returns `true` → caller returns early → GIF never sent. No change in this push.

🟡 P1 — `@All AIs` caption mention metadata lost.
`WKRichTextCaptionActivity` still has no `mentionAis` in the result Intent. No change in this push.

🟡 P1 — Luma-key video effect native resource leaks.
`AssetFileDescriptor` unclosed, `Surface` unreleased, `MediaPlayer.release()` skippable. No change in this push.

**CI gate:** Build red — `WKRichTextSenderTest.unsafeUrlSkipped_stillEnqueuesWithSafeImagesOnly` still failing at line 175 (same block-order mismatch as R1, just missed in the partial test update).

REVIEW_STATE: CHANGES_REQUESTED

lyj7890 (changes requested):
❌ R2 REQUEST_CHANGES — octo-android PR #39 (126e4832)

## 修复进度

Fix commit 修了部分问题，但仍有阻塞项：

✅ caption 字节限制 — 已修
✅ composer 文本丢失 — 已修
⚠️ block 顺序 test — **半修**：`uploadSucceeds` 更新了，但 `unsafeUrlSkipped`（line 175）仍是 text-first 断言 → CI 继续红

---

## Blocking

### 🔴 P0 — GIF 静默丢弃（未修）

`tryAddRichTextTray` 遇到 GIF 走 skip 分支仍返回 `true`，调用方 early return 不走原逐条发送路径，GIF 从消息里消失，用户无感知。

**修法**：区分 "skip/不支持"（返回 `false`，让调用方走 fallback 逐条发送）和 "handled"（返回 `true`）。只要批次里有 GIF 就返回 `false`。

### 🔴 block 顺序 test 漏改（CI 继续红）

`WKRichTextSenderTest.unsafeUrlSkipped` line 175 还是 text-first 断言，与现在的 image-first block 顺序冲突。需一并修掉，否则 CI 不会绿。

---

## Non-blocking (P1)

### 🟡 @All AIs mention 元数据丢失（未修）

`WKRichTextCaptionActivity` result Intent 只序列化了 `mentionUids` + `mentionAll`，缺少 `mentionAis` 字段。`TYPE_AT_AIS` 设的 `memberUID = "-2"` 未进入出站消息元数据，AI mention 通知链断掉，属于功能静默回归。

### 🟡 luma-key 资源泄漏（未修）

三处未释放：
1. `AssetFileDescriptor` 未 close
2. 内联 `Surface(st)` 从不释放
3. `stop()` + `release()` 在同一 try 块，`stop()` 抛异常直接跳过 `release()`

建议用 `try-finally` 确保 release，或切换到 ExoPlayer 统一管理生命周期。

---

建议 @mzanalyticst 优先处理 GIF P0 和漏改的 test，再一并修 mention 元数据和 luma-key 释放，推新 commit 后 re-review。

dolphinsboy (changes requested):
## R2D2 Re-Review — octo-android PR #39 @ 126e483

**结论：REQUEST_CHANGES** ❌

---

### ✅ 已修复（本次 fix commit 确认）

- **sendRichTextTray 字节限制绕过** — 已修复
- **composer 文本丢失** — 已修复
- **uploadSucceeds 测试断言顺序** — 已修复（image-first）

---

### 🔴 仍存在 / 新引入的 Blocking 问题

**1. CI 仍然红 — `unsafeUrlSkipped` 测试未修**

`WKRichTextSenderTest.java:175` 仍然断言 `blocks.get(0).isText()`（text-first），而 `uploadSucceeds` 已改成 image-first。两个测试断言方向矛盾，其中一个必然失败。这是 CI 持续红的直接原因。

**2. GIF 静默丢弃（P0）**

`WKRichTextCaptionActivity.kt:241`：
```kotlin
if (WKFileUtils.getInstance().isGif(result.path)) continue
```
跳过 GIF 后整个 `onBack` 回调继续执行并正常返回，调用方收到 `RESULT_OK` + 非空 paths（静态图）后直接走富文本路径，原本的 GIF 逐条发送逻辑完全跳过。用户选了 GIF+静态图的组合，发出去 GIF 无声消失，无任何提示。

修法：检测到 GIF 时 toast 提示用户"暂不支持混选 GIF"，或在返回前将 GIF 路径单独回传走逐条发送路径。

**3. `String.isBlank()` 在 minSdk 23 上 crash**

`GroupMdActivity.java:261,274` 调用 Java 11 的 `String.isBlank()`，项目 `minSdkVersion=23`，未启用 `coreLibraryDesugaring`，API 23–29 设备运行时抛 `NoSuchMethodError`。

修法：换成 `TextUtils.isEmpty(s) || s.trim().isEmpty()`，或在 `build.gradle` 启用 desugaring。

**4. MsgConfig 富文本行为被静默重置**

```java
// 旧：forward禁用，withdraw可用，pin可用
new MsgConfig(false, true, false, false, true, true)
// 新：全部 true，包括 forward 被打开
new MsgConfig(true)
```

`MsgConfig(true)` 把富文本消息的转发开关从 `false` 改成了 `true`，原来明确禁止富文本转发的设计意图被抹掉。应按原意显式传参或添加专用工厂方法。

---

### 🟡 Non-blocking（前序 review 已提，仍未修）

**5. `@All AIs` mention 元数据丢失**

`WKRichTextCaptionActivity.kt:188-194,135-136`：`TYPE_AT_AIS` 点击时 `memberUID="-2"` 被排除在 `mentionUids`（`!= "-1" && != "-2"` 才加），且 result Intent 只写了 `mentionUids` + `mentionAll`，没有 `mentionAis` 字段。`@All AIs` 只有视觉呈现，出站消息里 AI mention 语义完全丢失。

**6. luma-key 三处资源泄漏**

`LumaKeyVideoView.kt:92-107,188-193`：
- `AssetFileDescriptor` 用后未 `close()`
- `Surface(st)` 内联创建，`release()` 方法里不持有引用，永不释放
- `release()` 里 `stop()` 和 `release()` 在同一 `try` 块，`stop()` 抛异常时 `release()` 被跳过

建议用 `try-finally` 分离，并持有 `Surface` 引用。

---

请修复 Blocking 问题后再提新 commit。

yujiawei (changes requested):
# Code Review — PR #39 (octo-android)

Reviewed at head `126e483207487d7028c226c346ff91497dcd0681` against `main`. This is a large change (41 files, +1781/-79) spanning 8 feature areas: Markdown tables, GIF avatars, luma-key video effect, RichText render/compose, sub-zone revoke, and SSO login. The work is generally well-structured, but there is one runtime-crash blocker and several functional regressions that should be fixed before merge.

**Verdict: CHANGES_REQUESTED** — 1 P0 (crash on the majority of the install base) + 6 P1 (functional regressions / data loss).

---

## P0 — Blocker

### 1. `String.isBlank()` crashes on Android API < 33 (readonly GROUP.md table render)
`wkuikit/src/main/java/com/chat/uikit/group/GroupMdActivity.java:261, 274`
```java
if (firstSegment.toString().isBlank()) {
...
    if (nextSegment.toString().isBlank()) continue;
```
`java.lang.String.isBlank()` was added to the Android platform at **API level 33**. This module sets `minSdkVersion = 23` (`build.gradle:42`) and has **no** `coreLibraryDesugaring` enabled (verified — no occurrences in any `.gradle`). `compileSdk = 34`, so it compiles, but on any device running Android 6.0–12 (API 23–32) these calls throw `java.lang.NoSuchMethodError` at runtime.

This is on a common path: any non-admin member opening a `GROUP.md` / thread MD whose content contains a Markdown table (with matching placeholder/table counts) hits `renderMarkdown()` → crash. Note the parallel Kotlin file `WKMarkdownViewActivity.kt` is *not* affected, because Kotlin's `isBlank()` resolves to the bundled `kotlin.text` extension, not the platform method — so the bug is Java-side only.

**Fix:** Replace with a backport-safe helper, e.g.
```java
private static boolean isBlank(CharSequence cs){
    for(int i=0;i<cs.length();i++){ if(!Character.isWhitespace(cs.charAt(i))) return false; }
    return true;
}
```
and call `isBlank(firstSegment)` / `isBlank(nextSegment)`. (Or enable `coreLibraryDesugaring`, but the helper is the minimal fix.)

---

## P1 — Functional regressions / data loss

### 2. Removing `.dontAnimate()` globally auto-animates GIF avatars in EVERY list
`wkbase/src/main/java/com/chat/base/glide/GlideUtils.java:168-172, 202-205`

`showAvatarImg` / `showRoundedAvatar` are the central avatar loaders (used by `AvatarView` across ~45 layouts and by the conversation list). Dropping `.dontAnimate()` here is not scoped to the new GIF-avatar preview — it makes every GIF avatar auto-animate everywhere: conversation rows, member grids, chat message rows. The path also keeps `.thumbnail(0.1f)`, so each GIF avatar decodes **two** concurrent `GifDrawable`s. Across a scrolling list this is a real CPU/battery/memory regression (off-screen/recycled rows keep animating).

**Fix:** Keep `.dontAnimate()` on the shared list loaders; add a dedicated animated load method (or an `animate` flag) used only by the large/detail avatar surfaces, and drop `thumbnail(0.1f)` on the animated path.

### 3. Deleting an @mention in the caption modal does not remove it (removed @s are still sent)
`wkuikit/src/main/java/com/chat/uikit/chat/WKRichTextCaptionActivity.kt:63-64, 135-136, 194, 206`

`mentionUids` is only ever `.add()`-ed and `mentionAll` is only ever set `true`; neither is ever reduced. When the user deletes an inserted @mention span from `captionEt`, the uid stays in the accumulator and is serialized at send (lines 135-136). Deleting `@all` likewise can't be retracted; re-inserting the same user double-adds. The authoritative send-key path (`ChatPanelManager.applyInputMentionsTo`) instead derives mentions from `editText.allUIDs`, which reads the current spans and self-heals.

**Fix:** At send time, derive mentions from `binding.captionEt.getAllUIDs()` exactly like `applyInputMentionsTo` (map `-1`→all/humans, `-2`→ais, else uid) and drop the `mentionUids`/`mentionAll` fields entirely.

### 4. Over-byte-limit caption silently discards ALL picked images (and mentions)
`wkuikit/src/main/java/com/chat/uikit/chat/ChatActivity.java:3877-3882`
```java
if (!TextUtils.isEmpty(text) && chatPanelManager != null && chatPanelManager.isTextOverByteLimit(text)) {
    chatPanelManager.promptTextToFile(text);   // images + mentions dropped
} else {
    sendRichTextTray(text, paths, null, null);
}
```
When the caption exceeds the byte limit, `promptTextToFile` writes only the text to a `.txt` and sends a `WKFileContent`; the picked `paths` and pending mentions are never used and are nulled when the lambda returns. Unlike the legacy tray path (where images remain on-screen for a retry), here nothing holds the images — the user picks N images + a long caption and gets a lone `.txt` message with no images.

**Fix:** In the over-limit branch, still send the images via `sendRichTextTray` (with truncated/empty caption) and offer the over-limit text separately, or keep the modal open and surface the limit so the user can shorten the caption before images are committed.

### 5. Mixed image+gif album pick silently drops the GIF
`wkuikit/src/main/java/com/chat/uikit/WKUIKitApplication.java:916-924`

The gif branch changed from `return false` to `continue`. Previously any gif in the selection routed the whole pick to the per-item loop (gif sent as sticker + images individually). Now the gif is skipped while static images are collected and handed to the tray (method returns `true`). For `[image1, gif1, image2]` the GIF is never sent anywhere — silently lost. (Pure single-gif still works; video still returns false correctly. Only the mixed image+gif case regresses.)

**Fix:** Treat gif like video — if any gif is present, `return false` so the whole pick falls back to the per-item path; or explicitly send the skipped gifs.

### 6. Caption-modal member picker is unsearchable and leaves stray query text
`wkuikit/src/main/java/com/chat/uikit/chat/WKRichTextCaptionActivity.kt:142-155, 210`

`setupMentionDetection` only opens the picker on a bare `@` and never calls `mentionAdapter.onSearch(...)`, so the list stays unfiltered as the user types. On selection it dispatches a **single** `KEYCODE_DEL` (line 210), while the reference panel deletes `1 + searchKey.length` chars. So typing `@jo` then picking a member yields a corrupted caption like `@j@Name ` and the list never filters. The feature is effectively unusable in any group large enough to require typing to find a member.

**Fix:** Feed keystrokes after `@` to `mentionAdapter.onSearch` and delete `1 + queryLength` chars on selection (mirror `ChatPanelManager`), or delete the whole pending `@…` token before `addSpan`.

### 7. Sub-zone revoke: parent member lookup uses `parentChannelType` instead of GROUP
`wkbase/src/main/java/com/chat/base/msgitem/WKChatBaseProvider.kt:1214-1230`

For `COMMUNITY_TOPIC`, the new primary branch sets `memberChannelType = channel.parentChannelType` and passes it to `getMember()`, which does an exact `(channel_id, channel_type, uid)` DB match. Parent-group members are stored under `WKChannelType.GROUP(2)` everywhere else (e.g. `WKUIChatMsgItemEntity.getMemberLookupChannelType` hardcodes GROUP; `ChatActivity` resolves the parent via `getChannel(parentGroupNo, GROUP)`). But `parentChannelType` is set to `COMMUNITY(4)` for topics (in the SDK's `ConversationManager`) or defaults to `0` when unsynced — never GROUP. So the lookup queries the wrong key, returns `null`, `isManager` stays false, and a real parent-group admin can no longer withdraw others' messages in the sub-zone. The safe fallback that forces GROUP is the `else` branch, bypassed precisely when `getChannel()` succeeds.

**Fix:** Always set `memberChannelType = WKChannelType.GROUP` for `COMMUNITY_TOPIC` parent lookups (reuse the existing `getMemberLookupChannelID/Type` helpers for consistency).

---

## P2 — Should fix (non-blocking)

**Markdown tables**
- Post-table text segments: `setTextIsSelectable(true)` is called *after* `setMovementMethod(LinkMovementMethod)`, which resets the movement method to `ArrowKeyMovementMethod` — links render but don't respond to taps. Both `GroupMdActivity.java:280-282` and `WKMarkdownViewActivity.kt:197-199`. Fix: set selectable first, movement method last.
- On placeholder/table count mismatch, the fallback renders the raw text containing `U+FFFC` placeholders as visible tofu boxes (`GroupMdActivity.java:244-250`). Strip placeholders in the fallback.
- Markdown links inside table cells are parsed but rendered as plain text, not clickable (`WKMarkdownViewActivity.kt:306-345`).

**GIF avatar**
- `isAnimatedGif` only matches GIF magic bytes; animated WebP/APNG fall through to the crop path and are silently rasterized to a static frame (`AnimatedImageUtils.java:28-39`).
- The 5MB limit is enforced only in the preview Activity, not in `AnimatedImageUtils`/upload — other entry points can bypass it (`WKAnimatedAvatarPreviewActivity.java:59-63`).

**Luma-key video effect**
- Per-frame luma keying (`getBitmap` + `getPixels` + per-pixel loop over 540×960) runs entirely on the UI thread in `onSurfaceTextureUpdated` — jank / ANR risk (`LumaKeyVideoView.kt:63-65, 111-161`). Consider a background/render thread or a GLSL shader.
- An `IntArray(540*960)` (~2MB) is allocated every frame with no reuse, and the main-thread frame work has no try/catch (`LumaKeyVideoView.kt:133`). Hoist the buffer to a field.
- `buildCenterMask()`'s return value is discarded in the rebuild path, so the mask can never actually be rebuilt on size change (`LumaKeyVideoView.kt:50, 136-141, 163-181`).
- The `Surface` wrapping the `SurfaceTexture` for `MediaPlayer` is never released (`LumaKeyVideoView.kt:95, 188-196`).
- `detect()` uses exact equality `trimmed == "[崇尚行动]"` while all other effects use `contains` — the effect won't fire if the emoji is inline with other text (`MessageEffectType.kt:34-43`). Inconsistent with siblings.
- Playback isn't paused/stopped when the activity is backgrounded (`MessageEffectManager.kt:203-210`).
- A 3.2MB uncompressed `action_celebrate.mp4` is bundled unconditionally into the APK.

**RichText render**
- Image-click `imageIndex` can misalign with `allImageUrls` when some image blocks have empty URLs (rendered list and url list are built with different filters) → opening the preview at the wrong image (`WKRichTextProvider.kt:122-135, 215-217, 237-243`). Also the popup `ivList` puts the clicked view at slot 0 regardless of `imageIndex`, unlike `WKImageProvider`, so the shared-element transition can target the wrong thumbnail.
- `MAX_RENDER_IMAGES = 20` silently drops the 21st+ image with no indicator (`WKRichTextProvider.kt:68, 131-134`).
- `scaleToFit` applies `coerceAtLeast(60dp)` to width and height independently, distorting aspect ratio for extreme sizes (`WKRichTextProvider.kt:222-229`).

**RichText compose**
- Pre-existing @mentions typed in the composer are flattened to plain text (UID lost) when launching the modal via `editText.text.toString()` (`ChatPanelManager.kt:3933-3936`). Pass the composer's uids into the modal.
- Stale comment: the block above `MsgConfig(true)` for `richText` still says forward/reply/multichoice are intentionally disabled for Phase 1, but `MsgConfig(true)` now enables them (`WKUIKitApplication.java:330-334`). Update the comment and confirm forwarding a type=14 message serializes correctly.

**SSO login**
- `loginTitleTv` and `heroLogoIv` post the same `Runnable` on the same `Handler`; multi-touch can open `ApiUrlDialog` twice, and lifting one finger cancels the other's countdown (`WKLoginActivity.java:135-158`). Low impact (hidden dev entry) — add a re-entry guard.

## nits
- Unused import `android.widget.LinearLayout` (`WKMarkdownViewActivity.kt:33`).
- Stale comment in `WKRichTextSender.java:257` still says "text first, image after" and references the removed `ImageUtils.getImageWidthAndHeightToTalk`.

---

### Summary
The P0 (`String.isBlank()` NoSuchMethodError on API 23–32) must be fixed before merge — it crashes the readonly GROUP.md path on the majority of the active install base. The six P1 items are genuine functional regressions or silent data loss (dropped images/GIFs, stale mentions, broken sub-zone admin revoke, list-wide avatar animation) and should be addressed as well. The P2/nit items are quality improvements that can follow.

lml2468 (dismissed):
R3 re-review @ 063ad83058e2. All R1/R2 blocking and non-blocking issues addressed.

**CI gate:** ✅ Build green (152 tests, 0 failures). CodeQL in progress. `check-sprint` non-blocking.

**R2 blocking issues — all fixed:**

✅ `unsafeUrlSkipped` test — assertions updated to image-first order, matching `uploadSucceeds`. Both `WKRichTextSenderTest` and `WKRichTextComposeModelTest` now consistent.

✅ GIF silent drop (P0) — `tryAddRichTextTray` now sets `hasGif = true` instead of `continue`, returns `false` when any GIF is in the selection, whole batch falls back to original per-item send path. Clean fix.

✅ `String.isBlank()` crash — `GroupMdActivity` now uses a custom `isBlank(CharSequence)` with `Character.isWhitespace()` iteration. API 1+ compatible, no desugaring needed.

✅ MsgConfig — `MsgConfig(true)` is intentional for Phase 2 (all features enabled including forward/reply). Comment updated to reflect the design decision.

**R2 non-blocking issues — all fixed:**

✅ mentionAis metadata — `WKRichTextCaptionActivity` now reads `allUIDs` from `ContactEditText` spans at send time (fixing stale-mention issue simultaneously), derives `hasAll`/`hasAis` from sentinel UIDs, passes `mentionAis` boolean in result Intent. `ChatActivity` reads the new field, sets `content.mentionAis = 1` + `mInfo.ais = true`, and filters `-1`/`-2` out of `mInfo.uids`. Correct and clean.

✅ luma-key resource leaks — three fixes: (1) `videoSurface` field stored and released in `release()`, (2) `afd.close()` after `setDataSource()`, (3) `stop()` and `release()` in separate try-catch blocks so `release()` always runs. Good.

**Additional changes in this commit (verified, no issues):**

- `WKChatBaseProvider.canWithdraw`: `memberChannelType` simplified to unconditional `WKChannelType.GROUP` for COMMUNITY_TOPIC — correct since parent of a topic is always a group.
- `WKRichTextCaptionActivity`: mention search/filter with `mentionQueryLength` tracking — new feature, works correctly with the `@` detection and backspace cleanup.
- Caption byte-limit path: now sends images with empty text + prompts text-as-file separately. Reasonable UX — doesn't lose images when text is too long.

REVIEW_STATE: APPROVED

Jerry-Xin (changes requested):
The PR is relevant to this repository, but two user-facing regressions need fixes before merge.

**🔴 Blocking**
- 🔴 Critical — The new rich-text caption modal sends without the callbacks that clear the composer. `ChatPanelManager.addImagesToRichTextTray()` copies existing input text into the modal at `wkuikit/src/main/java/com/chat/uikit/chat/ChatPanelManager.kt:3931`, then `ChatActivity.richTextCaptionLauncher` sends it with `sendRichTextTray(text, paths, null, null)` at `wkuikit/src/main/java/com/chat/uikit/chat/ChatActivity.java:3894`. `sendRichTextTray()` only clears current UI through the `onEnqueued` callback at `ChatActivity.java:3354`, so after a successful rich-text send, the original composer text remains visible and can be sent again as a duplicate plain-text message. Pass the same snapshot-aware clear/reset callbacks used by the tray path, or explicitly clear only when the current composer still matches the consumed caption.
- 🔴 Critical — Rich-text image preview can be opened with an invalid index when earlier image blocks have empty or invalid URLs. `WKRichTextProvider` builds `allImageUrls` from non-empty URLs at `wkuikit/src/main/java/com/chat/uikit/chat/provider/WKRichTextProvider.kt:122`, but increments `imageCount` for every image block at `WKRichTextProvider.kt:131-133`; `addImageBlock()` then silently returns for an empty URL at `WKRichTextProvider.kt:216-217`. A later rendered image can therefore call `showImagePopup()` with an index beyond `allImageUrls`, and `WKDialogUtils` assumes that index is valid at `wkbase/src/main/java/com/chat/base/utils/WKDialogUtils.java:739-751`. Derive preview indices only from actually rendered, non-empty image URLs.

**💬 Non-blocking**
- 🟡 Warning — `LumaKeyVideoView` does per-frame `getBitmap()`, allocates a new `IntArray`, and loops through ~518k pixels on the UI callback path at `wkbase/src/main/java/com/chat/base/msgeffect/video/LumaKeyVideoView.kt:115-164`. This is likely to cause jank on lower-end devices. Consider reusing the pixel buffer and moving the luma-key work off the main thread or to GPU/shader processing.
- 🟡 Warning — I could not run Gradle verification because this environment has no Java Runtime installed, so compile/test status remains unverified.

**✅ Highlights**
- The PR keeps rich-text block ordering covered by existing unit tests.
- GIF avatar size validation is centralized through `AnimatedImageUtils.MAX_ANIMATED_AVATAR_BYTES`.

Jerry-Xin (dismissed):
The PR is in scope for this Android chat repository, but it introduces a blocking rich text send regression.

**🔴 Blocking**
- 🔴 Critical: Rich text images/caption can be silently dropped when the chat is in reply or edit mode. `ChatPanelManager.addImagesToRichTextTray()` launches `WKRichTextCaptionActivity` and returns `true` unconditionally after attempting the launch, so the original per-image send path is suppressed. When the modal returns, `ChatActivity.sendRichTextTray()` returns `false` for `editMsg != null || replyWKMsg != null`, but `richTextCaptionLauncher` ignores that return value and provides no fallback or user error. Result: a user can select images, enter a caption, tap send, and nothing is sent. See `wkuikit/src/main/java/com/chat/uikit/chat/ChatPanelManager.kt:3924-3938`, `wkuikit/src/main/java/com/chat/uikit/chat/ChatActivity.java:3288-3290`, and `wkuikit/src/main/java/com/chat/uikit/chat/ChatActivity.java:3893-3895`. Fix by either not intercepting image selection while reply/edit is active, or by handling `sendRichTextTray(...) == false` with the previous send path or a clear UI error.

**💬 Non-blocking**
- 🟡 Warning: Existing `@mention` spans typed in the normal composer are lost when the caption modal opens. `addImagesToRichTextTray()` passes only `editText.text.toString()` into the modal and clears the composer, while `WKRichTextCaptionActivity` reconstructs plain text with no `ContactEditText` spans, so `captionEt.allUIDs` is empty unless the user re-adds mentions inside the modal. See `wkuikit/src/main/java/com/chat/uikit/chat/ChatPanelManager.kt:3931-3935` and `wkuikit/src/main/java/com/chat/uikit/chat/WKRichTextCaptionActivity.kt:79-83,131-141`.
- 🟡 Warning: The luma-key video effect processes a 540x960 bitmap and loops every pixel on the main thread for each `TextureView` frame update. That is likely to cause visible jank or dropped input during chat rendering on mid/low-end devices. See `wkbase/src/main/java/com/chat/base/msgeffect/video/LumaKeyVideoView.kt:116-169`.
- 🔵 Suggestion: `LumaKeyVideoView.initMediaPlayer()` closes the `AssetFileDescriptor` only on the success path. If `setDataSource`, `setSurface`, or `prepareAsync` throws, the descriptor is leaked. See `wkbase/src/main/java/com/chat/base/msgeffect/video/LumaKeyVideoView.kt:92-113`.

**✅ Highlights**
- The markdown table rendering path reuses the existing Markwon/table extraction infrastructure and handles mixed text/table ordering.
- The rich text block order change is covered by updated unit tests for compose and sender models.

I could not run the Gradle test command because this environment has no Java Runtime installed.

yujiawei (changes requested):
# Code Review — PR #39 (octo-android)

Thanks for this PR — it's a large, well-organized batch of fixes and features (8 distinct areas, +1878/-99). The data-model work in the RichText path is genuinely solid: the 9-image cap is enforced redundantly and correctly (model `addAll` break, `ChatPanelManager.take`, modal picker guard) with no off-by-one, drag-reorder is bounds-checked and excludes the add-cell, GIF filtering is correct on both paths, and the image-first/text-last block reorder is consistent across model + sender and the unit tests were properly updated to assert the new order (not weakened). GIF-avatar wiring, manifest entries, string EN/zh parity, and the subchannel-revoke role lookup all check out.

I'm requesting changes for **two confirmed functional/quality issues** (one of which is a regression on a common flow), plus a batch of smaller P2 cleanups noted below.

---

## P1 — must fix before merge

### 1. Composer @mention spans are lost when entering the caption modal (functional regression)

`wkuikit/.../chat/ChatPanelManager.kt:3931-3934`

```kotlin
val existingText = editText.text?.toString() ?: ""
if (existingText.isNotBlank()) {
    intent.putExtra("caption", existingText)   // flattened to a plain String
    editText.text = null                        // MyImageSpan UIDs destroyed
}
```

When a user types an @mention in the composer and **then** attaches images, the composer text is handed to the modal as a flattened `String` and the original `EditText` is cleared, so the `MyImageSpan` UID information is gone. The modal re-shows the caption via `captionEt.setText(existingCaption)` with no spans (`WKRichTextCaptionActivity.kt:79-83`), so `captionEt.allUIDs` no longer contains the mentioned uid.

On send, the only remaining re-derivation is `scanPlainTextMentions` (`ChatPanelManager.kt:1873`), which matches member names with `@([^@\\s]+)` — the regex stops at the first whitespace. Since spans are created from the **full display name** including spaces (`addSpan("@$name ", uid)`, name = `memberName`/`channelName`), any member whose name or remark contains a space (e.g. `@John Smith`) fails the equality check at `1911-1912` and the mention — plus its group notification — is **silently dropped**.

This is a regression: in the prior inline-tray flow the composer text/spans stayed in `editText`, and `applyInputMentionsTo` read `editText.allUIDs` directly, so spans were preserved. Broadcast tokens (`@所有人`/`@所有AI`) and single-word names still survive, so it's not a total break → P1, not P0.

**Common path:** "type @member, then pick a photo" (callers `ChatActivity.java:3256`, `WKUIKitApplication.java:922`). The `RESULT_CANCELED` restore path (`ChatActivity.java:3900-3905`) also re-sets the caption as plain text, so spans are lost on cancel too.

**Fix:** carry the mention identity across the modal instead of relying on a plain-text re-scan — e.g. pass the composer's `allUIDs`/`allEntity` as extras and re-create spans in the modal (and forward them in the result), or seed `pendingCaptionMention*` from `editText.allUIDs` at launch time and merge with modal-added mentions.

### 2. Full-screen video effect does per-frame luma-key on the UI thread

`wkbase/.../msgeffect/video/LumaKeyVideoView.kt:64-66, 116-170`

```kotlin
override fun onSurfaceTextureUpdated(st: SurfaceTexture) {
    processFrame()   // runs on the UI thread
}
```

`SurfaceTextureListener` callbacks fire on the view's owning (UI) thread, and there's no Handler/coroutine hop. For **every** decoded frame, `processFrame()` does a synchronous `TextureView.getBitmap()` GPU→CPU readback, then `applyLumaKey()` loops over 540×960 ≈ 518k pixels in float math, with two full `getPixels`/`setPixels` array copies — all on the main thread. On mid/low-end devices this will cause visible jank/dropped frames and can trip an ANR; on capable devices it degrades to frame drops. The overlay is transient (≤6 s), so it's a quality defect rather than a guaranteed crash → P1.

**Fix:** move the readback + luma-key pass to a worker/`HandlerThread` (or replace the CPU keying with a GLSL shader on a `GLSurfaceView`/`SurfaceTexture`), and post only the finished bitmap back to the render view.

---

## P2 — recommended, not blocking

**RichText / compose**
- **Over-byte-limit caption applies @mentions to an empty-text RichText.** `ChatActivity.java:3890-3892` sends `sendRichTextTray("", paths, …)` while `pendingCaptionMention*` is still populated, so the @-notification lands on an images-only, empty-caption message while the @-bearing text is moved to a separate `.txt`. (No duplicate notification — the file path carries no mentions — but the mention should follow the text.) Clear `pendingCaptionMention*` before the empty send, or route it to the text-to-file message.
- **Composer text not persisted before launching the modal** (`ChatPanelManager.kt:3931-3937`): `editText` is nulled with no draft saved; recovery depends entirely on the modal returning `RESULT_CANCELED` *with* an Intent. On OS process death while the modal is foreground, the unsent caption is lost. Persist a draft before clearing.
- **Image-preview return transition uses the wrong view index** (`WKRichTextProvider.kt:240-247`): `ivList` is built as `[clickedView, null, null, …]` but `imageIndex` is passed as the start index, so the swipe-update / close-return shared-element transition points at `null`/the wrong slot. (The open transition is fine — it uses the explicit `imageView` param.) Size `ivList` to the image count and set `ivList[imageIndex] = imageView`.
- **Dead mention state in the modal** (`WKRichTextCaptionActivity.kt:63-64, 217, 229`): `mentionUids`/`mentionAll` are written but never read (send recomputes from `captionEt.allUIDs`); `mentionAll` is also never reset. Remove them to avoid a future-maintainer trap.

**Video effect**
- **`buildCenterMask()` return value discarded** (`LumaKeyVideoView.kt:51, 145-150`): `centerMask` is `by lazy`, but on first access `maskW/maskH` are still 0 so it builds at the fallback size; `applyLumaKey` then calls `buildCenterMask()` again and **throws away the result** (the `lazy val` can't be reassigned), so the mask used is the fallback-sized one and the rebuild is dead work. Make the mask a reassignable field and actually use the rebuilt array.
- **3.2 MB `action_celebrate.mp4` ships in every APK/AAB** for one optional effect, in `src/main/assets` (uncompressed). Consider on-demand download or an asset pack if size matters.

**Markdown / GROUP.md**
- **Table-cell hyperlinks are dropped to plain text** in `GroupMdActivity`/`WKMarkdownViewActivity` even though `WKTablePlugin` extracts per-cell link spans (`WKTableData.kt:21-30`) and the chat renderer (`WKTextProvider.kt:469-494`) makes them clickable. Mirror that ClickableSpan handling for consistency.
- **Remote `.md` is parsed/rendered on the UI thread with no size cap** (`WKMarkdownViewActivity.kt:127-142`); a large remote file can jank/ANR. Cap size and/or parse off-thread.
- **Hard-coded light-mode table colors** (`#F0F0F0`/`#FAFAFA`/`#333333`) ignore dark mode, unlike the code-block path which already detects it.
- **Read-only table cards aren't restored on config change** (`GroupMdActivity` `renderMarkdown` is only called from the network callback); rotation re-fetches and loses scroll/selection. Standard non-`configChanges` behavior — minor.

**Login / misc**
- **Avatar result handler missing the empty-path guard** that the other three handlers have (`MyHeadPortraitActivity.java` `chooseResultLac`) — defensive consistency; not crash-on-common-path. (Note: pre-existing, only adjacent to this PR's changes.)
- **GIF detection does 6-byte file I/O on the UI thread** (`AnimatedImageUtils.isAnimatedGif`, called from the picker `onBack`) — strict-mode flaggable; negligible in practice.
- **Detection is GIF-only**; animated WebP/APNG fall through to crop and lose animation. Given the now-global avatar animation, display will animate formats upload can't preserve — document the GIF-only scope or extend detection.
- **The hidden long-press → API-switch dialog is not debug-gated** and is now triggerable from two views (title + logo). This is pre-existing behavior the PR widens, and the dialog enforces HTTPS + a reachability probe, so it's low-risk — but consider `BuildConfig.DEBUG`-gating the entry. The two `OnTouchListener` bodies are also byte-for-byte identical; extract a shared handler.

---

## Notes (verified, not defects)

- **Removing `.dontAnimate()` in `GlideUtils` is global** — it enables GIF animation at every avatar render site (conversation list, member grids, etc.), not just the new profile/preview screens. This is deliberate and central to the feature, so it's not a bug; flagging it as a **product/perf trade-off** worth a conscious sign-off (many animated avatars scrolling concurrently = decode + invalidate cost). Optionally pause off-screen animation in lists.
- The subchannel-revoke role lookup (`WKChatBaseProvider.kt:1214-1233`) is correct and fail-closed, and is strictly more robust than the sibling pattern elsewhere. This is a client-side UI affordance; the server remains the authority on revoke permission.

Jerry-Xin (approved):
Round 5 re-review. Both R4 blocking items from yujiawei are fixed in 1fbb723:

1. **@mention preservation**: `ChatPanelManager` now extracts `allUIDs` from the composer and forwards them via `existingMentionUids` intent extra to the caption activity. `WKRichTextCaptionActivity` merges existing UIDs with any new modal UIDs on send, so mention metadata survives the modal transition.
2. **Luma-key off UI thread**: `LumaKeyVideoView` now runs `applyLumaKey()` on a dedicated `HandlerThread("LumaKeyWorker")` with double-buffered bitmaps (`captureBitmap`/`displayBitmap`), a `@Volatile processing` guard against frame buildup, and proper cleanup via `quitSafely()` in `release()`.

No regressions from R4 fixes (composer cleanup on modal launch, `renderedImageIndex` for valid URLs only, pixel buffer reuse).

**💬 Non-blocking**
- The merged UIDs include `existingMentionUids` unconditionally. If a user deletes an @mention's visible text in the caption editor, the UID is still sent with the message. This is a minor edge case (requires user to actively edit out a pre-filled mention) and is strictly better than the R4 state where mentions were lost entirely. A future enhancement could re-derive mention metadata from the final caption spans instead of merging.
- RichText image preview passes `ivList[0]` as the source view for all image indices, which could cause misaligned page transitions for multi-image previews (`WKRichTextProvider.kt:240-245`).

**✅ Highlights**
- Clean double-buffer pattern for luma-key with proper threading.
- Build CI green (152 tests).

Jerry-Xin (approved):
Round 5 re-review. Both R4 blocking items from yujiawei are fixed in 1fbb723:

1. **@mention preservation**: `ChatPanelManager` now extracts `allUIDs` from the composer and forwards them via `existingMentionUids` intent extra to the caption activity. `WKRichTextCaptionActivity` merges existing UIDs with any new modal UIDs on send, so mention metadata survives the modal transition.
2. **Luma-key off UI thread**: `LumaKeyVideoView` now runs `applyLumaKey()` on a dedicated `HandlerThread("LumaKeyWorker")` with double-buffered bitmaps (`captureBitmap`/`displayBitmap`), a `@Volatile processing` guard against frame buildup, and proper cleanup via `quitSafely()` in `release()`.

No regressions from R4 fixes (composer cleanup on modal launch, `renderedImageIndex` for valid URLs only, pixel buffer reuse).

**💬 Non-blocking**
- The merged UIDs include `existingMentionUids` unconditionally. If a user deletes an @mention's visible text in the caption editor, the UID is still sent with the message. This is a minor edge case (requires user to actively edit out a pre-filled mention) and is strictly better than the R4 state where mentions were lost entirely. A future enhancement could re-derive mention metadata from the final caption spans instead of merging.
- RichText image preview passes `ivList[0]` as the source view for all image indices, which could cause misaligned page transitions for multi-image previews (`WKRichTextProvider.kt:240-245`).

**✅ Highlights**
- Clean double-buffer pattern for luma-key with proper threading.
- Build CI green (152 tests).

Jerry-Xin (approved):
Round 5 re-review. Both R4 blocking items from yujiawei are fixed in 1fbb723:

1. **@mention preservation**: `ChatPanelManager` now extracts `allUIDs` from the composer and forwards them via `existingMentionUids` intent extra to the caption activity. `WKRichTextCaptionActivity` merges existing UIDs with any new modal UIDs on send, so mention metadata survives the modal transition.
2. **Luma-key off UI thread**: `LumaKeyVideoView` now runs `applyLumaKey()` on a dedicated `HandlerThread("LumaKeyWorker")` with double-buffered bitmaps (`captureBitmap`/`displayBitmap`), a `@Volatile processing` guard against frame buildup, and proper cleanup via `quitSafely()` in `release()`.

No regressions from R4 fixes (composer cleanup on modal launch, `renderedImageIndex` for valid URLs only, pixel buffer reuse).

**💬 Non-blocking**
- The merged UIDs include `existingMentionUids` unconditionally. If a user deletes an @mention's visible text in the caption editor, the UID is still sent with the message. This is a minor edge case (requires user to actively edit out a pre-filled mention) and is strictly better than the R4 state where mentions were lost entirely. A future enhancement could re-derive mention metadata from the final caption spans instead of merging.
- RichText image preview passes `ivList[0]` as the source view for all image indices, which could cause misaligned page transitions for multi-image previews (`WKRichTextProvider.kt:240-245`).

**✅ Highlights**
- Clean double-buffer pattern for luma-key with proper threading.
- Build CI green (152 tests).

lml2468 (approved):
R5 re-review @ 1fbb723995a8. Two new commits since my R3 APPROVED (063ad83). Both Steve R3 blocking findings resolved + additional polish.

**CI:** ✅ build green, ✅ CodeQL green, ✅ osv-scan green. (`check-sprint` administrative.)

**Steve R3 blocking issues — resolved in d0382d8:**

✅ Duplicate send risk after caption modal — fix is two-sided:
- `ChatPanelManager.addImagesToRichTextTray` now clears composer + passes existing mentions via `existingMentionUids` Intent extra before launching modal.
- `ChatActivity.richTextCaptionLauncher` restores caption text from Intent on `RESULT_CANCELED` (modal's `returnCaptionOnCancel()` always returns the caption Intent on cancel/back/empty-image).

✅ Image preview index OOB — `WKRichTextProvider` now uses `renderedImageIndex` (only increments for blocks whose `showUrl` is non-empty) and builds `allImageUrls` incrementally inline. `imageCount` stays for the MAX_RENDER_IMAGES cap, but never feeds the popup index.

**1fbb723 additional improvements:**

✅ LumaKey double-buffer — frame pixel processing now runs on `HandlerThread("LumaKeyWorker")`, swaps `captureBitmap` ↔ `displayBitmap` on main thread, reuses `pixelBuffer`. `processing` volatile flag prevents queue overflow. Worker thread properly `quitSafely()`'d in `release()`. Addresses the R3 non-blocking UI-thread jank concern.

✅ Mention preservation across modal — `LinkedHashSet` merge of `existingMentionUids` + modal-side `allUIDs` (preserves order, dedupes). Removes the stale `mentionUids`/`mentionAll` local mirror, derives everything from `ContactEditText` spans at send time. Cleaner.

**Known non-blocking limitations (Jerry-Xin already flagged, agree with non-blocking call):**

🟡 Mention span lost on modal cancel — `editText.text = null` strips spans before modal launches; cancel restores plain text only, mention spans not reconstructed. User's pre-existing `@张三 hello` becomes plain text without notification metadata if they open then cancel. Worth a follow-up but acceptable for this scope.

🟡 Multi-image preview source view — `ivList[0]` always gets the clicked `imageView` regardless of `imageIndex`. Page transition animation may start from wrong slot when clicking non-first images. Cosmetic.

After 4 rounds (R1→R2→R3→R5), all blocking issues across all reviewers resolved. LGTM 🚀

REVIEW_STATE: APPROVED

yujiawei (approved):
# Code Review — PR #39 (octo-android)

## Verdict: Approve

A large, multi-feature PR (≈1,910 additions across 42 files): Markdown table rendering, GIF animated avatars, a full-screen luma-key video effect, subtopic-admin message withdraw, an SSO logo long-press API switcher, and a full-screen RichText caption modal. I reviewed each feature cluster independently against the checked-out source at the head SHA (verifying cross-file method signatures, Intent contracts, lifecycle, and threading), and adversarially re-verified every candidate blocking issue.

**No P0/P1 blocking defects were confirmed.** All findings below are P2 (minor bug / maintainability / UX). The most important one to address is the video-effect bitmap-recycle race; the rest are polish or follow-ups. Approving on the basis that none of these block merge — please consider folding the higher-value items into a follow-up.

---

## Findings (all P2 — non-blocking)

### Luma-key video effect (highest-value to fix)

**1. `release()` recycles bitmaps without joining the worker thread (concurrency).**
`LumaKeyVideoView.release()` runs on the main thread and calls `workerThread?.quitSafely()` immediately followed by `captureBitmap?.recycle()` / `displayBitmap?.recycle()`. `quitSafely()` does not block, so a worker task mid-`applyLumaKey()` (`getPixels`/`setPixels` over a 540×960 buffer) can be reading/writing the very bitmap that is being recycled. The dominant outcome is benign — the resulting `IllegalStateException` is swallowed by the empty `catch (_: Exception) {}` — but recycling a bitmap concurrently with native pixel access is undefined behaviour and can crash natively on some Android versions. This teardown path runs on every effect dismissal (Activity destroy and the 6 s timeout) while frames are still in flight.
*Fix:* join the worker before recycling (`workerThread?.quitSafely(); workerThread?.join()`), or set a `@Volatile released` flag checked inside the worker runnable and post the recycle onto `workerHandler` so it runs after any in-flight frame.

**2. `AssetFileDescriptor` (`pendingAfd`) leaked if the view is released before the SurfaceTexture is available.**
`pendingAfd` is only closed inside `initMediaPlayer()`. If the view is released before `onSurfaceTextureAvailable` fires (fast Activity finish, or timeout on a slow device), `initMediaPlayer` never runs and the fd is never closed. A new fd is opened on every `[崇尚行动]` play.
*Fix:* in `release()`, add `try { pendingAfd?.close() } catch (_: Exception) {}; pendingAfd = null`, and also close it in the `catch` branch of `initMediaPlayer`.

**3. Minor robustness nits (optional):** a second `ActionVideo` arriving during the 600 ms fade-out is silently dropped (`isPlaying` stays true until animation end); a late `onSurfaceTextureUpdated` after `release()` can leave `processing = true` (harmless, latent if the view were ever reused). Both are acceptable as-is; worth a comment.

### Markdown tables

**4. Size-mismatch fallback renders raw U+FFFC placeholder glyphs.** When `placeholderPositions.size != tableDataList.size`, both `WKMarkdownViewActivity.kt` and `GroupMdActivity.java` fall back to `setText(rendered)`, where `rendered` still contains the `WKTablePlugin.TABLE_PLACEHOLDER` (U+FFFC) characters, which display as tofu/box glyphs. Edge case only.
*Fix:* strip `TABLE_PLACEHOLDER` from the text before `setText` on the fallback path.

**5. OkHttp call not cancelled on destroy.** `WKMarkdownViewActivity.loadMarkdown()` builds a fresh `OkHttpClient` per load and enqueues a callback capturing the Activity; the `Call` is never cancelled in `onDestroy()`, retaining the Activity until the request completes (up to the 30 s timeout). Minor leak + wasteful client allocation.
*Fix:* hold and cancel the `Call` in `onDestroy()`; reuse a shared `OkHttpClient`.

**6. Missing `values-en` translation for `group_md_placeholder`.** Defined only in `values/strings.xml` (zh); English-locale admins editing GROUP.md see Chinese hint text. (All other new strings have EN entries.)

### GIF avatar

**7. Removing `.dontAnimate()` in `GlideUtils` is app-wide, not upload-scoped.** It is dropped from both `showRoundedAvatar()` and `showAvatarImg()` — the shared avatar loaders used by `AvatarView` everywhere (conversation list, member lists). Every GIF avatar now animates in fast-scrolling RecyclerViews, which can cost scroll smoothness/battery on low-end devices and reverts the earlier perf tuning. Intended direction, but please confirm with QA on low-end hardware, or scope animation to detail/profile surfaces.

**8. `isAnimatedGif()` only checks magic bytes** — a static single-frame `.gif` also passes and skips cropping. Functionally harmless (render-time circle crop), but the name overstates the check (it is effectively `isGif`).

**9. 5 MB limit is client-only** (enforced in the preview activity). Fine as a UX gate; confirm the avatar endpoint enforces its own size/frame limits server-side.

### RichText caption modal

The Intent contract (sender → modal → result launcher) matches exactly on all keys; the three-state mention merge (`existingMentionUids` + modal UIDs, `-1`/`-2` sentinels → `mentionHumans`/`mentionAis`) is consistent with no double-count; the image-first/text-last block reorder is applied consistently across `previewBlocks`, `WKRichTextSender`, and the provider, with tests updated.

**10. Over-limit caption text can be lost.** On the `isTextOverByteLimit` branch, images are sent with empty text and `promptTextToFile(text)` is shown; the input box was already cleared at launch. If the user cancels that dialog (its negative button is `null`), the >10 KB caption is gone with no undo. Rare, but it is silent data loss.
*Fix:* restore the caption to the composer if the to-file dialog is cancelled.

**11. Image-preview shared-element transition mis-sources for images after the first (cosmetic).** In `WKRichTextProvider.addImageBlock`, `ivList` parks the single non-null `ImageView` at index 0 while opening the popup at `imageIndex>0`, so the enter/exit transition animates from the wrong thumbnail. No crash/NPE (null entries are tolerated).
*Fix:* place the clicked view at `ivList[imageIndex]` to match position.

**12. Inline RichText tray code is now dead.** `addImagesToRichTextTray` only launches the modal; nothing populates `richTextTray` anymore, so `flushRichTextTraySend()`, `WKRichTextTrayAdapter`, the tray `ItemTouchHelper`, and the related `TextWatcher` branches are unreachable in the shipping flow. Maintenance hazard / two divergent paths for one feature.
*Fix:* delete the inline-tray machinery, or comment + ticket it as scaffolding.

### Subtopic withdraw — verified correct

`WKChatBaseProvider.canWithdraw` resolves `COMMUNITY_TOPIC` → parent group via `channel.parentChannelID` (fallback `substringBefore("____")`) and looks up role as `GROUP`. The `____` format matches every other resolver in the repo; the pathological case (`channel == null` and no `____`) fails closed (withdraw denied, not wrongly granted). Only note: this is the 5th+ copy of topic→parent resolution — consider a shared helper. **No defect.**

### SSO logo long-press — verified safe

The added `heroLogoIv` touch listener is identical to the existing `loginTitleTv` one; `longPressHandler`/`Runnable` are declared, initialized, and cleaned up in `finish()` (no leak); `sso_hint` exists in both zh and en. **No defect.** One pre-existing concern, only widened here: the API-base-URL switcher dialog is gated solely by an undocumented 1.5 s long-press with no `BuildConfig.DEBUG`/role check, and this PR adds a second trigger surface (the logo) on the production login screen. Consider gating both triggers behind `BuildConfig.DEBUG` or the existing `can_modify_api_url` flag.

---

## Summary

Solid, well-tested work across six features. Cross-file contracts (Intent extras, mention merge, block ordering) are correct, and the two clusters most prone to subtle bugs (subtopic withdraw, SSO long-press) verified clean. The video-effect bitmap-recycle race and AFD leak are the items most worth fixing soon, but neither is a deterministic crash, so they don't block merge. Approving.

dolphinsboy:
Closing as requested.

dolphinsboy (approved):

PR #42, opened 2026-06-06
feat(richtext): mention highlighting + crash fixes (v1.3.2)

mzanalyticst:
## Summary

- **图文混排 @mention 高亮 + 点击跳转**：发送侧编码 `mention.entities`，接收侧 `WKRichTextProvider` 渲染高亮可点击的 @mention（含 @所有人/@所有AI 广播标签），点击跳转个人名片页面，对齐 iOS 端行为
- **Bugly 崩溃修复（7 项）**：ContactEditText 粘贴 NPE、ChatActivity refreshMsg 越界、OPPO Push ClassNotFound、RemindMemberAdapter ClassCast、ProhibitWordDB 连接池关闭、UserDetailActivity/MyHeadPortraitActivity 未注册 Launcher
- **DB 性能优化**：批量查询替代循环单查（`getMsgExtrasForChannels`）、`getMemberOfFromIfCached` 避免刷新回调触发主线程 DB 查询、DB 方法统一 try-catch 防连接池关闭崩溃
- **ProGuard 规则**：补充 `thread.msgmodel` keep 规则防止 release 包混淆崩溃
- **版本号**：1.3.1 (32) → 1.3.2 (33)

## Test plan

- [ ] 群聊中发送带 @mention 的图文混排消息，确认 @mention 高亮显示（紫色+粗体）
- [ ] 点击高亮 @mention 跳转到个人名片页面
- [ ] @所有人/@所有AI 只高亮不可点击
- [ ] 纯文本消息 @mention 仍正常工作（无回归）
- [ ] 无 mention 的图文混排消息正常渲染
- [ ] 剪贴板为空时粘贴不崩溃
- [ ] OPPO 设备推送初始化失败不崩溃
- [ ] 会话列表加载、频道刷新无 ANR
- [ ] release 包混淆后正常运行

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -37,8 +37,8 @@ android {
         applicationId appId
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
-        versionCode 32
-        versionName "1.3.1"
+        versionCode 33
+        versionName "1.3.2"
 
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
         vectorDrawables {
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -168,6 +168,7 @@
 -keep class com.chat.uikit.chat.msgmodel.** { *; }
 -keep class com.chat.uikit.enity.** { *; }
 -keep class com.chat.uikit.thread.service.entity.** { *; }
+-keep class com.chat.uikit.thread.msgmodel.** { *; }
 -keep class com.chat.uikit.sidebar.SidebarItemEntity { *; }
 -keep class com.chat.uikit.sidebar.SidebarSyncResponse { *; }
 -keep class com.chat.uikit.sidebar.FollowModel$FollowSortItem { *; }
--- a/wkbase/src/main/java/com/chat/base/ui/components/ContactEditText.java
+++ b/wkbase/src/main/java/com/chat/base/ui/components/ContactEditText.java
@@ -328,9 +328,9 @@ public class ContactEditText extends AppCompatAutoCompleteTextView {
     public boolean onTextContextMenuItem(int id) {
         if (id == android.R.id.paste) {
             ClipboardManager cm = (ClipboardManager) getContext().getSystemService(CLIPBOARD_SERVICE);
-            assert cm != null;
+            if (cm == null) return super.onTextContextMenuItem(id);
             ClipData data = cm.getPrimaryClip();
-            assert data != null;
+            if (data == null || data.getItemCount() == 0) return super.onTextContextMenuItem(id);
             ClipData.Item item = data.getItemAt(0);
             String editContent = "";
             if (item.getText() != null)
--- a/wkim/src/main/java/com/xinbida/wukongim/db/ConversationDbManager.java
+++ b/wkim/src/main/java/com/xinbida/wukongim/db/ConversationDbManager.java
@@ -454,7 +454,8 @@ public class ConversationDbManager {
                 }
                 cursor.close();
             }
-        } catch (Exception ignored) {
+        } catch (Exception e) {
+            WKLoggerUtils.getInstance().e(TAG, "queryMsgExtraWithChannel error");
         }
         return msgExtra;
     }
@@ -466,15 +467,20 @@ public class ConversationDbManager {
 
     private List<WKConversationMsgExtra> queryWithExtraChannelIds(List<String> channelIds) {
         List<WKConversationMsgExtra> list = new ArrayList<>();
-        try (Cursor cursor = WKIMApplication.getInstance().getDbHelper().select(conversationExtra, "channel_id in (" + WKCursor.getPlaceholders(channelIds.size()) + ")", channelIds.toArray(new String[0]), null)) {
-            if (cursor == null) {
-                return list;
+        // SQLite 变量上限 999，分片查询避免超限
+        int chunkSize = 500;
+        for (int start = 0; start < channelIds.size(); start += chunkSize) {
+            int end = Math.min(start + chunkSize, channelIds.size());
+            List<String> chunk = channelIds.subList(start, end);
+            try (Cursor cursor = WKIMApplication.getInstance().getDbHelper().select(conversationExtra, "channel_id in (" + WKCursor.getPlaceholders(chunk.size()) + ")", chunk.toArray(new String[0]), null)) {
+                if (cursor == null) continue;
+                for (cursor.moveToFirst(); !cursor.isAfterLast(); cursor.moveToNext()) {
+                    WKConversationMsgExtra extra = serializeMsgExtra(cursor);
+                    list.add(extra);
+                }
+            } catch (Exception e) {
+                WKLoggerUtils.getInstance().e(TAG, "queryWithExtraChannelIds chunk error");
             }
-            for (cursor.moveToFirst(); !cursor.isAfterLast(); cursor.moveToNext()) {
-                WKConversationMsgExtra extra = serializeMsgExtra(cursor);
-                list.add(extra);
-            }
-        } catch (Exception ignored) {
         }
         return list;
     }
--- a/wkim/src/main/java/com/xinbida/wukongim/db/ConversationDbManager.java
+++ b/wkim/src/main/java/com/xinbida/wukongim/db/ConversationDbManager.java
@@ -443,17 +443,25 @@ public class ConversationDbManager {
 
     public WKConversationMsgExtra queryMsgExtraWithChannel(String channelID, byte channelType) {
         WKConversationMsgExtra msgExtra = null;
-        String selection = "channel_id=? and channel_type=?";
-        Cursor cursor = WKIMApplication
-                .getInstance()
-                .getDbHelper().select(conversationExtra, selection, new String[]{channelID, String.valueOf(channelType)}, null);
-        if (cursor != null) {
-            if (cursor.moveToFirst()) {
-                msgExtra = serializeMsgExtra(cursor);
+        try {
+            String selection = "channel_id=? and channel_type=?";
+            Cursor cursor = WKIMApplication
+                    .getInstance()
+                    .getDbHelper().select(conversationExtra, selection, new String[]{channelID, String.valueOf(channelType)}, null);
+            if (cursor != null) {
+                if (cursor.moveToFirst()) {
+                    msgExtra = serializeMsgExtra(cursor);
+                }
+                cursor.close();
             }
-            cursor.close();
+        } catch (Exception ignored) {
         }
         return msgExtra;
+    }
+
+    public List<WKConversationMsgExtra> queryMsgExtrasForChannelIds(List<String> channelIds) {
+        if (channelIds == null || channelIds.isEmpty()) return new ArrayList<>();
+        return queryWithExtraChannelIds(channelIds);
     }
 
     private List<WKConversationMsgExtra> queryWithExtraChannelIds(List<String> channelIds) {
@@ -466,6 +474,7 @@ public class ConversationDbManager {
                 WKConversationMsgExtra extra = serializeMsgExtra(cursor);
                 list.add(extra);
             }
+        } catch (Exception ignored) {
         }
         return list;
     }
--- a/wkim/src/main/java/com/xinbida/wukongim/db/ReminderDBManager.java
+++ b/wkim/src/main/java/com/xinbida/wukongim/db/ReminderDBManager.java
@@ -74,6 +74,8 @@ public class ReminderDBManager {
         if (dbHelper == null) return false;
         try (Cursor cursor = dbHelper.rawQuery(sql, new Object[]{prefix, reminderType})) {
             return cursor != null && cursor.moveToFirst();
+        } catch (Exception ignored) {
+            return false;
         }
     }
 
--- a/wkim/src/main/java/com/xinbida/wukongim/db/ReminderDBManager.java
+++ b/wkim/src/main/java/com/xinbida/wukongim/db/ReminderDBManager.java
@@ -74,7 +74,8 @@ public class ReminderDBManager {
         if (dbHelper == null) return false;
         try (Cursor cursor = dbHelper.rawQuery(sql, new Object[]{prefix, reminderType})) {
             return cursor != null && cursor.moveToFirst();
-        } catch (Exception ignored) {
+        } catch (Exception e) {
+            WKLoggerUtils.getInstance().e(TAG, "hasUndoneReminderWithChannelPrefix error");
             return false;
         }
     }
--- a/wkim/src/main/java/com/xinbida/wukongim/entity/WKMsg.java
+++ b/wkim/src/main/java/com/xinbida/wukongim/entity/WKMsg.java
@@ -234,6 +234,15 @@ public class WKMsg implements Parcelable {
         return memberOfFrom;
     }
 
+    /**
+     * 返回已缓存的发送者成员信息，不触发数据库查询。
+     * 用于主线程批量刷新场景（onRefreshChannel / onRefreshChannelMember），
+     * 避免对未加载的消息发起同步 DB 查询导致 ANR。
+     */
+    public WKChannelMember getMemberOfFromIfCached() {
+        return memberOfFrom;
+    }
+
     public void setMemberOfFrom(WKChannelMember memberOfFrom) {
         this.memberOfFrom = memberOfFrom;
     }
--- a/wkim/src/main/java/com/xinbida/wukongim/manager/ChannelManager.java
+++ b/wkim/src/main/java/com/xinbida/wukongim/manager/ChannelManager.java
@@ -77,6 +77,16 @@ public class ChannelManager extends BaseManager {
     }
 
     /**
+     * 仅从内存缓存返回频道信息，不触发 DB 查询。
+     * 用于主线程渲染路径（RecyclerView bind），避免 cache miss 时
+     * 同步 DB 查询导致 ANR。
+     */
+    public WKChannel getChannelIfCached(String channelID, byte channelType) {
+        if (TextUtils.isEmpty(channelID)) return null;
+        return channelInfoCache.get(cacheKey(channelID, channelType));
+    }
+
+    /**
      *  · 慢路径：cache miss 时走 synchronized + list 线性扫描 + DB fallback。
      *
      * <p>保留 {@link #wkChannelList} 的原有语义是因为别的路径（比如
--- a/wkim/src/main/java/com/xinbida/wukongim/manager/ConversationManager.java
+++ b/wkim/src/main/java/com/xinbida/wukongim/manager/ConversationManager.java
@@ -348,6 +348,19 @@ public class ConversationManager extends BaseManager {
         return ConversationDbManager.getInstance().queryMsgExtraWithChannel(channelID, channelType);
     }
 
+    /**
+     * 批量查询会话 extras（草稿等），一次 DB 查询替代 N 次单条查询。
+     * 返回 Map 以 channelID:channelType 为复合 key。
+     */
+    public java.util.Map<String, WKConversationMsgExtra> getMsgExtrasForChannels(java.util.List<String> channelIds) {
+        java.util.Map<String, WKConversationMsgExtra> map = new java.util.HashMap<>();
+        java.util.List<WKConversationMsgExtra> list = ConversationDbManager.getInstance().queryMsgExtrasForChannelIds(channelIds);
+        for (WKConversationMsgExtra extra : list) {
+            map.put(extra.channelID + ":" + extra.channelType, extra);
+        }
+        return map;
+    }
+
     public void updateMsgExtra(WKConversationMsgExtra extra) {
         boolean result = ConversationDbManager.getInstance().insertOrUpdateMsgExtra(extra);
         if (result) {
--- a/wkpush/src/main/java/com/chat/push/WKPushApplication.java
+++ b/wkpush/src/main/java/com/chat/push/WKPushApplication.java
@@ -143,7 +143,12 @@ public class WKPushApplication {
     }
 
     private void initOPPO() {
-        HeytapPushManager.init(mContext.get(), true);
+        try {
+            HeytapPushManager.init(mContext.get(), true);
+        } catch (Throwable e) {
+            Log.w("Push", "OPPO push init failed", e);
+            return;
+        }
         //  P-11: AppExecutors.io() 取代 new Thread()
         AppExecutors.io().execute(() -> HeytapPushManager.register(mContext.get(), BuildConfig.OPPO_APP_KEY, BuildConfig.OPPO_APP_SECRET, new ICallBackResultService() {
             @Override
--- a/wkuikit/src/main/java/com/chat/uikit/chat/ChatActivity.java
+++ b/wkuikit/src/main/java/com/chat/uikit/chat/ChatActivity.java
@@ -1562,11 +1562,11 @@ public class ChatActivity extends SwipeBackActivity implements IConversationCont
                         chatAdapter.getData().get(i).wkMsg.setFrom(channel);
                         isRefresh = true;
                     }
-                    if (chatAdapter.getData().get(i).wkMsg.getMemberOfFrom() != null && chatAdapter.getData().get(i).wkMsg.getMemberOfFrom().memberUID.equals(channel.channelID) && channel.channelType == WKChannelType.PERSONAL) {
-                        chatAdapter.getData().get(i).wkMsg.getMemberOfFrom().memberRemark = channel.channelRemark;
-                        chatAdapter.getData().get(i).wkMsg.getMemberOfFrom().memberName = channel.channelName;
-                        chatAdapter.getData().get(i).wkMsg.getMemberOfFrom().memberAvatar = channel.avatar;
-                        chatAdapter.getData().get(i).wkMsg.getMemberOfFrom().memberAvatarCacheKey = channel.avatarCacheKey;
+                    if (chatAdapter.getData().get(i).wkMsg.getMemberOfFromIfCached() != null && chatAdapter.getData().get(i).wkMsg.getMemberOfFromIfCached().memberUID.equals(channel.channelID) && channel.channelType == WKChannelType.PERSONAL) {
+                        chatAdapter.getData().get(i).wkMsg.getMemberOfFromIfCached().memberRemark = channel.channelRemark;
+                        chatAdapter.getData().get(i).wkMsg.getMemberOfFromIfCached().memberName = channel.channelName;
+                        chatAdapter.getData().get(i).wkMsg.getMemberOfFromIfCached().memberAvatar = channel.avatar;
+                        chatAdapter.getData().get(i).wkMsg.getMemberOfFromIfCached().memberAvatarCacheKey = channel.avatarCacheKey;
                         isRefresh = true;
                     }
                     if (chatAdapter.getData().get(i).wkMsg.baseContentMsgModel != null && WKReader.isNotEmpty(chatAdapter.getData().get(i).wkMsg.baseContentMsgModel.entities)) {
@@ -1597,10 +1597,10 @@ public class ChatActivity extends SwipeBackActivity implements IConversationCont
                     } else {
                         //成员名字改变
                         for (int i = 0, size = chatAdapter.getData().size(); i < size; i++) {
-                            if (chatAdapter.getData().get(i).wkMsg != null && chatAdapter.getData().get(i).wkMsg.getMemberOfFrom() != null && !TextUtils.isEmpty(chatAdapter.getData().get(i).wkMsg.getMemberOfFrom().memberUID) && chatAdapter.getData().get(i).wkMsg.getMemberOfFrom().memberUID.equals(channelMember.memberUID)) {
-                                chatAdapter.getData().get(i).wkMsg.getMemberOfFrom().memberName = channelMember.memberName;
-                                chatAdapter.getData().get(i).wkMsg.getMemberOfFrom().memberRemark = channelMember.memberRemark;
-                                chatAdapter.getData().get(i).wkMsg.getMemberOfFrom().memberAvatar = channelMember.memberAvatar;
+                            if (chatAdapter.getData().get(i).wkMsg != null && chatAdapter.getData().get(i).wkMsg.getMemberOfFromIfCached() != null && !TextUtils.isEmpty(chatAdapter.getData().get(i).wkMsg.getMemberOfFromIfCached().memberUID) && chatAdapter.getData().get(i).wkMsg.getMemberOfFromIfCached().memberUID.equals(channelMember.memberUID)) {
+                                chatAdapter.getData().get(i).wkMsg.getMemberOfFromIfCached().memberName = channelMember.memberName;
+                                chatAdapter.getData().get(i).wkMsg.getMemberOfFromIfCached().memberRemark = channelMember.memberRemark;
+                                chatAdapter.getData().get(i).wkMsg.getMemberOfFromIfCached().memberAvatar = channelMember.memberAvatar;
                                 chatAdapter.getData().get(i).isRefreshAvatarAndName = true;
                                 chatAdapter.notifyItemChanged(i, chatAdapter.getData().get(i));
                             }
@@ -3313,6 +3313,9 @@ public class ChatActivity extends SwipeBackActivity implements IConversationCont
                 mInfo.uids = filteredUids;
             }
             content.mentionInfo = mInfo;
+            if (pendingCaptionMentionEntities != null && !pendingCaptionMentionEntities.isEmpty()) {
+                content.entities = pendingCaptionMentionEntities;
+            }
         } else if (chatPanelManager != null) {
             chatPanelManager.applyInputMentionsTo(content, rawText);
         }
@@ -3877,6 +3880,7 @@ public class ChatActivity extends SwipeBackActivity implements IConversationCont
     private List<String> pendingCaptionMentionUids;
     private boolean pendingCaptionMentionAll;
     private boolean pendingCaptionMentionAis;
+    private List<com.xinbida.wukongim.msgmodel.WKMsgEntity> pendingCaptionMentionEntities;
 
     ActivityResultLauncher<Intent> richTextCaptionLauncher = registerForActivityResult(new ActivityResultContracts.StartActivityForResult(), result -> {
         if (result.getResultCode() == Activity.RESULT_OK && result.getData() != null) {
@@ -3885,6 +3889,8 @@ public class ChatActivity extends SwipeBackActivity implements IConversationCont
             pendingCaptionMentionUids = result.getData().getStringArrayListExtra("mentionUids");
             pendingCaptionMentionAll = result.getData().getBooleanExtra("mentionAll", false);
             pendingCaptionMentionAis = result.getData().getBooleanExtra("mentionAis", false);
+            pendingCaptionMentionEntities = parseMentionEntitiesFromJson(
+                    result.getData().getStringExtra("mentionEntities"));
             if (paths != null && !paths.isEmpty()) {
                 String text = caption != null ? caption : "";
                 if (!TextUtils.isEmpty(text) && chatPanelManager != null && chatPanelManager.isTextOverByteLimit(text)) {
@@ -3897,6 +3903,7 @@ public class ChatActivity extends SwipeBackActivity implements IConversationCont
             pendingCaptionMentionUids = null;
             pendingCaptionMentionAll = false;
             pendingCaptionMentionAis = false;
+            pendingCaptionMentionEntities = null;
         } else if (result.getResultCode() == Activity.RESULT_CANCELED && result.getData() != null) {
             String restored = result.getData().getStringExtra("caption");
             if (restored != null && !restored.isEmpty() && chatPanelManager != null) {
@@ -3940,6 +3947,28 @@ public class ChatActivity extends SwipeBackActivity implements IConversationCont
             }
         }
     });
+
+    private static List<com.xinbida.wukongim.msgmodel.WKMsgEntity> parseMentionEntitiesFromJson(String json) {
+        if (TextUtils.isEmpty(json)) return null;
+        try {
+            org.json.JSONArray arr = new org.json.JSONArray(json);
+            List<com.xinbida.wukongim.msgmodel.WKMsgEntity> result = new ArrayList<>();
+            for (int i = 0; i < arr.length(); i++) {
+                org.json.JSONObject obj = arr.getJSONObject(i);
+                com.xinbida.wukongim.msgmodel.WKMsgEntity entity = new com.xinbida.wukongim.msgmodel.WKMsgEntity();
+                entity.type = com.chat.base.msg.ChatContentSpanType.getMention();
+                entity.value = obj.optString("uid", "");
+                entity.offset = obj.optInt("offset", -1);
+                entity.length = obj.optInt("length", 0);
+                if (!entity.value.isEmpty() && entity.offset >= 0 && entity.length > 0) {
+                    result.add(entity);
+                }
+            }
+            return result.isEmpty() ? null : result;
+        } catch (org.json.JSONException e) {
+            return null;
+        }
+    }
 
     private void handleFileResult(android.net.Uri uri) {
         try {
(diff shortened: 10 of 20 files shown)


Jerry-Xin (changes requested):
The PR is relevant to this repository, but the new rich-text mention entity encoding has a blocking offset-contract bug.

**🔴 Blocking**
- 🔴 Critical: `mention.entities` offsets are written from the composer’s raw text, but rich-text messages are serialized with image blocks before the text block. `ChatPanelManager.applyInputMentionsTo()` copies `editText.allEntity` directly into `content.entities` at `wkuikit/src/main/java/com/chat/uikit/chat/ChatPanelManager.kt:463`, `:497`, while `WKRichTextSender.finish()` builds `blocks` as images first, then text, and builds `plain` from those blocks at `wkuikit/src/main/java/com/chat/uikit/chat/manager/WKRichTextSender.java:278`. Example: one image plus caption `@Bob hi` sends `plain` like `[图片]@Bob hi`, but the mention entity still has `offset=0`, so standards-compliant receivers will apply it to the image placeholder or drop it. The new Android renderer then partially masks this by counting only text block lengths and ignoring image placeholders at `wkuikit/src/main/java/com/chat/uikit/chat/provider/WKRichTextProvider.kt:126`, `:143`, which also means it will fail for valid incoming payloads whose offsets are relative to rich-text `plain`. Please align the contract before merging: either make block order match the composer text-first contract, adjust outgoing offsets by preceding block placeholder lengths, or explicitly make `mention.entities` block-local and update all send/render paths accordingly.

**💬 Non-blocking**
- 🟡 Warning: The new batch extra query can exceed SQLite bind limits for large conversation lists because it builds one unchunked `IN (...)` query from every channel id at `wkim/src/main/java/com/xinbida/wukongim/db/ConversationDbManager.java:467`. Since exceptions are swallowed at `:477`, drafts/extras can silently disappear for large accounts. Consider chunking IDs and logging failures.
- 🟡 Warning: Several new DB catch blocks silently ignore all exceptions, for example `ConversationDbManager.java:457` and `:477`. This avoids crashes, but it also hides schema/query/regression issues and makes production diagnosis difficult. Prefer logging at least a lightweight warning with method context.

**✅ Highlights**
- The PR is scoped to existing chat, message, DB, push, and UI crash-fix functionality.
- The delayed revoke-neighbor refresh now rechecks current adapter bounds before notifying, which addresses a real RecyclerView race in `ChatActivity.java:4539`.
- The clipboard and OPPO push guards are pragmatic crash hardening.

lml2468 (changes requested):
🔴 REQUEST_CHANGES

**octo-android PR #42** — feat(richtext): mention highlighting + crash fixes (v1.3.2)
by @mzanalyticst · SHA `da11f2be36c4`

**Blocking (P0/P1):**
- 🔴 RichText mention rendering can hit DB on the main bind path — `wkuikit/src/main/java/com/chat/uikit/chat/provider/WKRichTextProvider.kt:272` — each non-sentinel mention calls `channelManager.getChannel(uid, PERSONAL)`; cache miss falls through to `ChannelDBManager.getInstance().query(...)` at `wkim/src/main/java/com/xinbida/wukongim/manager/ChannelManager.java:99`, partially undoing this PR's own main-thread DB-query avoidance (the very reason `getMemberOfFromIfCached` was introduced in `WKMsg.java:240`). For long messages with many @mentions on a cold cache, this can produce the same ANR pattern. Fix: use an `IfCached` accessor (channelInfoCache.get only, no slow path), or pre-warm the cache for all mention uids off-thread before bind.

**Non-blocking (P2):**
- 🟡 Silent DB error swallowing — `wkim/src/main/java/com/xinbida/wukongim/db/ConversationDbManager.java:457`, `:478` — `queryMsgExtraWithChannel` and `queryWithExtraChannelIds` now `catch (Exception ignored)`. Acceptable for "DB pool closed during shutdown", but a malformed-SQL or schema-mismatch regression would also be swallowed silently. Suggest narrowing the catch (e.g. `IllegalStateException`/`SQLiteException` only) or at minimum `Log.w` so future bugs are not invisible.
- 🟡 Cross-block mentions silently dropped — `wkuikit/src/main/java/com/chat/uikit/chat/provider/WKRichTextProvider.kt:147` — partition predicate `offset >= startOff && offset + length <= endOff` skips any entity straddling a text→image→text seam. Today the server doesn't emit those, but please add a code comment marking this as out-of-spec, or a unit test pinning the contract.

**Verified:**
- ✅ `WKRichTextContent.encodeMsg` writes mention `{all/humans/ais/uids/entities}`; SDK `getSendPayload` won't double-write (`json.has("mention")` short-circuit).
- ✅ UI decode path merges `mention.entities` via `MentionEntityHelper.mergeMentionEntities` at `WKUIChatMsgItemEntity.java:163`.
- ✅ `buildMentionSpans` sorts entities descending and uses `lastProcessedStart` to skip overlaps; trailing display-name space does not corrupt earlier offsets because replacements run from tail to head.
- ✅ Caption mention flow: `WKRichTextCaptionActivity.kt:139` → `ChatActivity.parseMentionEntitiesFromJson` → `pendingCaptionMentionEntities` → `sendRichTextTray` → `content.entities`. `pendingCaptionMentionEntities` is cleared after `RESULT_OK`; `RESULT_CANCELED` path matches existing `pendingCaptionMentionUids` behavior (not a regression).
- ✅ Crash fixes: ContactEditText paste null-guard, ChatActivity revoke `postDelayed` bounds check (uses live `chatAdapter.getData()` snapshot), OPPO init `Throwable` catch with early-return before `register`, RemindMemberAdapter `instanceof NoEventRecycleView`, ProhibitWordDB `helper.isClosed` + `cursor.use`, UserDetail/MyHeadPortrait `isFinishing()/isDestroyed()` guards.
- ✅ Batch extras path: `getMsgExtrasForChannels` map keyed as `channelID + ":" + channelType` at both writer (`ConversationManager.java:359`) and all three reader call sites in `ChatFragment.java`.
- ✅ All 5 `getMemberOfFromIfCached()` swaps in `ChatActivity.attachChannelListeners` are in refresh callbacks on main thread — correct surface for the cache-only accessor.
- ✅ ProGuard `-keep class com.chat.uikit.thread.msgmodel.**` justified — directory contains reflectively-deserialized `WKThreadCreatedContent` registered via SDK content-type registry.
- ✅ Version: `1.3.1 (32)` → `1.3.2 (33)`.

齐静春 📖 (via Codex)

yujiawei (changes requested):
# Code Review — PR #42 (octo-android)

## Summary

This PR delivers two things: (1) @mention highlighting + click-to-profile for RichText (type=14) mixed image/text messages, and (2) seven crash hardenings plus a conversation-extras batch-query optimization. The crash fixes are well-targeted and the mention pipeline is correct on the Android→Android happy path (the descending-offset `ssb.replace` index math is sound, and send/receive offset bases are self-consistent). However, two issues should be resolved before merge.

**Verdict: Changes requested.** Two functional blockers (P1): a draft-loss regression in the batch query, and an unverified cross-platform offset contract behind a clickable-navigation feature. Everything else is a non-blocking suggestion.

---

## P1 — Blocking

### P1-1. Batch draft query has no chunking; a new silent catch turns the failure into total draft loss

- `wkim/.../db/ConversationDbManager.java` — `queryWithExtraChannelIds` (~L467-480) builds `"channel_id in (" + getPlaceholders(channelIds.size()) + ")"`, one bind variable per id, **no chunking**.
- Callers in `wkuikit/.../fragment/ChatFragment.java` (the conversation-list rebuild ~L953, the `refresh_conversation_extras` endpoint ~L1252, and `refreshExtrasIfNeeded` ~L2007) now pass the **full, unbounded** conversation list.
- On Android with SQLite < 3.32 (i.e. Android 11 / API ≤ 30, still a large install base), `SQLITE_MAX_VARIABLE_NUMBER` is **999**. A user with > 999 conversations triggers `SQLiteException: too many SQL variables`.
- This PR newly added `catch (Exception ignored)` (~L477) around that query, so the exception is swallowed and an **empty list** is returned → `getMsgExtrasForChannels` returns an empty map → **every conversation's draft/extra silently vanishes from the list**.

This is strictly worse than the pre-PR per-item path (`getMsgExtraWithChannel`, 2 bind vars each), which could never hit the limit. Drafts remain in the DB (no data corruption), but they disappear from the UI with zero diagnosability.

**Fix:** chunk `channelIds` into batches (≤ 900) inside `queryMsgExtrasForChannelIds`/`getMsgExtrasForChannels`, run one IN-query per chunk, and merge. Do not rely on swallow-and-return-empty for correctness. Add a test with > 999 ids.

### P1-2. Cross-platform mention offset base is unverified; a mismatch silently highlights/links the wrong user

- Receive: `wkuikit/.../chat/provider/WKRichTextProvider.kt` (~L126-159, L243-264) accumulates `textOffset` **only over text-block lengths** (image blocks do not advance it), and computes `localStart = entity.offset - blockTextOffset`. The only guard is `localEnd <= ssb.length`; there is no validation of the offset *base* and no fallback.
- Android→Android is self-consistent: `WKRichTextSender.finish` emits image blocks first then one text block (= caption), and caption entity offsets (`ContactEditText.getAllEntity`) are 0-based over caption text only. So offsets line up. ✅
- But the PR explicitly targets "对齐 iOS 端行为" (align with iOS). `plain` for an images-first message is `[图片][图片]caption` — placeholders included. **If iOS/server bases `entity.offset` on `plain` rather than on the images-excluded text-block concatenation, every offset shifts by `len("[图片]") * imageCount`** and the highlight (and the clickable profile link) lands on the wrong characters / wrong user — silently.

Because the @mention is clickable and navigates to a profile, a shifted-but-in-range offset is more than cosmetic. This contract cannot be verified from octo-android alone.

**Fix:** pin the wire contract with iOS/server: `entity.offset` MUST be the character offset into the text-block concatenation (images excluded). Add a cross-platform fixture test for an images-first (and ideally multi-text-block) message originating from a non-Android sender. If the agreed base is `plain`-relative, advance `textOffset` by the placeholder length on image blocks too.

---

## P2 — Non-blocking (suggestions / nits)

**Mention rendering**

- **Double-spanning of `@所有人`/`@所有AI`** (`WKRichTextProvider.kt` ~L258-317): sentinel entities (uid `-1`/`-2`) are stripped from `mInfo.uids` but still travel in `mention.entities`, so on receive they get an entity span *and* a broadcast-highlight span over the same range. Same color/weight, so visually harmless, but redundant. Consider skipping sentinel entities in `buildMentionSpans` and letting `applyBroadcastHighlight` be the sole owner of broadcast styling.
- **Asymmetric word-boundary check** (`WKRichTextProvider.kt` ~L350-356): the broadcast guard only inspects the char *after* the token, not before, so `abc@所有人` would still highlight mid-word. Low impact (the picker inserts the token standalone). Mirror the trailing check on `currentText[idx-1]`.
- **Locale-fragile broadcast tokens** (`applyBroadcastHighlight`): highlighting relies on a hardcoded literal token list plus a localized resource. It happens to cover the two shipping locales (zh default + en) today, but a future third `strings.xml` without updating the list would silently stop highlighting. Driving the token set from resources for all locales would be more robust. (Note: carrying explicit offsets for broadcast tags is *not* a viable fix here — sentinels are deliberately kept off the wire to avoid the iOS PR#128 issue.)
- **`LinkMovementMethod` vs. long-press menu** (`WKRichTextProvider.kt` ~L229): every text block sets `movementMethod = LinkMovementMethod` (pre-existing, but now carries real `ClickableSpan`s). `LinkMovementMethod` can suppress the parent `BubbleLayout`'s `OnLongClickListener`, risking loss of the standard copy/reply/forward/delete/revoke/reaction menu when long-pressing on the text region. Please verify long-press still works on a RichText message that contains a mention; if it regresses, use a movement method that defers non-link long-presses to the parent.
- **Redundant wire bytes** (`WKRichTextContent.java` ~L219-223): mention entities are serialized both under `mention.entities` (this PR) and in the SDK's top-level `entities` (`WKProto.getSendPayload`). Receive dedups by offset+length, so rendering is fine — but the same data is sent twice. Acceptable (the top-level copy is a highlight safety net), just worth noting.

**DB hardening / batch query**

- **Cursor leak on the single-query path** (`ConversationDbManager.java` ~L453-455): `cursor.close()` is inside the `try`, after `serializeMsgExtra(cursor)`. If serialize throws (e.g. cursor index OOB / pool closed mid-read), the new `catch (Exception ignored)` swallows it and the cursor leaks. The batch path correctly uses try-with-resources — make the single path consistent (try-with-resources or `finally`).
- **Silent catches with no logging** (`ConversationDbManager.java` ~L457/L477, `ProhibitWordDB.kt`, `ReminderDBManager.java`): the crash-hardening catches swallow *all* exceptions with no log, unlike the OPPO fix in this same PR which logs. Real schema/SQL regressions (including P1-1) become invisible. At minimum log before swallowing; ideally narrow to the expected pool-closed/`IllegalStateException`. The two new `ConversationDbManager` read methods also skip the `helper == null || isClosed()` pre-guard used elsewhere in the file.
- **`getMemberOfFromIfCached` skips in-place refresh for not-yet-cached members** (`ChatActivity.java` ~L1565-1571, L1600-1607): the old `getMemberOfFrom()` hit the DB on a cache miss (the ANR source being fixed). The new cached-only accessor returns null when `memberOfFrom` isn't loaded, so the whole rename/avatar in-place update block is skipped for those messages. This is an acceptable ANR tradeoff — bind-time lazy-load (`ChatConversationAdapter.getFromName`) still resolves the name on next rebind — but it's a behavior change worth calling out in the PR description. (Note the bind path still does a main-thread DB read, so the ANR root cause is only partially addressed.)
- **Undeduplicated / inconsistent caller lists** (`ChatFragment.java` ~L953/L1252/L2007): the three batch-build sites don't dedup ids, and the list-rebuild site omits the draft-needed filter the other sites use, so they over-fetch and push the IN-clause closer to the variable cliff (P1-1). Functionally harmless (the `id:type` map key disambiguates correctly), but extracting one shared helper that dedups (and pre-filters) would reduce both waste and the variable count.

**Tests**

- **Zero test changes in this PR** (`git show --stat` of the head commit = 18 files, none under `src/test`). The highest-risk new logic — `encodeMsg` mention assembly, `parseMentionEntitiesFromJson` round-trip, and the cross-block `textOffset` accumulation/filter — has no automated coverage. The pre-existing RichText tests are from prior work and don't touch the entities path. Given P1-2, a cross-platform offset fixture test is the most valuable addition.

---

## Verified-correct (no action)

- `refreshMsg` IndexOutOfBounds fix: caching `entity = list.get(i)` and re-reading `chatAdapter.getData()` with bounds checks in the revoke neighbor-refresh `postDelayed` is correct. ✅
- ContactEditText paste NPE, OPPO `initOPPO` Throwable guard, RemindMemberAdapter `instanceof` guard, `MyHeadPortraitActivity`/`UserDetailActivity` `isFinishing()||isDestroyed()` guards, `ReminderDBManager` catch → false: all correct. ✅
- ProGuard `com.chat.uikit.thread.msgmodel.**` keep rule: package exists (`WKThreadCreatedContent`), rule is valid. ✅
- Batch-query map key `channelID:channelType`: `serializeMsgExtra` populates `channelType` per row, so keys match caller lookups exactly — the `IN`-by-`channel_id`-only over-fetch is harmless (no extras missed). ✅
- Version bump 1.3.1(32) → 1.3.2(33). ✅

Jerry-Xin (changes requested):
Summary: The PR is in scope for this Android chat project, but the new rich-text mention payload path can emit invalid mention entities for broadcast mentions and inconsistent offsets.

**🔴 Blocking**
- 🔴 Critical: Rich-text sending writes broadcast sentinel IDs into `mention.entities`. `scanPlainTextMentions()` adds broadcast entries with values `-1` / `-2`, and `applyInputMentionsTo()` now assigns all mention entities directly to `content.entities` without the sentinel filtering already used by the normal text-send path. The caption modal has the same issue: it creates `-1` / `-2` entries and serializes them into `mentionEntities`, then `ChatActivity` copies them to `content.entities`. `WKRichTextContent.encodeMsg()` then writes those values into `mention.entities`. This violates the existing send-path contract that sentinel IDs must not be emitted as mention entities, and can make servers or older clients treat `-1` / `-2` as real user IDs. See `wkuikit/src/main/java/com/chat/uikit/chat/ChatPanelManager.kt:497`, `wkuikit/src/main/java/com/chat/uikit/chat/ChatPanelManager.kt:1989`, `wkuikit/src/main/java/com/chat/uikit/chat/WKRichTextCaptionActivity.kt:140`, `wkuikit/src/main/java/com/chat/uikit/chat/ChatActivity.java:3316`, and the existing text-path guard at `wkuikit/src/main/java/com/chat/uikit/chat/ChatPanelManager.kt:1635`.

- 🔴 Critical: Rich-text mention offsets are serialized in two incompatible coordinate systems. `WKRichTextContent.encodeMsg()` converts `content.entities` into plain-relative `mention.entities`, but `WKProto.getSendPayload()` later also writes the original `baseContentMsgModel.entities` as top-level `entities` without the rich-text image-placeholder offset adjustment. Since `WKRichTextSender` currently builds blocks as images first, then text, a caption mention at offset `0` is correctly shifted inside `mention.entities` but remains `0` in top-level `entities`, pointing into the image placeholder area instead of the caption text. That creates a malformed payload for any consumer that reads top-level entities. See `wkuikit/src/main/java/com/chat/uikit/chat/msgmodel/WKRichTextContent.java:221`, `wkim/src/main/java/com/xinbida/wukongim/message/WKProto.java:361`, and `wkuikit/src/main/java/com/chat/uikit/chat/manager/WKRichTextSender.java:278`.

**💬 Non-blocking**
- 🟡 Warning: `buildPlainRelativeEntitiesJson()` applies the first text block’s plain offset to every mention entity. That works only while rich text has a single text block; it will produce wrong offsets once true interleaved text/image blocks are supported. See `wkuikit/src/main/java/com/chat/uikit/chat/msgmodel/WKRichTextContent.java:274`.

- 🟡 Warning: Several new DB error handlers silently return empty/default data. That avoids crashes, but it also hides corruption or query-shape bugs during development. Consider logging the exception object, not only a static message. See `wkim/src/main/java/com/xinbida/wukongim/db/ConversationDbManager.java:457` and `wkuikit/src/main/java/com/chat/uikit/db/ProhibitWordDB.kt:92`.

**✅ Highlights**
- The clipboard null checks, lifecycle guards around picker callbacks, and delayed adapter bounds checks are good crash-hardening changes.
- Bulk conversation-extra loading is a useful direction for reducing main-thread N+1 DB work.

Verification note: I attempted to run `./gradlew :wkuikit:compileDebugKotlin --no-daemon`, but this environment has no Java runtime installed, so build verification could not be completed.

lml2468 (approved):
✅ APPROVED

**octo-android PR #42** R2 — feat(richtext): mention highlighting + crash fixes (v1.3.2)
SHA `b6fa9d337f10` (prior `da11f2be36c4`)

**R1 blockers — fixed:**
- ✅ RichText mention offset 与 plain 不对齐 — sender 始终 image blocks 在前 + 单一 text block，`buildPlainRelativeEntitiesJson` 用 `textBlockPlainStart` 把 entity offset 提升为 plain-relative；接收侧 `plainOffset` 累加同一个 `[图片]` 占位长度后再 partition 文本块 — `wkuikit/.../WKRichTextSender.java:278`、`WKRichTextContent.java:271`、`WKRichTextProvider.kt:126`
- ✅ 跨端 @mention offset 基准 — wire 上的 `mention.entities` 现在统一 plain-relative，与 server/iOS 契约一致；接收侧用 plain-relative 区间过滤 entity — `WKRichTextContent.java:295`、`WKRichTextProvider.kt:146`
- ✅ SQLite 999 参数限制 + 静默吞错 — `queryWithExtraChannelIds` 按 500 分片，分片失败逐片记录；`queryMsgExtraWithChannel` 也改为 `WKLoggerUtils.e()` — `ConversationDbManager.java:444`、`:468`
- ✅ RichText bind 主线程 DB 慢路径 — 新增 `ChannelManager.getChannelIfCached`（只查 `channelInfoCache`，无 DB），`WKRichTextProvider` 改用之，ANR 风险消除 — `ChannelManager.java:84`、`WKRichTextProvider.kt:273`
- ✅ 跨 block mention 处理 — sender 当前只产出单一 text block + 前置 images，provider 用 `plainOffset` 跨过 image 占位后再定位 text block 边界，entity 正确落在目标文本块

**Follow-up (non-blocking):**
- 🟡 `buildPlainRelativeEntitiesJson` 默认 sender 仅出单一 text block；若将来 sender 演化为 text/image/text 穿插，应改成逐 block 维护当前 plainOffset 而非一次性的 `textBlockPlainStart`
- 🟡 加 `>1000` channel IDs 的 chunking 回归测试，锁住该 DB 行为
- 🟡 `buildPlainRelativeEntitiesJson` 里的 `catch (JSONException ignored)` 是新增的唯一一处静默吞错，建议同样加一行 log 保持一致

齐静春 📖 (via Codex)

lml2468 (changes requested):
🔴 REQUEST_CHANGES (撤回我的 R2 APPROVE)

**octo-android PR #42** R2 — feat(richtext): mention highlighting + crash fixes (v1.3.2)
SHA `b6fa9d337f10`

Jerry-Xin 这一轮挖出我漏掉的契约问题，复核后属实，撤回我前一条 APPROVE。

**Blocking (P0):**
- 🔴 `mention.entities` 写入了广播哨兵 uid `-1`/`-2`，违反 wire 契约 — `ChatPanelManager.kt:497` 在文本-only 路径用 `entities.filter { it.type == mention }` 把全部 mention entity 灌进 `content.entities`，但 `scanBroadcastMentions` 在 `ChatPanelManager.kt:1991` 会给 `@所有人 / @所有AI` 造 entity 时直接把 `value = "-1"/"-2"` 哨兵写进去。结果 `buildPlainRelativeEntitiesJson` 把这两条也序列化到 wire，iOS/web/server 收到后会按假 uid 解析或丢弃，跨端高亮/通知错乱。修法：发送侧序列化前过滤掉 `value == "-1" || value == "-2"` 的 entity，让 `mention.entities` 只承载真实用户 uid；广播仍由 `mention.all/humans/ais` 标志位承担。

**Blocking (P0，沿用 Jerry-Xin 判断):**
- 🔴 富文本 mention 偏移坐标系仍有不兼容路径 — Jerry-Xin 指出本次序列化仍可能生成错误 payload。请作者按 Jerry-Xin 复核意见对齐：明确 sender 单/多 text block 演化时 `buildPlainRelativeEntitiesJson` 的契约边界，并补对端解码用例验证。本人 R2 误判了「sender 单 text block + 前置 images」是 settled contract，事实是 sender 行为尚未约束在 schema 层，应至少加 unit test/断言锁死契约后再合入。

**Non-blocking (P2):**
- 🟡 `buildPlainRelativeEntitiesJson` 假设只有单一 text block 的契约要么写成 sender 端的 assertion，要么改成逐 block 维护当前 plainOffset
- 🟡 `>1000 channel IDs` 分片回归测试
- 🟡 `buildPlainRelativeEntitiesJson` 内 `catch (JSONException ignored)` 加 log，保持与本 PR 其它 catch 一致

**自检 note:** R2 APPROVE 漏掉了「发送侧 entity 序列化前未过滤哨兵」这条，根因是只看了 receiver 侧 `isSentinel` 的兜底渲染，没回查 sender 侧 entity 入口；下一轮 review 会先列「entity 从 composer 到 wire 的全路径过滤点」清单再下结论。

齐静春 📖 (via Codex)

lml2468 (changes requested):
🔴 REQUEST_CHANGES (撤回我的 R2 APPROVE)

**octo-android PR #42** R2 — feat(richtext): mention highlighting + crash fixes (v1.3.2)
SHA `b6fa9d337f10`

Jerry-Xin 这一轮挖出我漏掉的契约问题，复核后属实，撤回我前一条 APPROVE。

**Blocking (P0):**
- 🔴 `mention.entities` 写入了广播哨兵 uid `-1`/`-2`，违反 wire 契约 — `ChatPanelManager.kt:497` 在文本-only 路径用 `entities.filter { it.type == mention }` 把全部 mention entity 灌进 `content.entities`，但 `scanBroadcastMentions` 在 `ChatPanelManager.kt:1991` 会给 `@所有人 / @所有AI` 造 entity 时直接把 `value = "-1"/"-2"` 哨兵写进去。结果 `buildPlainRelativeEntitiesJson` 把这两条也序列化到 wire，iOS/web/server 收到后会按假 uid 解析或丢弃，跨端高亮/通知错乱。修法：发送侧序列化前过滤掉 `value == "-1" || value == "-2"` 的 entity，让 `mention.entities` 只承载真实用户 uid；广播仍由 `mention.all/humans/ais` 标志位承担。

**Blocking (P0，沿用 Jerry-Xin 判断):**
- 🔴 富文本 mention 偏移坐标系仍有不兼容路径 — Jerry-Xin 指出本次序列化仍可能生成错误 payload。请作者按 Jerry-Xin 复核意见对齐：明确 sender 单/多 text block 演化时 `buildPlainRelativeEntitiesJson` 的契约边界，并补对端解码用例验证。本人 R2 误判了「sender 单 text block + 前置 images」是 settled contract，事实是 sender 行为尚未约束在 schema 层，应至少加 unit test/断言锁死契约后再合入。

**Non-blocking (P2):**
- 🟡 `buildPlainRelativeEntitiesJson` 假设只有单一 text block 的契约要么写成 sender 端的 assertion，要么改成逐 block 维护当前 plainOffset
- 🟡 `>1000 channel IDs` 分片回归测试
- 🟡 `buildPlainRelativeEntitiesJson` 内 `catch (JSONException ignored)` 加 log，保持与本 PR 其它 catch 一致

**自检 note:** R2 APPROVE 漏掉了「发送侧 entity 序列化前未过滤哨兵」这条，根因是只看了 receiver 侧 `isSentinel` 的兜底渲染，没回查 sender 侧 entity 入口；下一轮 review 会先列「entity 从 composer 到 wire 的全路径过滤点」清单再下结论。

齐静春 📖 (via Codex)

Jerry-Xin (changes requested):
Scope check passed: this PR modifies in-scope Android chat, IM, push, and UI behavior for `octo-android`, but one blocking notification/mention bug needs fixing before merge.

**🔴 Blocking**
- 🔴 Critical: Oversized rich-text captions can send invisible mention notifications with an image-only message. In `ChatActivity.richTextCaptionLauncher`, when the caption exceeds the byte limit, the code calls `sendRichTextTray("", paths, null, null)` while `pendingCaptionMentionUids`, `pendingCaptionMentionAll`, `pendingCaptionMentionAis`, and `pendingCaptionMentionEntities` are still populated from the caption result (`wkuikit/src/main/java/com/chat/uikit/chat/ChatActivity.java:3889`, `wkuikit/src/main/java/com/chat/uikit/chat/ChatActivity.java:3896`). `sendRichTextTray` then applies those pending mentions to a `WKRichTextContent` even though `rawText` is empty (`wkuikit/src/main/java/com/chat/uikit/chat/ChatActivity.java:3295`, `wkuikit/src/main/java/com/chat/uikit/chat/ChatActivity.java:3316`). Result: recipients can receive a mention alert from an image-only message that contains no visible mention text, and the serialized mention entity offsets no longer correspond to any rendered text block. Clear or ignore pending caption mentions when the caption is not included in the rich-text message, or only apply pending mention metadata when `rawText` is non-empty and actually contains the mention spans.

**💬 Non-blocking**
- 🟡 Warning: `ConversationDbManager.queryWithExtraChannelIds` queries `conversationExtra` by `channel_id in (...)` only, while callers consume results by `channelID:channelType` (`wkim/src/main/java/com/xinbida/wukongim/db/ConversationDbManager.java:475`, `wkim/src/main/java/com/xinbida/wukongim/manager/ConversationManager.java:356`). Current lookups avoid applying the wrong type, but the API shape is easy to misuse and may fetch unnecessary rows when IDs overlap across channel types. Prefer passing channel ID/type pairs and constraining both columns.
- 🔵 Suggestion: Several new DB crash guards silently swallow exceptions (`wkuikit/src/main/java/com/chat/uikit/db/ProhibitWordDB.kt:94`, `wkim/src/main/java/com/xinbida/wukongim/db/ReminderDBManager.java:77`). Returning safe defaults is reasonable for closed-pool crashes, but consider lightweight logging so future data corruption or lifecycle issues are diagnosable.
- 🔵 Suggestion: Add focused tests for mention serialization/rendering around image-before-text offsets, empty caption with images, and oversized caption fallback. This PR changes both notification payloads and UI click behavior, and the edge cases are easy to regress.

**✅ Highlights**
- The delayed revoke-neighbor refresh now re-reads the adapter data and bounds-checks both indexes before notifying (`wkuikit/src/main/java/com/chat/uikit/chat/ChatActivity.java:4538`), which addresses the reported crash pattern.
- The OPPO push initialization guard catches `Throwable`, which is appropriate for optional vendor SDK class-loading failures (`wkpush/src/main/java/com/chat/push/WKPushApplication.java:143`).
- The rich-text provider validates entity bounds before applying spans (`wkuikit/src/main/java/com/chat/uikit/chat/provider/WKRichTextProvider.kt:263`), reducing malformed-payload crash risk.

I could not run the Gradle compile because this environment has no Java runtime installed.

lml2468 (changes requested):
🔴 REQUEST_CHANGES

**octo-android PR #42** R3 — feat(richtext): mention highlighting + crash fixes (v1.3.2)
SHA `2d457fc974dc` (prior `b6fa9d337f10`)

R2 三个 🔴 blocker（哨兵 / 坐标系 / 单 text block 契约）本次都修对了。但 Jerry-Xin 这一轮挖出一个我也漏了的 P0，复核属实，沿用他的判断：

**Blocking (P0):**
- 🔴 超长 caption 降级路径发送幻影 @mention 通知 — `ChatActivity.java:3896-3898` — 当 caption 超字节限制时执行 `sendRichTextTray("", paths, ...)`（文本被剥离改走文件），但 `pendingCaptionMentionUids / pendingCaptionMentionAll / pendingCaptionMentionAis / pendingCaptionMentionEntities` 仍然带着旧 caption 的 mention 元数据流到 `sendRichTextTray:3295` 并被灌进 `WKRichTextContent`。结果：图片-only 消息携带 mention 通知 + entity offset，但消息没有任何可见文本，接收侧收到看不出 mention 的通知，offset 也指不到任何 rendered 文本块。修法二选一：a) 超长降级分支里把四个 `pendingCaptionMention*` 主动清空再调 `sendRichTextTray("")`；b) `sendRichTextTray` 在 `rawText` 为空时直接忽略 pending mentions。

**R2 blockers — fixed:**
- ✅ 哨兵 UID 不再上 wire — `WKRichTextSender.adjustEntitiesForPlain` 入队前过滤 `value == "-1"/"-2"`
- ✅ Offset 坐标系统一 — 偏移调整搬到 sender 端一次性 mutate `content.entities.offset`，`encodeMsg()` 与 SDK `getSendPayload()` 单源一致
- ✅ 单 text block 假设当前 sender 满足 — `finish()` 始终 images-first + 单一可选 text block

**Non-blocking (沿用 Jerry-Xin):**
- 🟡 `queryWithExtraChannelIds` 只按 `channel_id IN (...)` 查，callers 用 `channelID:channelType` 复合键消费；当前不会错配但 API 形状容易误用，建议传 (id, type) pair 双列约束
- 🟡 sender 若演化为多 text block / image-text 穿插，`adjustEntitiesForPlain(imageCount)` 需改成逐 block 推进 plainOffset
- 🟡 mention 序列化 / image-before-text offset / 空 caption + 图片 / 超长 caption 降级 这几个边界加 unit test

**自检 note:** 我前一次 R2 漏了哨兵入口检查，这次 R3 又漏了 caption 超长降级路径的 pending state 清理，根因都是同类——只验主路径不扫边界 / 失败分支。下一轮 review 会先扫 sender entity 全部入口 + 所有早退/降级分支再下结论。

齐静春 📖 (via Codex)

lml2468 (approved):
✅ APPROVED

**octo-android PR #42** R4 — feat(richtext): mention highlighting + crash fixes (v1.3.2)
SHA `d3805420286b` (prior `2d457fc974dc`)

**R3 blocker — fixed:**
- ✅ 超长 caption 降级路径不再发送幻影 @mention — `ChatActivity.java:3897-3900` — 在 `sendRichTextTray("", paths, ...)` 之前主动清空 4 个 `pendingCaptionMention*`，确保图片-only 消息不携带 caption 的 mention 元数据，offset 也不再悬空指向被剥离的 caption。修法干净，只动了 caption 超长这一条分支，未影响主路径。

**Bonus 修复（我 R1 提的 🟡）：**
- ✅ `ProhibitWordDB.getMaxVersion/getAll/queryWithsIds` 静默吞错改为 `android.util.Log.w(..., e)` — `ProhibitWordDB.kt:97/117/147`
- ✅ `ReminderDBManager.hasUndoneReminderWithChannelPrefix` 静默吞错改为 `WKLoggerUtils.e()` — `ReminderDBManager.java:78`

**Verified (re-check 关键路径):**
- 普通 caption 路径（≤ 字节限制）走 `sendRichTextTray(text, paths, ...)`，pending mentions 在 `sendRichTextTray` 内被消费，外层最后 `pendingCaptionMention* = null` 清理 — 未回归
- `RESULT_CANCELED` 路径不动 pending，但 launcher 下次 `RESULT_OK` 时会覆盖整组字段，无残留隐患
- R1/R2/R3 累计修过的 5 个 🔴（哨兵 / 坐标系 / 单 block 契约 / DB 主线程 / 幻影 mention）以及 4 个 🟡（DB 静默吞错 ×4）现在全部闭环
- Sender / receiver / encodeMsg / SDK 序列化各路径未变更，回归面收窄到 caption 降级 + 三处 DB log，diff `+12 -4` 干净

**遗留 follow-up（不阻塞合入）：**
- 🟡 `queryWithExtraChannelIds` 仅按 `channel_id` 查（不约束 channel_type），Jerry-Xin 提议改为 (id, type) 双列；现实场景没人复用 channel_id 跨 type，OK 作为后续 hardening
- 🟡 sender 演化到多 text block 时 `adjustEntitiesForPlain(imageCount)` 需改逐 block 推进 plainOffset
- 🟡 `>1000 channel IDs` 分片 + caption 边界 unit test

齐静春 📖 (via Codex)

Jerry-Xin (commented):
The PR is in scope for this repository and addresses relevant Android chat, rich text, DB, push, and crash-fix code paths; I found no blocking issues.

💬 Non-blocking

🟡 Warning: [ConversationDbManager.java](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-z6wro57t/repo/wkim/src/main/java/com/xinbida/wukongim/db/ConversationDbManager.java:475) batches extras by `channel_id` only, while callers later key by `channelID:channelType`. This should not assign the wrong extra because the map includes `channelType`, but it can fetch unrelated rows when the same ID exists under multiple channel types. Passing `(channelID, channelType)` pairs would keep the query precise.

🟡 Warning: [WKRichTextContent.java](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-z6wro57t/repo/wkuikit/src/main/java/com/chat/uikit/chat/msgmodel/WKRichTextContent.java:201) only writes nested `mention.entities` when mention uids or broadcast flags are present. Current send paths set those fields, but a future/bot path that sets only `entities` would lose the nested mention payload on encode. Consider including a `hasMentionEntities` condition.

🔵 Suggestion: [WKRichTextSender.java](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-z6wro57t/repo/wkuikit/src/main/java/com/chat/uikit/chat/manager/WKRichTextSender.java:46) and [WKRichTextComposeModel.java](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-z6wro57t/repo/wkuikit/src/main/java/com/chat/uikit/chat/manager/WKRichTextComposeModel.java:157) have comments saying text comes before images, while the implementation and tests use images first, then text. Cleaning this up would avoid future offset mistakes.

✅ Highlights

The bounds checks added around delayed neighbor refresh in [ChatActivity.java](/private/var/folders/fj/dr1p6h5n55l0ny89qk2n26tw0000gn/T/pr-review-z6wro57t/repo/wkuikit/src/main/java/com/chat/uikit/chat/ChatActivity.java:4538) address the reported out-of-range crash cleanly.

The rich text mention implementation keeps offsets plain-relative after image placeholders and filters sentinel broadcast IDs before sending, which matches the receiver’s block-offset rendering model.

I attempted to run the focused rich text unit tests, but this environment has no Java runtime installed, so test execution could not start.

yujiawei (approved):
# Code Review — PR #42 (octo-android)

**Verdict: APPROVED.** This is a well-structured release (v1.3.2) combining a new feature (mention highlighting in rich-text/图文混排 messages) with seven targeted crash fixes and a DB-performance/ANR pass. I traced the highest-risk areas end-to-end and found no correctness, security, data-loss, or build-breaking issues. The items below are non-blocking suggestions.

## What I verified

### 1. Mention-entity offset math (send → wire → render) ✅
Traced the full path: `WKRichTextSender.finish()` builds `blocks = [images..., text]`, `buildPlainFromBlocks` produces `[图片]...[图片]<text>`, and `adjustEntitiesForPlain()` shifts each entity offset by `imageCount * IMAGE_PLACEHOLDER.length()`. The receive side (`WKRichTextProvider`) accumulates `plainOffset` the same way (`+= IMAGE_PLACEHOLDER.length` per image block, `+= blockLen` per text block) and converts back to block-local offsets. The two sides agree, including for `>MAX_RENDER_IMAGES` cases (offset accumulation counts all image blocks; the cap only affects how many images render). No off-by-one. The descending-offset `ssb.replace()` with the `lastProcessedStart` guard correctly avoids corrupting earlier spans.

### 2. Seven crash fixes ✅
- **ContactEditText paste NPE** — `assert` replaced with real `null`/`getItemCount()==0` guards; paste still delegates correctly.
- **ChatActivity.refreshMsg** — the `list.get(i)` → captured `entity` refactor is behavior-preserving (`list == chatAdapter.getData()`); the revoke `postDelayed` lambda now re-reads the current list and bounds-checks both `previousIndex` and `nextIndex` before access — this is the actual IndexOutOfBounds fix and it is sound.
- **OPPO push** — `try/catch(Throwable)` + early return is correct: if `init()` fails (missing SDK class), `register()` would also fail, so skipping it avoids a cascade.
- **RemindMemberAdapter ClassCast** — `instanceof` guard before the `NoEventRecycleView` cast; `setItemCount` is an optimization-only call, safe to skip.
- **ProhibitWordDB / ConversationDbManager / ReminderDBManager** — `isClosed` checks, `cursor.use {}`, and `try/catch` around queries with safe defaults (null / false / empty) correctly prevent connection-pool-closed crashes.
- **MyHeadPortrait / UserDetail** — `isFinishing() || isDestroyed()` guards in the `ActivityResultLauncher` callbacks are the standard lifecycle-safe pattern.

### 3. ANR / DB-performance changes ✅
- `getMsgExtrasForChannels` batches N single queries into one (chunked at 500, under SQLite's 999 var limit, with an empty-list guard). The map is keyed `channelID:channelType` and every call site looks up with the same composite key, so no conversation loses its draft/extra. (The batch SQL filters by `channel_id` only and over-fetches rows of other `channel_type`s, but those land under distinct keys and are never mis-resolved — no data loss.)
- `getChannelIfCached` reads a `ConcurrentHashMap` — thread-safe for the render path.
- `getMemberOfFromIfCached()` intentionally skips the DB load to avoid main-thread queries during batch refresh; off-screen/uncached members are simply not updated in that pass and load normally on bind. Acceptable trade-off for the ANR fix.

### 4. ProGuard / version
`-keep class com.chat.uikit.thread.msgmodel.**` and the `1.3.1 (32) → 1.3.2 (33)` bump are correct.

## Non-blocking suggestions

- **(nit) Sentinel filtering symmetry** — `WKRichTextSender.adjustEntitiesForPlain()` drops `-1`/`-2` from `entities`, but `MentionEntityHelper.mergeMentionEntities()` (receive) does not. The renderer already guards `isSentinel` so there is no user-visible bug, but filtering sentinels on the receive merge too would keep the invariant consistent end-to-end.
- **(nit) Silent `catch` blocks** — a couple of the new DB `catch` blocks swallow without logging. Most fixes here do log (`WKLoggerUtils` / `Log.w`); applying the same to the remaining ones would preserve Bugly/diagnostic visibility.
- **(nit) Stable ordering in `buildMentionSpans`** — `sortedByDescending { it.offset }` is not stable; a secondary sort key would make same-offset handling deterministic. Practically unreachable since the composer prevents two mentions at the same offset.

## Test plan
The PR's manual test checklist is appropriate and covers the key regression surfaces (plain-text mention non-regression, no-mention rich-text rendering, paste-on-empty-clipboard, OPPO init failure, conversation-list/refresh ANR, and release-build obfuscation).

Jerry-Xin (approved):
R4: All R1-R3 blockers resolved. LGTM.

PR #52, opened 2026-06-08
ci: pass TRIAGE_WEBHOOK_URL to octo-issue-notify

lml2468:
The issue notification caller only forwarded OCTO_BOT_TOKEN, so the triage webhook mode was silently skipped and the triage agent was never invoked. Forward the org-level TRIAGE_WEBHOOK_URL secret so triage runs on new issues.

### DIFF
--- a/.github/workflows/octo-issue-feed.yml
+++ b/.github/workflows/octo-issue-feed.yml
@@ -18,3 +18,4 @@ jobs:
       event_action: ${{ github.event.action }}
     secrets:
       OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}
+      TRIAGE_WEBHOOK_URL: ${{ secrets.TRIAGE_WEBHOOK_URL }}


Jerry-Xin (approved):
This PR is relevant to `Mininglamp-OSS/octo-android` because it updates the repository’s issue automation workflow, and the change correctly forwards the optional triage webhook secret to the reusable issue notification workflow.

💬 Non-blocking
- 🔵 Suggestion: Confirm the org-level `TRIAGE_WEBHOOK_URL` secret is granted access to this repository; otherwise `.github/workflows/octo-issue-feed.yml:21` will pass an empty value and triage will remain disabled operationally.

✅ Highlights
- `.github/workflows/octo-issue-feed.yml:21` matches the reusable workflow contract: `octo-issue-notify.yml@v1` declares `TRIAGE_WEBHOOK_URL` as an optional `workflow_call` secret and consumes it for the triage webhook.
- The change is narrowly scoped and does not broaden workflow permissions.
- No blocking security, correctness, performance, or style issues found.

Validation note: I reviewed the local workflow context and the referenced reusable workflow definition. `actionlint` is not installed in this environment, so I did not run automated workflow linting.

yujiawei (approved):
# Code Review — PR #52 (octo-android)

## Summary

One-line additive change to `.github/workflows/octo-issue-feed.yml`: forward the org-level `TRIAGE_WEBHOOK_URL` secret into the reusable `octo-issue-notify.yml@v1` workflow so the optional triage-agent step runs on new issues. The diff is minimal and the intent matches the stated problem (the secret was previously not forwarded, so the triage webhook mode was silently inactive).

## Verification (cross-file dependency)

Because this is a CI/secret wiring change, the key risk is a mismatch between the caller and the reusable workflow it pins to. I verified the full chain at the pinned ref:

- ✅ **Secret is declared by the callee.** `Mininglamp-OSS/.github/.github/workflows/octo-issue-notify.yml@v1` declares `TRIAGE_WEBHOOK_URL` under `secrets:` with `required: false`. Passing it from the caller is therefore valid — it will not error as an "unexpected secret".
- ✅ **Secret is actually consumed.** The callee forwards it to the shared `octo-notify@v1` composite action via `webhook-url: ${{ secrets.TRIAGE_WEBHOOK_URL }}`, which enables "webhook mode" and POSTs the triage payload. So the wiring closes the loop end-to-end; this is not a no-op.
- ✅ **Graceful when unset.** If the org/repo secret is not configured, `secrets.TRIAGE_WEBHOOK_URL` resolves to an empty string. The composite action gates webhook mode behind `if webhook_url:`, so an empty value simply skips the webhook with no failure. IM notification continues to work independently. No build/run break is introduced by merging this.

## Security notes (security-sensitive PR)

- ✅ The capability URL is passed through the `secrets:` mechanism, never hardcoded and never interpolated into a `run:` shell. Downstream it is consumed via `env:` in the composite action, and the only log line is a static `webhook` label — the URL is not echoed.
- ✅ The triage payload built by the callee is minimal (repo, number, title, url, author, action) with control-character sanitization; the issue body is fetched by the agent via the GitHub API rather than egressed through Actions. Nothing in this PR weakens that.
- 🔎 **For a human to confirm (ops, not code):** this change only has an effect once `TRIAGE_WEBHOOK_URL` is configured as an org/repo Actions secret. Its absence is safe (feature stays dormant), so this is a deployment follow-up, not a merge blocker.

## Non-blocking observations

- The caller still subscribes to `issues: [opened, reopened]`, while the reusable workflow centrally forwards only `opened` (it skips other actions by design). This is pre-existing and unrelated to this PR; no action needed here.

## Verdict

No P0/P1 issues. The change is correct, minimal, fails safe when the secret is absent, and the caller↔callee secret contract is satisfied. Approving.

PR #67, opened 2026-06-22
fix: 多项 bug 修复与功能优化（聊天/子区/转发/通知/登录/群管理）

### LINKED ISSUE #65: 【Bug】合并转发消息创建子区时，消息内容未正确转发，显示为 [聊天记录]
## 问题描述

在 Android 端，针对一条**合并转发**的消息，执行**子区创建**操作时，出现 bug：合并转发的消息内容没有被正确转发到子区，而是显示为一条 `[聊天记录]` 占位文本。

## 复现步骤

1. 在任意会话中找到一条**合并转发**类型的消息
2. 长按该消息，选择**创建子区**
3. 进入新创建的子区
4. 查看子区内的消息内容

## 实际行为

子区内的消息显示为 `[聊天记录]`，合并转发的原始内容未正确携带。

## 预期行为

子区内应正确展示合并转发的消息内容（与直接转发到普通会话的效果一致）。

## 环境

- 平台：Android
- 功能：合并转发 + 子区创建

## 备注

by @[27mZV61NBBAf44b4ab9_bot:猴小宝] 代提

### LINKED ISSUE #66: [Android][Bug] 新注册用户停留在邀请码页面时，无退出登录入口
## 问题描述

新注册用户登录后，停留在「输入邀请码加入团队」页面时，页面上只有「加入」按钮和「← 返回」，**没有退出登录的入口**，无法切换账号。

## 复现步骤

1. 使用新注册账号登录 App
2. 进入「欢迎使用 Octo！输入邀请码加入团队」页面
3. 发现无法退出登录或切换账号

## 期望行为

参考 Web 端，在该页面增加「退出登录」按钮，方便用户切换账号。

## 截图

**iOS 端（缺少退出登录入口）：**
ios-no-logout

**Web 端对比（有退出登录按钮）：**
web-contrast

## 环境

- 平台：Android
- 场景：新注册用户首次登录，尚未加入任何团队空间

mzanalyticst:
## Summary

本次合集包含 17 个 commit，覆盖聊天、子区、转发、通知、登录、群管理、智能总结等模块的 bug 修复与功能增强。release 包已本地 `assembleDevRelease` 编译通过，R8 + shrinkResources 均启用、无新增 proguard 规则需求。

Closes #65
Closes #66

### Bug 修复
- `2b67a96` fix(login,thread): 登录页协议链接走 BuildConfig + 子区创建透传原 payload — **修复 #65**（合并转发创建子区显示 `[聊天记录]`），顺带修了图片/文件/语音等富类型同样被降级为纯文本的问题
- `028914c` fix(login): 邀请码页加退出登录入口 — **修复 #66**
- `a4ecca9` fix(mention): 接收侧用 mentionInfo.uids + plain 权威重建 mention entities
- `00f1e82` fix(thread): 子区被创建者删除/关闭后，源消息长按菜单回退到「创建子区」
- `3dfb252` fix(chat): markdown 表格消息时间错位修复（时间视图搬到 bubble 末尾右下角）
- `783ae80` fix(chat): markdown 表格 header 前缺空行不识别 + 表格后文本段长按菜单缺失
- `9a8ca20` fix(ui): 未读数胶囊文字按 paint metrics 动态居中
- `989e8f3` fix(notify): 跳转改 app 级通知页 + 清理排查日志
- `4934a55` fix(notify): 通知权限引导加冷却 + 诊断日志
- `2bee5ff` fix(forward): 逐条转发文件/合并消息丢失类型
- `fd9bd41` fix(summary): 智能总结链路问题修复

### 功能增强
- `ab186b1` feat(input): 选图改为输入框上方内联缩略图带，移除全屏 caption 编辑页
- `10cb17c` feat(emoji,msgeffect): 表情面板按使用时间排序 + 👍 特效优化
- `ba72bac` feat(msgeffect): 替换 [崇尚行动] 优化版视频 + 新增 [有品位] 动画
- `6467b02` feat(group): 群消息推送（入站 Webhook）模块
- `f03fd85` feat(thread): 已建子区的消息长按改为「进入子区」入口
- `31261ca` feat(group): 转让群主功能

## 混淆 / Release 风险审查

新增代码扫描，无需新增 keep 规则：

| 风险点 | 现状 |
|---|---|
| 新 Retrofit 接口 `IncomingWebhookService` / `GroupService` 新方法 | Retrofit 3.0.0 consumer-rules 自动 keep |
| 数据模型 `IncomingWebhook` | `fromJson(JSONObject)` 走字符串字面量 key + 直接 field 访问，R8 安全 |
| `WKThreadCreatedContent` | 已被 `-keep class com.chat.uikit.thread.msgmodel.**` 覆盖 |
| `JSON.toJSONString(body)` | 只在 `BuildConfig.DEBUG` 调试 log 中，release 不执行 |
| `Class.forName` / `@JSONField` / `@SerializedName` 新增 | 零命中 |

## Test Plan
- [x] `./gradlew :app:assembleDevRelease` 本地编译通过（minifyEnabled=true, shrinkResources=true）
- [ ] 真机回归 #65：合并转发消息长按创建子区，确认子区内显示原始内容（非 `[聊天记录]`）
- [ ] 真机回归 #66：新注册账号停留在邀请码页面，确认右上角有「退出登录」入口可切换账号
- [ ] 真机回归图文混排：输入框上方内联缩略图带选图 / 删图 / 加图、点发送
- [ ] 真机回归群管理：入站 Webhook 创建/编辑/启停/删除/复制 URL；转让群主流程
- [ ] 真机回归子区：长按已建子区消息进入子区入口；子区被删除/关闭后长按回退到「创建子区」
- [ ] 真机回归通知：权限引导冷却生效；通知跳转到 app 级通知页
- [ ] 真机回归智能总结：列表/详情/创建链路


Jerry-Xin (changes requested):
This PR is relevant to octo-android, but the new webhook flow has a blocking regression in the empty-state create path.

**🔴 Blocking**
- 🔴 Critical: Creating the first webhook from an empty list does not refresh the list when the user returns. The empty-state CTA opens the edit screen at [ChannelWebhookActivity.java:126](/Users/mlamp/.openclaw/tmp/pr-review-ap1h0_no/repo/wkuikit/src/main/java/com/chat/uikit/group/webhook/ChannelWebhookActivity.java:126), and create success starts the one-time URL screen then finishes at [ChannelWebhookEditActivity.java:224](/Users/mlamp/.openclaw/tmp/pr-review-ap1h0_no/repo/wkuikit/src/main/java/com/chat/uikit/group/webhook/ChannelWebhookEditActivity.java:224). When the user closes the URL screen, `ChannelWebhookActivity.onResume()` only reloads if `!data.isEmpty()` at [ChannelWebhookActivity.java:163](/Users/mlamp/.openclaw/tmp/pr-review-ap1h0_no/repo/wkuikit/src/main/java/com/chat/uikit/group/webhook/ChannelWebhookActivity.java:163). Since this was the first webhook, `data` is still empty, so the new webhook remains invisible and the screen stays in empty state. Remove that gate, or track a dirty/result flag from edit/create/url and reload on return.

**💬 Non-blocking**
- 🟡 Warning: I could not run the requested compile check because this environment has no Java runtime installed. `./gradlew :app:compileDevDebugKotlin :wkuikit:compileDevDebugKotlin :wkbase:compileDevDebugKotlin` failed before Gradle startup with “Unable to locate a Java Runtime.”
- 🔵 Suggestion: After transferring group ownership, the parent `GroupDetailActivity` is launched with plain `startActivity()` at [GroupDetailActivity.java:275](/Users/mlamp/.openclaw/tmp/pr-review-ap1h0_no/repo/wkuikit/src/main/java/com/chat/uikit/group/GroupDetailActivity.java:275), while the transfer screen only triggers member sync at [TransferGroupOwnerActivity.java:248](/Users/mlamp/.openclaw/tmp/pr-review-ap1h0_no/repo/wkuikit/src/main/java/com/chat/uikit/group/TransferGroupOwnerActivity.java:248). Consider also fetching channel info or using an activity result to force parent refresh, so owner-only controls do not remain stale if sync is delayed.

**✅ Highlights**
- The PR is in scope for this repository.
- The rich-text tray and thread source-payload changes are generally well scoped and preserve existing fallback paths.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-android PR #67 Review Report
**PR:** Mininglamp-OSS/octo-android #67 — 聊天/子区/转发/通知/登录等多项修复与功能优化
**Head SHA:** ab186b1736b9cf82ac7dece21ec8aa481fccd7d6
**Scope:** 82 files, ~7700 lines diff

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| EmojiFragment ClassCastException fix | ✅ | Correct cast to EmojiEntry at line ~85 |
| WKMarkwonProvider table delimiter detection | ✅ | isTableDelimiterLine() + empty-line injection for GFM |
| MentionEntityHelper @-highlight reconstruction | ✅ | indexOf-based claiming works at text end; sentinel filtering correct |
| ChatActivity WKChannelType→OctoSourceType mapping | ✅ | PERSONAL(1)→3, COMMUNITY_TOPIC(5)→2, else→1; single caller verified |
| SmartSummaryCreateActivity.newIntentForChat callers | ✅ | Exactly 1 caller (ChatActivity:353), passes correctly mapped value |
| MessageForwardSupport whitelist | ✅ | Mirrors iOS WKApp.allowForwards pattern |
| TransferGroupOwnerActivity | ✅ | Admin-only gate, hidden self+bot users, single-select |
| IncomingWebhookManager CRUD + cooldown | ⚠️ | Functional but has concurrency/lifecycle concerns (see findings) |
| SummaryActionToast lifecycle hosting | ✅ | ActivityLifecycleCallbacks pattern is sound |
| SpaceGuideActivity logout button | ✅ | Reasonable escape hatch for stuck users |
| WKLoginActivity terms/privacy URLs | ✅ | Now reads from WKApiConfig instead of hardcoded |
| TabActivity 7-day permission cooldown | ✅ | SharedPreferences-based, reasonable |

---

## 2. Findings

### F1 — WKThreadCreatedProvider: thread closed on ANY error (potential over-aggressive suppression)

**Severity:** P1 (potential)
**Diff-scope:** NEW — `markThreadClosedForSourceMessageId()` is a new method, called from new error-handling paths.

**Description:** `WKThreadCreatedProvider.kt` calls `markThreadClosedForSourceMessageId()` both when `status==3` (explicitly closed) AND on any API error. If "any error" includes transient network failures (timeout, 500, DNS failure), a thread that is still open on the server would be permanently marked as closed in the process-level map. All subsequent thread messages from that source message would be suppressed for the rest of the app session.

**Data flow trace:**
- `WKThreadCreatedContent.sourceMessageThreadMap` (ConcurrentHashMap, line 52) stores closed-thread state
- `markThreadClosedForSourceMessageId()` (line 117) puts into this map
- `WKThreadCreatedProvider.kt` line ~104 and ~116: calls mark on status==3 OR error
- Once marked, `isThreadClosedForSourceMessageId()` returns true → thread card hidden

**Impact:** User-visible — a thread that exists and is open on the server becomes invisible in the UI until app restart. This is worse than the pre-PR behavior (where the thread card would remain visible).

**Recommendation:** Only mark closed on definitive signals: `status==3` or HTTP 404. Transient errors (timeout, 5xx, network unreachable) should NOT close the thread. Consider logging the error and retrying on next bind.

**⚠️ Confidence note:** I could not fully verify the exact error-handling code path from the diff alone. If the "error" path only fires on definitive 404/deleted responses (not transient errors), this is not a bug. Please confirm the error classification in the provider.

---

### F2 — IncomingWebhookManager: hasTestInFlight can latch on exception

**Severity:** P2
**Diff-scope:** NEW — entire IncomingWebhookManager is new code.

**Description:** `hasTestInFlight` (boolean, line ~45 of IncomingWebhookManager.java) is set `true` before a test-send network call and reset `false` in the success/error callback. If the network layer throws a synchronous exception before the callback is registered (e.g., malformed URL, immediate Retrofit rejection), the flag stays `true` permanently. The 3-second per-webhook cooldown (`cooldownExpire`) provides partial mitigation, but the global in-flight guard would block ALL webhook test sends until app restart.

**Data flow trace:**
- `hasTestInFlight = true` set at call site before `IncomingWebhookService.testWebhook()` invocation
- Reset in `onNext`/`onError` RxJava callbacks
- If `subscribe()` itself throws (before any callback), reset never happens

**Recommendation:** Wrap the network call in try/finally, or reset `hasTestInFlight` in the RxJava `doFinally`/`doOnTerminate` operator.

---

### F3 — GroupDetailActivity.WEBHOOK_PREFETCH_IN_FLIGHT: permanent leak on callback miss

**Severity:** P2
**Diff-scope:** NEW — webhook prefetch is new code.

**Description:** `WEBHOOK_PREFETCH_IN_FLIGHT` is a `static final` synchronized HashSet (line 710). The `groupNo` is added before the network call and removed inside the callback. If the callback never fires (process killed mid-request, network stack failure), the groupNo remains permanently, and future visits to that group's detail page will skip the webhook count prefetch silently.

**Recommendation:** Add a TTL-based expiry or remove the entry in `onDestroy()` of the activity. Alternatively, use a `ConcurrentHashMap<String, Long>` with timestamp and check staleness.

---

### F4 — WKRichTextTrayAdapter.richTextTraySendingGate() is dead code

**Severity:** P2 (maintainability)
**Diff-scope:** NEW — the gate function is new.

**Description:** `richTextTraySendingGate()` at line 225 always returns `false`. The comment explicitly states it's a stub and the real guard lives in `ChatPanelManager.richTextTraySending`. Two call sites exist (ThumbViewHolder remove button, AddViewHolder click), both in the same file. If `ChatPanelManager`'s guard is ever removed or refactored, this stub would silently allow double-sends with no protection.

**Recommendation:** Either wire the stub to the real guard (delegate to `ChatPanelManager.richTextTraySending`) or add a `@Deprecated` / `TODO` annotation making the dependency explicit.

---

### F5 — SmartSummaryListActivity: SummaryEvents collected without repeatOnLifecycle

**Severity:** P2
**Diff-scope:** NEW — SummaryEvents collection is new code.

**Description:** `SmartSummaryListActivity` collects from `SummaryEvents` SharedFlow inside `lifecycleScope.launch { ... }` without `repeatOnLifecycle(STARTED)`. The collector remains active when the activity is stopped (but not destroyed). If `postListShouldRefresh()` fires while the activity is in the background, it triggers `reloadAndScrollToTop()` which calls `submitList` + `scrollToPosition(0)` on a stopped RecyclerView. This may cause visual glitches or, in edge cases, IllegalStateException if the layout manager is mid-detach.

**Recommendation:** Use `lifecycleScope.launch { repeatOnLifecycle(Lifecycle.State.STARTED) { SummaryEvents.collect { ... } } }` to only process events when the activity is visible.

---

### F6 — WKThreadCreatedContent.sourceMessageThreadMap: unbounded process-level growth

**Severity:** P2 (low impact)
**Diff-scope:** NEW — the map and its lifecycle policy are new.

**Description:** `sourceMessageThreadMap` (ConcurrentHashMap, line 52) and `sourceMessageIdSet` grow on every `decodeMsg` and are only pruned via explicit `markThreadClosedForSourceMessageId`. The code comment at lines 47-50 explicitly states "进程级别永不清空" (never cleared within process lifetime). In a long session with many threads across many groups, this accumulates without bound. Each entry holds a full WKThreadCreatedContent object with Parcel overhead.

**Recommendation:** Consider an LRU eviction (e.g., cap at 500 entries) or periodic cleanup tied to conversation list visibility.

---

### F7 — WKTextProvider.relocateMsgTimeView: parent-swapping in RecyclerView

**Severity:** nit
**Diff-scope:** NEW — relocateMsgTimeView() is new.

**Description:** `relocateMsgTimeView()` detaches the msgTimeView from its original parent and re-attaches it to the bubble layout for table messages. In a RecyclerView, views are recycled; if a recycled view previously had its time view relocated, the next bind may find the time view in the bubble layout instead of the original parent. The null guards in `ChatTextTimeLayout` mitigate crash risk, but the pattern is fragile.

**Recommendation:** Consider using a ViewStub or conditional layout inflation instead of runtime parent-swapping.

---

## 3. Suggestions

1. **F1 (thread closure on error):** This is the highest-priority item. Verify the error classification in WKThreadCreatedProvider — if transient errors trigger closure, add error-type discrimination before calling `markThreadClosedForSourceMessageId`.

2. **F2 (hasTestInFlight):** Use RxJava `doFinally` to guarantee flag reset. One-line fix.

3. **F3 (prefetch leak):** Replace HashSet with ConcurrentHashMap<String, Long> and check `SystemClock.elapsedRealtime()` staleness on read.

4. **F5 (lifecycle):** Standard `repeatOnLifecycle` fix — well-documented pattern.

---

## 4. Additional Findings

- **Positive:** The `MentionEntityHelper.reconstructMentionEntitiesFromPlain()` implementation is well-designed. The claimed-index approach with BooleanArray correctly handles overlapping mentions, end-of-text mentions, and sentinel UIDs. Data flow from `WKMentionInfo.uids` through display-name resolution to entity reconstruction is sound.

- **Positive:** The WKChannelType→OctoSourceType mapping in ChatActivity is correct and verified against the SDK byte values (PERSONAL=1, GROUP=2, CUSTOMER_SERVICE=3, COMMUNITY=4, COMMUNITY_TOPIC=5). The single caller of `newIntentForChat` passes the correctly mapped value.

- **Positive:** `ThumbsUpEffect` bubble hit detection uses screen-space rectangle intersection over RecyclerView children — robust against null/missing views with proper early returns.

- **Positive:** `sendVoiceTextDirect` in ChatPanelManager correctly constructs mention-bearing messages from STT output, following the same entity/mention-info pattern as the regular text send path.

- **Observation:** `ChannelWebhookUrlActivity.shellQuote()` uses standard `'\''` escaping for single-quoted shell strings — correct and safe for user-defined webhook names in curl examples.

---

## 5. Data Flow Tracing

| Consumed Data | Upstream Source | Verified Flow |
|---|---|---|
| mentionInfo.uids in MentionEntityHelper | WKMentionInfo from wire protocol or send-side scanPlainTextMentions | ✅ uids flow through sentinel filter → display-name lookup → indexOf claiming |
| originChannelType in SmartSummaryCreateActivity | ChatActivity.openSummaryCreate → WKChannelType byte → mapped Int | ✅ Single caller, correct mapping (1→3, 5→2, else→1) |
| hasTestInFlight in IncomingWebhookManager | Set before network call, cleared in callback | ⚠️ Gap: synchronous exception before callback registration leaves flag stuck |
| sourceMessageThreadMap entries | WKThreadCreatedContent.decodeMsg (add), markThreadClosedForSourceMessageId (remove) | ⚠️ No bounded eviction; explicit removal only |
| SummaryEvents SharedFlow | SmartSummaryDetailViewModel posts events | ⚠️ Collected without repeatOnLifecycle; processes events while stopped |
| msgTimeView parent in WKTextProvider | relocateMsgTimeView swaps parent at bind time | ⚠️ Fragile under RecyclerView recycling |

---

## 6. Blind-spot Checklist (R5)

- **C1 双路径 parity:** N/A — no symmetric add/remove or subscribe/unsubscribe paths touched.
- **C2 control-flow ordering / 嵌套复用:** N/A — no shared emit/sanitize functions with multiple call sites at risk.
- **C3 授权边界 ≠ 能力边界:** N/A — no new endpoints/tools exposed to unauthenticated callers (webhook endpoints are server-side; Android is a client).
- **C4 授权生命周期 / 容器-成员状态级联:** N/A — no auth changes in this PR.
- **C5 build/note ≠ 运行期路径正确:** N/A — no build artifacts, extensions, or relative-URL changes.
- **C6 治理/策略/安全文档自洽性:** N/A — no governance/policy documents changed.

---

## 7. Cross-round Blocker Recheck (R6)

N/A — first review of this PR.

---

## Verdict

**[Octo-Q] verdict: APPROVE**

This is a large, well-structured PR covering chat fixes, rich text refactoring, message effects, forwarding, thread management, group transfer, incoming webhooks, and smart summary improvements. The code quality is generally high with thoughtful patterns (ActivityLifecycleCallbacks for toast hosting, ConcurrentHashMap for thread-safe caches, proper mention entity reconstruction).

**F1 (thread closure on any error) is the most important item to verify** — if transient errors trigger `markThreadClosedForSourceMessageId`, it would be a P1 user-visible regression. The remaining findings (F2-F7) are P2/nit: real but non-blocking concerns that can be addressed in follow-up.

No P0 issues found. F1 is flagged as "potential P1" pending confirmation of error classification in WKThreadCreatedProvider.

yujiawei (changes requested):
# Code Review — PR #67 (octo-android)

Reviewed at head SHA `ab186b1` against merge-base with `main`. Scope: 82 files, +5167/-1015 — chat compose tray rework, group-owner transfer, incoming webhooks, thread "enter existing thread" menu, cross-end mention highlighting, notification-permission cooldown, markdown table boundaries, and several UI/effect fixes.

Overall this is a careful, well-documented PR with strong iOS parity notes and good defensive coding (null-safety, in-flight guards, lifecycle ordering). I have **one functional regression** that should be addressed before merge, plus a few non-blocking notes.

## Blocking

### P1 — Cross-end mention rebuild can drop valid mentions on multi-mention messages
`wkbase/src/main/java/com/chat/base/msg/MentionEntityHelper.kt:120-183`, called unconditionally on every message bind at `wkbase/src/main/java/com/chat/base/msgitem/WKUIChatMsgItemEntity.java:168`.

`reconstructMentionEntitiesFromPlain` discards **all** wire `type=mention` entities (including the correct, plain-relative ones that `mergeMentionEntities` just attached at line 167) and rebuilds them by scanning the plain text for the **receiver's** local display names (`memberRemark` / `memberName` / personal channel name) of each uid (`collectMentionNameCandidates`, lines 190-201).

The problem: at send time the composer inserts `@<sender's local remark or memberName>` into the text (see `SearchWithMemberActivity.java:74` — `memberRemark` takes precedence). When the sender used a custom remark that differs from the name the receiver sees for that uid, the rebuild's `indexOf("@<receiver-name>")` finds nothing for that uid.

Trace of the loss (lines 175-182):
```kotlin
val nonMention = (model.entities ?: emptyList()).filter { it.type != ChatContentSpanType.mention }
if (rebuilt.isEmpty() && nonMention.isEmpty()) return  // safe: pure-text, total miss → wire mentions kept
model.entities = ... nonMention + rebuilt   // wire mentions NOT in rebuilt are permanently dropped
```
- Single-mention total-miss on a plain-text message is **safe** (early return keeps the merged wire entities).
- But a message with ≥2 mentions where **some** uids resolve and some don't (e.g. `@AliasOnlySenderSet` + `@CanonicalName`) takes the non-early path: `rebuilt` contains only the matched uid(s), and the unmatched uid's correct wire mention is lost. Result: that `@name` renders as plain text — no highlight, no tap-to-profile — for a message that rendered correctly before this PR.

Impact is limited to highlight/tap (the `mentionInfo.uids` payload is untouched, so **push/notification delivery is not affected**), but it is a visible regression of a shipped feature on native Android→Android messages, triggered by the common "custom group remark + multi-mention" case.

The root cause is over-broadening: the PR's actual target is iOS-sent **richText/caption** messages whose wire offsets are caption-relative, but the fix is applied to every message type and **replaces** authoritative wire data instead of only filling gaps. Suggested directions (any one):
- Gate the rebuild to `WKContentType.richText` (the only type with the caption-relative offset mismatch), or
- Only rebuild a mention when the wire entity is missing or its `offset/length` falls outside the plain-text bounds — keep valid wire entities as-is, or
- When a uid cannot be re-matched, **retain** its original wire entity rather than dropping it.

The author's own KDoc hedges that wire offsets "可能是 caption-relative" (*might be*) — making the unconditional wholesale replace the weak point.

## Non-blocking (P2 / nits)

- **Webhook & owner-transfer authorization is client-side only.** `ChannelWebhookActivity.canManageWebhook` (`:226`) and the transfer/webhook entry gating in `GroupDetailActivity` (`memberRole` checks) correctly mirror iOS, but they are UI gates. This is expected for a client PR; just confirming the server enforces creator/manager ownership on `POST groups/{groupNo}/transfer/{toUid}`, the webhook CRUD endpoints, and `test` — the client cannot be the trust boundary. No change requested.
- **`IncomingWebhookManager` test state machine isn't synchronized.** `hasTestInFlight` and `cooldownExpire` (`service/IncomingWebhookManager.java`) are mutated from network callbacks without `volatile`/locking. In practice Retrofit/BaseObserver delivers on the main thread so the serial+cooldown guard holds, but a one-line `volatile` on `hasTestInFlight` would make the "only one test in flight" invariant explicit. Nit.
- **`BubblePulseHelper.pulse` on recycled views** (`msgeffect/BubblePulseHelper.kt`): cancelling the prior animator clears the tag in `onAnimationCancel` but does not reset `scaleX/Y`; the next `pulse` resets pivot/scale on start, and RV rebinds reset transforms, so this is benign — flagging only as something to keep an eye on if cells are ever reused mid-pulse without a rebind.
- **`reconstructMentionEntitiesFromPlain` cold-cache behavior**: even with the P1 fix, on first render before the member cache is warm, `collectMentionNameCandidates` returns empty and mentions won't rebuild until a later bind. The early-return guard mitigates the pure-text case; just noting the dependency on member-cache warmth.
- **Webhook list goes stale after creating the first webhook from the empty state** — `wkuikit/src/main/java/com/chat/uikit/group/webhook/ChannelWebhookActivity.java:162` (`onResume`). The reload guard is `if (adapter != null && !data.isEmpty())`. When a group starts with zero webhooks and the user creates one from the empty-state CTA, the create page finishes and the one-time URL page shows on top of this still-empty list; on returning, `data.isEmpty()` is still true, so `loadList(false)` is skipped and the list (and the count subtitle cache) stay on the empty state until the page is left and reopened — the just-created webhook appears missing. Dropping the `!data.isEmpty()` condition (always reload on resume) fixes it. Minor first-run UX, not data loss.

## Verified good

- `WKRichTextCaptionActivity` removal is clean — no dangling references to the activity, launcher, layout binding, or manifest entry remain (grep-confirmed); manifest correctly drops it and adds the new activities.
- New layout binding IDs referenced from code (`transferOwnerLayout`, `incomingWebhookLayout`, `incomingWebhookSubtitleTv`, `bubble_pulse_animator_tag`, transfer/webhook layouts and strings) all resolve.
- `WKThreadCreatedProvider` change correctly stops the string-match fallback from opening dead thread channels on 404; `markThreadClosedForSourceMessageId` is wired at all three documented call sites.
- Notification-permission cooldown (`TabActivity.checkNotificationPermission`) and `WKOSUtils.openChannelSetting` → app-level settings fix are sound and address the HarmonyOS `areNotificationsEnabled()==false` re-prompt loop with a proper try/fallback.
- `EmojiFragment` `(String)`→`EmojiEntry` cast fix and prefs canonicalization are correct; recent-emoji ordering is consistent between `ChatPanelManager` and `EmojiFragment` (both key on emoji `.text`).
- Markdown table boundary fix (`WKMarkwonProvider.ensureTableTermination`) handles **both** the before-header and after-table empty-line insertion — verified the after-table branch is still present (a concern raised during review was a false alarm).
- Login terms/privacy URLs moved from hardcoded placeholders to `WKApiConfig`, matching existing usage across the app.
- `transferGroupOwner` reading `result.status` matches the established `GroupModel` convention (10+ sibling endpoints do the same); the webhook endpoints' "treat onSuccess as success" handling is correctly scoped to those endpoints only.

## Coverage notes (not verified from the diff)

- Server-side authorization for transfer/webhook endpoints (out of repo).
- The actual rendering of partial-mention messages on a device (the P1 trace is static; a runtime repro with a custom remark + two mentions would confirm).
- Binary asset swaps (`action_celebrate.mp4`, `classy_celebrate.mp4`) were not inspected.
- Summary list/detail flows and `LumaKeyVideoEffectPlayer` rename were reviewed for interface consistency but not run.

Jerry-Xin (changes requested):
This PR is in scope for `octo-android`, but it has a blocking RecyclerView reuse regression in chat message rendering.

**🔴 Blocking**

🔴 Critical: `WKTextProvider` does not reset `contentTv` visibility when binding a non-table message after a table-starting message.

In [WKTextProvider.kt](/Users/mlamp/.openclaw/tmp/pr-review-r_55ezvd/repo/wkuikit/src/main/java/com/chat/uikit/chat/provider/WKTextProvider.kt:349), table messages whose first segment is blank set `contentTv.visibility = View.GONE`. When that same holder is later reused for a normal message, the no-table branch at [WKTextProvider.kt](/Users/mlamp/.openclaw/tmp/pr-review-r_55ezvd/repo/wkuikit/src/main/java/com/chat/uikit/chat/provider/WKTextProvider.kt:279) only assigns `contentTv.text` and relocates the time view; it never restores `contentTv.visibility = View.VISIBLE`. This can make ordinary text messages render with hidden content after scrolling.

The placeholder-mismatch table fallback at [WKTextProvider.kt](/Users/mlamp/.openclaw/tmp/pr-review-r_55ezvd/repo/wkuikit/src/main/java/com/chat/uikit/chat/provider/WKTextProvider.kt:319) has the same reuse risk. Reset the visibility before assigning text in both branches.

**💬 Non-blocking**

🟡 Warning: I could not run the requested Gradle compile locally because this environment has no Java Runtime installed. The attempted command failed before Gradle started with: `Unable to locate a Java Runtime`.

**✅ Highlights**

The PR is clearly relevant to the project and the changed areas match the described Android chat, group, summary, notification, login, and thread functionality. The webhook and transfer-owner flows also follow existing Activity/ViewBinding patterns closely.

mochashanyao (changes requested):
[Octo-Q · automated review]

---

# octo-android PR#67 Review Report (automated review)
**Head SHA:** 7681b553246b4817f0a68ad41f4b6451d375eb0a  
**Scope:** 9 files as specified in task

---

## 1. Verification Summary

| File | Status |
|------|--------|
| MentionEntityHelper.kt | ⚠️ 1x P1 (data-flow gap) |
| MessageForwardSupport.kt | ✅ Clean |
| WKThreadCreatedContent.java | ⚠️ 1x P1 (memory leak risk) |
| CreateThreadActivity.java | ⚠️ 1x P2 (silent fallback) |
| ChatPanelManager.kt | ⚠️ 1x P1 (race condition), 1x P2 |
| WKRichTextTrayAdapter.kt | ✅ Clean |
| WKTextProvider.kt | ⚠️ 1x P2 (RV reuse edge) |
| ChatActivity.java | ✅ Clean (deletion-only) |
| WKUIChatMsgItemEntity.java | ✅ Clean |

---

## 2. Findings

### F1 — P1: MentionEntityHelper.reconstructMentionEntitiesFromPlain — `displayContent` may not match entity offsets for non-text messages

**File:** `wkbase/src/main/java/com/chat/base/msg/MentionEntityHelper.kt`, line 143  
**Diff-scope:** NEW (introduced in this PR)

**Issue:** The function uses `model.displayContent` as the `plain` text for entity offset validation and `indexOf` lookups. However, `displayContent` for non-text message types (e.g., RichText, image with caption) may include placeholder prefixes like `[图片]` that are NOT present in the original wire content where entity offsets were computed by the sender.

**Data-flow trace:**
- `model.displayContent` is set by each `WKMessageContent` subclass's `getDisplayContent()` method
- For `WKRichTextContent`, `getDisplayContent()` concatenates text blocks with image placeholders
- Entity offsets from iOS are caption-relative (as documented in the function's own comment)
- The function validates `plain[e.offset] == '@'` — but if `plain` has a `[图片]` prefix that the sender's caption didn't have, the offset check will incorrectly reject valid entities OR accept invalid ones

**Code snippet:**
```kotlin
val plain = model.displayContent ?: return  // line 143
// ...
val validWireMentions = existingMentions.filter { e ->
    e.offset >= 0
            && e.length > 0
            && e.offset + e.length <= plain.length
            && plain[e.offset] == '@'  // line 151 — offset may not align with displayContent
}
```

**Impact:** For RichText messages with leading images, the `[图片]` placeholder shifts all subsequent text positions. A valid caption-relative entity at offset 0 would point into the placeholder text, failing the `@` check and being incorrectly discarded. The gap-filler then tries to rebuild via `indexOf("@name")` in `displayContent`, which may find the `@` at a different position than the sender intended — or not find it at all if the placeholder shifts things enough.

**Mitigation already present:** The function is called after `mergeMentionEntities`, and for plain text messages (the common case) `displayContent == content`, so this only affects RichText cross-platform messages — which is exactly the scenario the function was designed to fix.

**Severity rationale (R1):** This can cause @ mentions to not highlight in RichText messages received from iOS — the exact scenario this function aims to fix. If the gap-filler itself has an offset mismatch, the user-visible bug persists.

---

### F2 — P1: WKThreadCreatedContent static maps grow unbounded — process-lifetime memory leak

**File:** `wkuikit/src/main/java/com/chat/uikit/thread/msgmodel/WKThreadCreatedContent.java`, lines 42-60  
**Diff-scope:** NEW (introduced in this PR)

**Issue:** `sourceMessageThreadMap` and `sourceMessageIdSet` are `static final ConcurrentHashMap` / `Set` that are populated in `decodeMsg()` but only cleaned by `markThreadClosedForSourceMessageId()`. The comment says "进程级别永不清空" (never cleared during process lifetime). For active users in many threads, these maps grow without bound:

- Each `WKThreadCreatedContent` object holds references to `content`, `from_uid`, `from_name`, `thread_name`, etc.
- `clearCache()` only clears `messageCountCache`, NOT `sourceMessageThreadMap` or `sourceMessageIdSet`
- There is no LRU eviction, no size cap, no periodic cleanup

**Code snippet:**
```java
private static final ConcurrentHashMap<String, WKThreadCreatedContent> sourceMessageThreadMap =
        new ConcurrentHashMap<>();

private static final Set<String> sourceMessageIdSet =
        Collections.newSetFromMap(new ConcurrentHashMap<>());
```

**Data-flow trace:**
- `decodeMsg()` is called for every thread-created system message during message list loading
- Each call does `sourceMessageThreadMap.put(srcIdStr, this)` — storing the full content object
- `markThreadClosedForSourceMessageId()` is only called on explicit close/delete actions
- Threads that are simply abandoned (never closed by creator) stay in the map forever

**Impact:** In a busy workspace with hundreds of threads over time, this accumulates hundreds of `WKThreadCreatedContent` objects in static memory. Each holds string fields. Not a crash, but a steady memory pressure increase that contributes to OOM risk on low-memory devices.

**Severity rationale:** P1 rather than P0 because it's a slow leak (not an immediate crash), but it's user-visible in the form of increased memory footprint over time. The iOS counterpart (`dispatch_once` singleton) has the same design, but iOS has ARC and memory warnings that trigger cleanup; Android's static fields survive memory warnings.

---

### F3 — P1: ChatPanelManager voice+image send path — no `richTextTraySending` guard on image burst

**File:** `wkuikit/src/main/java/com/chat/uikit/chat/ChatPanelManager.kt`, lines 2458-2468  
**Diff-scope:** NEW (introduced in this PR)

**Issue:** In `onSendVoice`, after sending the voice message, the code snapshots `richTextTray.orderedPaths()`, clears the tray, then sends each image as a separate `WKImageContent`. However, unlike `flushRichTextTraySend()` which sets `richTextTraySending = true` to prevent re-entry, this voice+image path does NOT set the guard. If the user taps send again quickly (or another `onSendVoice` fires), the tray has already been cleared so there's no double-send risk — BUT the `sendMessage(WKImageContent(p))` calls are fire-and-forget with no error handling. If any image path is invalid (deleted file, permission revoked), the send silently fails with no user feedback.

**Code snippet:**
```kotlin
override fun onSendVoice(audioPath: String, seconds: Int, waveform: String) {
    val voiceContent = WKVoiceContent(audioPath, seconds)
    voiceContent.waveform = waveform
    iConversationContext.sendMessage(voiceContent)
    if (!richTextTray.isEmpty()) {
        val snapshot = richTextTray.orderedPaths()
        richTextTray.clear()
        refreshRichTextTray()
        for (p in snapshot) {
            iConversationContext.sendMessage(WKImageContent(p))  // no error handling
        }
    }
}
```

**Data-flow trace:**
- `richTextTray.orderedPaths()` returns `List<String>` of local file paths
- Paths are added to tray via `addImagesToRichTextTray()` → `pickMorePendingImages()` from gallery picker
- Between picking and sending, user could delete the photo from gallery, revoke storage permission, or the file could be on removable storage that's unmounted
- `WKImageContent(p)` constructor doesn't validate the path exists
- `sendMessage()` failure callback is not wired here

**Severity rationale (R1):** If a pending image file is deleted between selection and voice-send, the user gets a voice message but the image silently disappears. User-visible data loss (the image they intended to send is gone with no error toast). This is a new path introduced in this PR.

---

### F4 — P2: CreateThreadActivity — `JSONObject.parseObject` is FastJSON, silently drops unknown fields

**File:** `wkuikit/src/main/java/com/chat/uikit/thread/CreateThreadActivity.java`, line 142  
**Diff-scope:** NEW (introduced in this PR)

**Issue:** The code uses `com.alibaba.fastjson.JSONObject.parseObject(sourceContentRaw)` to parse the raw message content. FastJSON's `parseObject` is lenient and silently drops fields it doesn't recognize. If the source message has custom fields (e.g., bot-specific payload, custom mention entities), they'll be lost in the thread source payload.

**Code snippet:**
```java
sourcePayload = JSONObject.parseObject(sourceContentRaw);
```

**Impact:** Minor — the thread source preview may lose some metadata. The fallback path (plain text) still works. But if the intent is to "透传原 payload" (pass through the original payload), FastJSON's lenient parsing may silently alter it.

---

### F5 — P2: WKTextProvider.relocateMsgTimeView — RV reuse edge case when `textContentLayout` is not a ViewGroup

**File:** `wkuikit/src/main/java/com/chat/uikit/chat/provider/WKTextProvider.kt`, lines 418-420  
**Diff-scope:** NEW (introduced in this PR)

**Issue:** The function does `contentTvLayout.findViewById<View>(R.id.textContentLayout) as? ViewGroup` — if `textContentLayout` exists but is not a `ViewGroup` (e.g., layout XML change), the `as? ViewGroup` returns null and the function returns early without relocating. This is safe (no crash), but the time view would remain in its previous position from a prior bind, causing visual misalignment.

**Code snippet:**
```kotlin
val textContentLayout = contentTvLayout.findViewById<View>(R.id.textContentLayout) as? ViewGroup
    ?: return
```

**Impact:** Low — only triggers if layout XML is inconsistent. Defensive coding is correct.

---

### F6 — P2: ChatPanelManager emoji recent-sort — `readRecentEmojiTexts` splits on comma, emoji text may contain comma

**File:** `wkuikit/src/main/java/com/chat/uikit/chat/ChatPanelManager.kt`, lines 2381-2384  
**Diff-scope:** NEW (introduced in this PR)

**Issue:** `readRecentEmojiTexts()` splits the stored preference string by comma: `raw.split(",")`. If any emoji entry's `text` field contains a comma (unlikely for standard emoji but possible for custom emoji/stickers), the split would corrupt the entry.

**Code snippet:**
```kotlin
private fun readRecentEmojiTexts(): List<String> {
    val raw = WKSharedPreferencesUtil.getInstance().getSPWithUID("common_used_emojis") ?: ""
    if (raw.isEmpty()) return emptyList()
    return raw.split(",").filter { it.isNotEmpty() }
}
```

**Impact:** Low — standard emoji don't contain commas. Custom emoji entries with commas would be split incorrectly, causing them to not match in `buildOrderedEmojiList` and thus not appear in the "recent" section. No crash, just a minor UX glitch.

---

## 3. Suggestions

1. **F1 (MentionEntityHelper):** For RichText messages, consider using the caption-only text (without `[图片]` placeholders) as the `plain` for entity offset validation. The `WKRichTextContent` model should expose a `captionContent` or similar field that matches the sender's reference frame.

2. **F2 (WKThreadCreatedContent):** Add a size cap (e.g., 200 entries) with LRU eviction to `sourceMessageThreadMap`. Alternatively, clear the maps in `clearCache()` or on logout.

3. **F3 (Voice+image send):** Add a file-existence check before each `WKImageContent(p)` send, and show a toast if any image file is missing. Or wire up the send callback to report failures.

4. **F4 (CreateThreadActivity):** Consider using `org.json.JSONObject` instead of FastJSON for stricter parsing, or explicitly copy all fields from the parsed object.

---

## 4. Additional Findings

- **WKRichTextTrayAdapter `richTextTraySendingGate()` always returns false** (line 213): The adapter's internal gate is a no-op stub. The real gating is done by the caller's `onRemove` / `onAddTapped` lambdas. This is documented and intentional, but the dead code is confusing.

- **ChatActivity deletion of `richTextCaptionLauncher`** is clean — all references removed, no dangling imports.

---

## 5. Data-Flow Trace Summary

| Consumed Data | Source | Validated? |
|---|---|---|
| `model.displayContent` in MentionEntityHelper | `WKMessageContent.getDisplayContent()` | ⚠️ May include placeholders that shift offsets (F1) |
| `mInfo.uids` in MentionEntityHelper | `WKMsg.baseContentMsgModel.mentionInfo.uids` from SDK parse | ✅ Populated by SDK from wire `mention.uids` |
| `sourceContentRaw` in CreateThreadActivity | `msg.content` via Intent extra from WKUIKitApplication | ✅ Direct wire content string |
| `richTextTray.orderedPaths()` in voice send | `WKRichTextComposeModel.items[].localPath` from gallery picker | ⚠️ File may be deleted between pick and send (F3) |
| `sourceMessageThreadMap` entries | `decodeMsg()` on each threadCreated system message | ⚠️ Never evicted (F2) |

---

## 6. Blindspot Checklist (R5)

- **C1 (Dual-path parity):** N/A — no symmetric add/remove paths in this PR's scope.
- **C2 (Control-flow ordering):** N/A — no nested reuse of emit/sanitize functions.
- **C3 (Authorization boundary):** N/A — no new endpoints or tool exposure.
- **C4 (Authorization lifecycle):** N/A — no auth changes.
- **C5 (Build ≠ runtime):** N/A — no build/packaging changes.
- **C6 (Governance docs):** N/A — no policy/doc changes.

---

## 7. Cross-Round Blocker Recheck (R6)

N/A — first review of this PR.

---

## Verdict

**[Octo-Q] verdict: REQUEST_CHANGES**

Rationale: F1 (P1) — MentionEntityHelper's gap-filler may fail on the exact RichText cross-platform scenario it was designed to fix, due to `displayContent` including image placeholders that shift offsets. F2 (P1) — static maps grow unbounded with no eviction. F3 (P1) — voice+image send path has no error handling for missing files, leading to silent image loss.

All three are new code introduced in this PR and satisfy R1's definition: they can cause user-visible incorrect behavior (missing @ highlights, memory pressure, silent image loss).

yujiawei (changes requested):
# Code Review — PR #67 (octo-android)

Re-review at head `7681b553` against merge-base with `main` (`f91bfb47`). This round focuses on the two newest commits — `3ba7a134` and `7681b553` — which address the prior round's blockers, plus a full re-pass of the touched areas.

## Status of prior blockers — both resolved

- **Mention rebuild dropping valid mentions (prior round, blocking):** `reconstructMentionEntitiesFromPlain` was rewritten from wholesale-replace to a gap-filler (`MentionEntityHelper.kt:129-223`). Valid plain-relative wire entities are now kept; only out-of-bounds / non-`@` entities are discarded and rebuilt; the `claimed` cursor prevents new entities colliding with kept ones. The original "custom group remark + multi-mention drops a valid `@`" regression is fixed. Confirmed.
- **Webhook list stale after first create from empty state (prior round, blocking):** `ChannelWebhookActivity` now gates the `onResume` reload on an explicit `initialLoaded` flag set in the list callback (`ChannelWebhookActivity.java:75, 172, 183`) instead of `!data.isEmpty()`. First entry does not double-load (the flag is set inside the async callback, after `initData()`'s initial load); returning from the create/edit/one-time-URL pages now reloads even when the list started empty. Confirmed.

## Blocking

### P1 — Mention rebuild can link a mention to the wrong user on cross-end richText with images
`wkbase/src/main/java/com/chat/base/msg/MentionEntityHelper.kt:147-152`

The v2 "valid wire entity" test only checks that `offset+length` is in bounds and `plain[offset] == '@'`. It does not verify that the substring at that offset actually belongs to the entity's `uid`. For iOS-sent richText, `getDisplayContent()` returns the placeholder-inclusive `plain` (image blocks serialize to the 4-char token `[图片]`, see `WKRichTextContent.buildPlainFromBlocks` / `getDisplayContent`), while the wire `entities[i].offset` is caption-relative. The two frames differ by the total length of image placeholders before the text — 4 per leading image.

When a later mention's caption-relative offset coincidentally equals an earlier mention's plain offset, that later entity passes the `plain[offset] == '@'` test and is trusted at the wrong location.

Concrete repro (one leading image, two mentions, 2-char first name):
- caption: `@张三 @李四` → wire offsets `张三`=0, `李四`=4
- plain: `[图片]@张三 @李四` → plain offsets `张三`=4, `李四`=8 (shift = +4)
- `李四`'s wire entity (offset 4) → `plain[4] == '@'` (the start of `@张三`) → trusted as valid, marks `uid_李四` covered, and `claimed[4..6]` is set.
- `张三`'s wire entity (offset 0) → `plain[0] == '['` → discarded → goes to rebuild, but `@张三` at plain offset 4 is already `claimed` → not rebuilt.

Result: `@张三` is highlighted and clickable but resolves to `李四`'s uid; the real `@张三` mention is lost. This is a wrong-target mislink (tapping one person's name opens another's profile), not just a missing highlight. It is bounds-checked so there is no crash, and `mentionInfo.uids` is untouched so push/delivery is unaffected — but it is a user-visible correctness defect in the exact subsystem this commit changes. The trigger (iOS richText + a leading image + two mentions whose name lengths align with the placeholder shift, e.g. a 2-char first name) is narrow but realistic for Chinese-language chat.

Suggested fix (small): when classifying a wire entity as valid, also require that `plain.substring(offset + 1, offset + length)` matches one of that uid's name candidates (`collectMentionNameCandidates(e.value, …)`). An entity that lands on a `@` belonging to a different uid then fails validation, is discarded, and is rebuilt at its correct position. Alternatively, detect the global caption→plain shift once and reject per-entity wire offsets when any image block precedes the text.

## Non-blocking (P2 / advisory)

- **Coverage is uid-set based before dedup/uid filtering** — `MentionEntityHelper.kt:155`. `coveredUids`/`claimed` are computed from `validWireMentions` before filtering to `realUids` or resolving duplicate/overlapping offsets. Two valid-looking entities at the same offset mark both uids covered though only one range renders, and an entity whose `value` is not in `realUids` can survive the fast path. Filter `validWireMentions` to real uids and dedupe overlapping offsets before computing coverage. (Same area as the P1; folding both into one pass is natural.)
- **`IncomingWebhookManager.test` try/catch** — `service/IncomingWebhookManager.java:219-245`. Wrapping the synchronous `createService(...).test(...)` call so a synchronous throw resets `hasTestInFlight` and reports back is a correct, defensive addition. Note `hasTestInFlight` is still a plain (non-`volatile`) field mutated from callbacks; fine while delivery is main-thread, a one-line `volatile` would make the single-in-flight invariant explicit. Nit.
- **`TransferGroupOwnerActivity` fetchChannelInfo** — `TransferGroupOwnerActivity.java:251`. Adding `fetchChannelInfo` after `groupMembersSync` so the parent detail page's `addOnRefreshChannelInfo` listener re-renders role permissions is correct and matches the documented iOS parity. No issue.
- **Thread-name truncation can split a surrogate pair** — `WKUIKitApplication.java:391`. `threadName.substring(0, 11)` cuts by UTF-16 units; if index 11 falls inside an emoji/surrogate pair the result contains a broken char. Use code-point-aware truncation.
- **Voice-to-text reuses the live input field** — `ChatPanelManager.kt:2440`. With a pending image tray, STT text is staged via `editText.setText(text)` then `flushRichTextTraySend()`, restoring the previous text on the fallback path. This fires the input `TextWatcher`s transiently (typing state, send-button, and the `@` mention picker if the transcript contains `@`). Works, but passing the caption directly into `flushRichTextTraySend(caption)` instead of round-tripping the UI field would avoid the side effects. Advisory.
- **Recent-emoji prefs keep unrecognized tokens** — `EmojiFragment.java:204`. A token that resolves to `null` is skipped for the list but still appended to the persisted `canonicalIds`, so dead tokens accumulate in prefs. Move the append inside the `entry != null` branch.
- **Notification-prompt cooldown not monotonic** — `TabActivity.java:250`. `now - lastTs < COOLDOWN_MS` goes negative (and stays "in cooldown") if the system clock moves backward. Guard with `Math.abs(...)` or reset `lastTs` when it exceeds `now`. Minor.
- **Bubble pulse animator on recycled cells** — `BubblePulseHelper.kt`. A pulse in flight when a RecyclerView cell is recycled keeps animating the rebound cell until the next bind resets transforms. Benign in practice (rebind resets), worth a detach-time cancel if cells are ever reused mid-pulse. Nit.

## Verified good

- Gap-filler correctly preserves valid wire mentions for the common "custom remark + multi-mention" case (the prior blocker), keeps non-mention entities (links etc.) untouched, and does not fabricate highlights when a name cannot be resolved.
- `initialLoaded` flag: no double-load on first entry; reload-on-return now works from the empty state; first-load failure still allows manual pull-to-refresh.
- Single call site for the mention helpers (`WKUIChatMsgItemEntity.java:164,168`), wrapped in try/catch so any rebuild exception degrades to plain text rather than breaking the bubble.

## Coverage notes (not verified from the diff)

- The P1 repro is static analysis (frame-alignment + render-path tracing). A device repro with an iOS-sent richText (one image + `@<2-char-name> @<other>`) would confirm the mislink end-to-end.
- Server-side authorization for the transfer / webhook endpoints is out of repo and unverified.
- Binary asset swaps (`action_celebrate.mp4`, `classy_celebrate.mp4`) were not inspected.
- The non-mention P2 items above sit in code introduced earlier in this PR (not in the two re-review commits); they are advisory and do not gate this round.

## Verdict

One blocking item: the wrong-user mention mislink (P1) in the same code path this commit touches, fixable with a few lines of substring/uid validation. Both prior-round blockers are resolved. With the P1 closed this is otherwise ready.

Jerry-Xin (approved):
Re-reviewed at head `2f0d08781c49a90dacd410d60049176fbbcf10e1`. Both remaining P1 blockers from the prior round are correctly fixed, verified against actual bytes. This APPROVE supersedes my earlier CHANGES_REQUESTED (the `contentTv` visibility item — now fixed).

**✅ FIX 1 — Mention rebuild wrong-uid mislink (yujiawei P1): resolved**

`MentionEntityHelper.reconstructMentionEntitiesFromPlain` v3 adds two checks to the "valid wire entity" classifier:
- `(c)` `value` (uid) must be in `realUidSet`;
- `(d)` `plain.substring(offset+1, offset+length)` must NOT match any *other* uid's name candidates:
```kotlin
val needle = plain.substring(e.offset + 1, e.offset + e.length)
for (other in realUids) {
    if (other == uid) continue
    if (needle in candidatesFor(other)) return@filter false
}
```
Walking the documented repro (`@张三 @李四` + 1 leading image → plain `[图片]@张三 @李四`): 李四's wire offset=4 lands on `@张三`'s `@`; v2 trusted it. v3 computes `needle="张三"`, finds it in 张三's candidates → rejects it → 李四 no longer mislinks. 张三's wire (offset 0, `plain[0]='['`) is also rejected; both go to the gap-filler, which rebuilds `@张三`@4 and `@李四`@8 at the correct positions (claim was not pre-occupied by the rejected 李四 entity). Mislink no longer possible.

Correctly scoped: it only checks "collides with *another* uid", deliberately NOT "matches *self*", so the previously-fixed "custom group remark + multi-@" case (receiver doesn't know sender's remark name but wire offset is correct) stays valid. Earlier-good behavior intact: fast-path preserved (now also requires `existingMentions.size == validWireMentions.size`), `claimed` bitmap, length-desc candidate ordering, sentinel-uid skip, cross-platform iOS RichText caption-relative→rebuild, and the no-fabrication guarantee all unchanged. The candidate-name cache (`candidatesByUid`) is a clean perf refactor shared across validation + rebuild — no behavioral change.

**✅ FIX 2 — Table-message RV-reuse visibility残留 (my prior CR): resolved**

`WKTextProvider` +7 lines reset `contentTv.visibility = View.VISIBLE` before `setText` in both reuse-vulnerable branches:
- no-table branch (line ~282, before `OrderedListItemSpan.measure`/`setText`);
- placeholder-mismatch fallback (line ~325, before `contentTv.text = displaySpans`).

The INTERLEAVED branch's `contentTv.visibility = View.GONE` (blank first segment, ~line 357) is the only GONE setter; both early-return text paths now restore VISIBLE, so a holder previously bound to a table-starting message renders normal text correctly on reuse. No new regression — VISIBLE is only forced on the text-display paths; the GONE-on-blank-first-segment logic is untouched.

**Prior fixes still intact (grep-confirmed, builds on 7681b55):** webhook `initialLoaded` refresh gate (`ChannelWebhookActivity.java:75/172/183`), transfer-owner `fetchChannelInfo` (`WKChatBaseProvider.kt:1000` + transfer path), webhook test `hasTestInFlight` try/catch reset (`IncomingWebhookManager.java:50/192/214`). None reverted.

No blocking items. LGTM.

mochashanyao (approved):
[Octo-Q · automated review]

---

# octo-android PR #67 Review Report
**Head SHA:** 2f0d08781c49a90dacd410d60049176fbbcf10e1
**Scope:** 82 files changed, +5260 / -1015 lines

---

## 1. Verification Conclusions

| Area | Status | Evidence |
|------|--------|----------|
| SpaceGuideActivity logout button | ✅ | Correct fix for #66 — new users stuck on invite code page can now log out |
| EmojiFragment ClassCastException | ✅ | Adapter data type changed from String to EmojiEntry; refreshRecentEmojiHeader() correctly handles new type |
| WKMarkwonProvider table parsing | ✅ | Inserts blank line before table header when previous line is non-empty non-pipe — fixes commonmark GFM table recognition |
| MentionEntityHelper cross-platform reconstruction | ✅ | 200+ line function with proper null guards, collision detection, sentinel UID handling; single call site in WKUIChatMsgItemEntity constructor |
| MessageForwardSupport whitelist | ✅ | Forwardable content type whitelist prevents forwarding unsupported types |
| BubblePulseHelper scale animation | ✅ | Clean scale pulse animation for message bubbles |
| ThumbsUpEffect rewrite | ✅ | Complete rewrite rendering 👍 emoji characters with bubble hit detection |
| ChatPanelManager emoji/tray/voice | ✅ | Emoji panel sort, inline image tray, sendVoiceTextDirect, pickMorePendingImages all follow existing patterns |
| WKRichTextCaptionActivity deletion | ✅ | 412-line activity correctly replaced by inline tray editing |
| TabActivity notification cooldown | ✅ | 7-day cooldown with proper reset on grant; cannot permanently suppress prompt |
| WKTextProvider RV reuse fixes | ✅ | contentTv.visibility restoration, relocateMsgTimeView with null guards in ChatTextTimeLayout |
| TransferGroupOwner feature | ✅ | Single-select member picker with search, pagination, proper admin-only visibility |
| Incoming Webhook module | ✅ | 6 REST endpoints, test cooldown state machine, one-time URL display |
| Summary module StateFlow/SharedFlow | ✅ | SharedFlow with replay=0 but lifecycleScope.launch keeps collector alive during STOPPED; StateFlow for pendingScrollToTop survives configuration changes |
| Thread source message tracking | ✅ | ConcurrentHashMap-based tracking with proper cleanup on thread close |
| WKThreadCreatedProvider error handling | ✅ | Status==3 or ANY error → mark closed + toast (replaces string-matching heuristic) |
| CreateThreadActivity source payload | ✅ | Prefers raw JSON parse with contentType override, falls back to plain text |
| SummaryRelatedChatSheet DM channel split | ✅ | Correctly splits composite "<myUid>@<peerUid>" channel IDs for PERSONAL type only |

## 2. Findings

### P2-1: WKThreadCreatedContent.sourceMessageThreadMap unbounded growth

**File:** `wkim/src/main/java/com/xinbida/wukongim/msgmodel/WKThreadCreatedContent.java`

The global `sourceMessageThreadMap` (ConcurrentHashMap) is populated in `decodeMsg` for every thread-created message received. Cleanup only occurs via `markThreadClosedForSourceMessageId`, which is called when:
- User explicitly deletes a thread (ThreadDetailActivity/ThreadListActivity)
- User taps a closed/error thread card (WKThreadCreatedProvider)

**Concern:** In long-running sessions with many active threads, this map grows without bound. Each entry holds a reference to the full `WKThreadCreatedContent` object. While these objects are also held by the IM SDK's message store (so no additional memory allocation), the map itself adds overhead proportional to total threads seen.

**Diff-scope:** New (introduced by this PR).
**Severity rationale:** P2 — resource concern, not user-visible error. Mitigated by the fact that entries are just references to already-cached objects, and typical session lengths won't cause meaningful pressure. A periodic cleanup (e.g., evict entries older than N hours or cap at N entries) would be a nice-to-have.

### P2-2: WKThreadCreatedProvider marks thread closed on ANY API error

**File:** `wkuikit/src/main/java/com/chat/uikit/conversation/provider/WKThreadCreatedProvider.kt`

On card tap, if the thread API returns any error (not just 404), the thread is marked closed and removed from the "enter thread" menu. This is aggressive for transient errors (500, timeout, network unavailable) — the thread likely still exists server-side.

**Diff-scope:** New behavior (replaces previous string-matching heuristic that missed 404s).
**Severity rationale:** P2 — the previous heuristic was worse (missed real 404s), and the new behavior provides user feedback via toast. Users can pull-to-refresh to restore access after transient errors. Consider differentiating: 404 → permanent close; other errors → toast "unable to enter thread, try again" without marking closed.

### P2-3: SummaryEvents.listShouldRefresh scrollToTop=true path never exercised

**File:** `wkbase/src/main/java/com/chat/base/summary/SummaryEvents.kt`

The SharedFlow is defined with `replay=0, extraBufferCapacity=1, DROP_OLDEST`. The consumer in SmartSummaryListActivity handles both `scrollToTop=true` and `scrollToTop=false`. However, all three current emit sites in SmartSummaryDetailViewModel pass `scrollToTop=false`. The `scrollToTop=true` path is documented as "reserved for future use" but is currently dead code.

**Diff-scope:** New (introduced by this PR).
**Severity rationale:** P2/nit — dead code path with clear documentation. Not a bug, but worth tracking for future implementation.

### P2-4: MentionEntityHelper offset mismatch on edited messages

**File:** `wkbase/src/main/java/com/chat/base/msg/MentionEntityHelper.kt`

`reconstructMentionEntitiesFromPlain` computes mention offsets against `model.displayContent` (original content). The downstream `formatSpans` in WKUIChatMsgItemEntity renders against `getContent()`, which returns edited content when `contentEditMsgModel` is present. If a message with mentions is edited, reconstructed entity offsets may not align with the edited text, causing mention highlights to silently disappear (downstream bounds check at line 387 prevents crash via `continue`).

**Diff-scope:** Pre-existing design gap — the legacy `mentionInfo.uids` path (line 337-375) has the same issue since it also uses `indexOf` on `displaySpans` built from edited content. This PR does not introduce or amplify it.
**Severity rationale:** P2 — pre-existing, silent degradation (no crash), edge case (edited messages with mentions).

### P2-5: SmartSummaryCreateActivity DEBUG-only logging in production code

**File:** `wkuikit/src/main/java/com/chat/uikit/summary/SmartSummaryCreateActivity.kt`

The `channelToSource` function contains `if (BuildConfig.DEBUG) Log.d(...)` logging. While correctly gated behind BuildConfig.DEBUG, the function itself and its OctoSourceType mapping (1=group, 2=thread, 3=DM) are critical for server-side attribution. The mapping logic is correct but would benefit from a unit test to prevent regression.

**Diff-scope:** New (introduced by this PR).
**Severity rationale:** P2/nit — correct implementation, but critical mapping with no test coverage.

## 3. Suggestions

1. **sourceMessageThreadMap cleanup:** Consider adding a periodic eviction (e.g., LRU cap at 500 entries, or time-based expiry) to prevent unbounded growth in long-running sessions.

2. **Thread error differentiation:** In WKThreadCreatedProvider, differentiate between 404 (permanent close) and transient errors (toast + retry). This preserves the fix for the 404 blind spot while avoiding false closures.

3. **channelToSource unit test:** Add a simple unit test for the OctoSourceType mapping to prevent regression — this is a critical server-side attribution path.

4. **Webhook test cooldown:** The `hasTestInFlight` global flag in IncomingWebhookManager is correctly protected by try-catch around synchronous exceptions. Consider adding a timeout-based reset (e.g., 30s) as a belt-and-suspenders measure against edge cases where the network callback is never invoked.

## 4. Extra Findings

### Notification permission cooldown is well-designed
The 7-day cooldown in TabActivity correctly handles: default value "0" ensures first-time display; grant resets to "0"; corrupted SharedPreferences values fall back to 0L; clock rollback triggers re-prompt. No permanent suppression path exists.

### SummaryEvents SharedFlow design is correct
Despite `replay=0`, events are NOT lost while the activity is STOPPED. The collector uses `lifecycleScope.launch` (alive during CREATED-to-DESTROYED), not `repeatOnLifecycle(STARTED)`. Active collectors receive `tryEmit` immediately regardless of lifecycle state. The `uiState` collect correctly uses `repeatOnLifecycle(STARTED)` since it touches UI widgets.

### RV reuse fixes are thorough
WKTextProvider correctly restores `contentTv.visibility=VISIBLE` when the previous bind was a table message. `relocateMsgTimeView` safely swaps the msgTimeView parent between BubbleLayout and textContentLayout. ChatTextTimeLayout adds null guards for both `onMeasure` and `onLayout` to handle the relocated state.

### Webhook module is well-structured
The 6-endpoint IncomingWebhookManager has proper error handling, the test cooldown state machine correctly resets `hasTestInFlight` on all paths (success/fail/synchronous exception), and the one-time URL display with shell-safe quoting is a nice security touch.

### TransferGroupOwner correctly hides ineligible members
The member picker hides the current user and robots, showing only eligible transfer targets. Pagination (50/page) with search is appropriate for large groups.

## 5. Data Flow Tracing

### MentionEntityHelper.reconstructMentionEntitiesFromPlain
- **Input:** `msg.baseContentMsgModel.mentionInfo.uids` ← populated from wire JSON `mention.uids` array by MsgManager.java (line 208-216)
- **Input:** `model.displayContent` ← for WKTextContent: `content` field; for WKRichTextContent: `plain` field (text blocks + `[图片]` placeholders)
- **Input:** `model.entities` ← populated by SDK parsing + mergeMentionEntities (runs immediately before this function)
- **Output:** Mutates `model.entities` with validated wire entities + reconstructed entities for uncovered UIDs
- **Consumer:** `formatSpans` in WKUIChatMsgItemEntity (line 385) reads entities and applies spans to `displaySpans` built from `getContent()`
- **Gap:** Edited messages (contentEditMsgModel present) cause offset mismatch — pre-existing, silent degradation

### SummaryEvents.listShouldRefresh
- **Producer:** SmartSummaryDetailViewModel posts events via `postListShouldRefresh(scrollToTop)` on cancel/regenerate/delete
- **Channel:** MutableSharedFlow(replay=0, extraBufferCapacity=1, DROP_OLDEST)
- **Consumer:** SmartSummaryListActivity collects in `lifecycleScope.launch` (alive during STOPPED) → calls viewModel.reload() or reloadAndScrollToTop()
- **Data flow verified:** Events reach consumer during STOPPED state because collector coroutine is not cancelled until DESTROYED

### WKThreadCreatedContent.sourceMessageThreadMap
- **Producer:** `decodeMsg` populates map from `source_message_id` JSON field (handles both long and string types)
- **Consumer:** WKUIKitApplication reads map for long-press "enter thread" menu; WKThreadCreatedProvider reads for card tap
- **Cleanup:** `markThreadClosedForSourceMessageId` removes from both sourceMessageThreadMap and sourceMessageIdSet
- **Gap:** No periodic cleanup — map grows with every thread-created message in the session

### IncomingWebhookManager test cooldown
- **State:** `hasTestInFlight` (global boolean) + `cooldownExpire` (HashMap<String,Long> per-webhook)
- **Flow:** test() checks hasTestInFlight → checks per-webhook cooldown → fires API → resets hasTestInFlight in callback (success/fail) → sets cooldownExpire to now+3s
- **Protection:** try-catch around entire method body prevents hasTestInFlight=true deadlock on synchronous exceptions
- **Verified:** All exit paths reset hasTestInFlight

### SmartSummaryCreateActivity originChannelType
- **Producer:** ChatActivity.newIntentForChat maps WKChannelType → OctoSourceType (GROUP→1, COMMUNITY→2, PERSONAL→3)
- **Consumer:** SmartSummaryCreateActivity reads `originChannelType` extra → passes to server via createSummary API
- **Verified:** Mapping is correct and consistent with server expectations

## 6. Blindspot Checklist (R5)

- **C1 Dual-path parity:** N/A — no symmetric add/remove or create/delete paths in this PR's changes.
- **C2 Control-flow ordering / nested reuse:** Clear — MentionEntityHelper runs after mergeMentionEntities in a defined sequence; no double-application risk. Webhook test cooldown uses global flag + per-webhook cooldown without nesting issues.
- **C3 Authorization boundary ≠ capability boundary:** N/A — no new tool/endpoint exposure beyond the webhook module's own API endpoints (which are server-side authorized).
- **C4 Authorization lifecycle / container-member state cascade:** N/A — no auth changes in this PR.
- **C5 Build/note ≠ runtime path:** N/A — no build artifacts, extensions, or CLI packaging changes.
- **C6 Governance/policy document consistency:** N/A — no governance or policy documents changed.

## 7. Cross-round Blocker Recheck (R6)

N/A — first review of this PR.

---

## Summary

This is a substantial PR (82 files, +5260/-1015) covering bug fixes, new features (incoming webhooks, group ownership transfer, thread source tracking), and UX improvements (notification permission cooldown, summary toast actions). The code quality is consistently high across all modules. Data flow tracing confirmed correct runtime behavior for all critical paths. No P0 or P1 issues found.

**[Octo-Q] verdict: APPROVE** — No blocking issues. Five P2 findings noted for future improvement (thread map cleanup, error differentiation, dead code path, edited-message offset gap, missing unit test for critical mapping).

yujiawei (approved):
# Code Review — PR #67 (octo-android)

Reviewed at head `2f0d0878`, against `main` (merge-base `f91bfb47`). 82 files, +5260/-1015, 17 commits across chat / thread / forward / notification / login / group-management / smart-summary.

## Verdict: Approve

No correctness, security, data-loss, or build-breaking issues found. The change set is large but well-factored: each module is internally consistent, dead code from the removed full-screen caption flow (`WKRichTextCaptionActivity`, `richTextCaptionLauncher`, the `pendingCaption*` fields) is fully cleaned up with no dangling references, and the new activities are all registered in the manifest. The findings below are all non-blocking polish items.

## Verification of the headline fixes

- **#65 (thread source content downgraded to `[聊天记录]`)** — `WKUIKitApplication` now passes `sourceContentRaw` (`msg.content`) + `sourceContentType`, and `CreateThreadActivity` parses the raw payload and overrides `type`, falling back to the old plain-text path only when raw is empty. Rich types (merge-forward / image / file) are preserved. ✅
- **#66 (logout entry on invite-code page)** — covered by `028914c`. ✅
- **Cross-end @ highlight** — `MentionEntityHelper.reconstructMentionEntitiesFromPlain` only discards wire entities that are provably wrong (out-of-range, not pointing at `@`, or whose substring collides with *another* uid's candidate name) and rebuilds the rest via a claim-cursor gap-filler. The "don't require self-name match" rule correctly preserves the custom-group-remark case. Authoritative source is `mentionInfo.uids`; no fabricated highlights. ✅
- **Thread close/delete state sync** — `markThreadClosedForSourceMessageId` is wired into all three close paths (card-tap status==3, detail-page close, list-page delete), and the card-tap error branch now fails closed (any error → treat as closed) instead of fragile string matching. ✅
- **Notification prompt cooldown** — 7-day cooldown with reset-on-grant, and `openChannelSetting` correctly switched to `ACTION_APP_NOTIFICATION_SETTINGS` with a `gotoSet` fallback. ✅
- **Markdown table boundary** — `ensureTableTermination` now inserts a blank line on *both* sides of a table; `isTableDelimiterLine` handles optional leading/trailing `|`. ✅
- **CounterView vertical centering** — `getTextTopInPill()` derives the baseline from paint metrics instead of the hardcoded `dp(4)`; consistently applied to all draw branches. ✅
- **Incoming webhook module** — model null-safety, `absoluteURL` `/v1` de-duplication, and the test-send state machine (single in-flight + 3s cooldown, with a `try/catch` reset guarding synchronous throws) are all sound. ✅

## Non-blocking findings (P2)

1. **Long-press "enter thread" skips the status check** — `WKUIKitApplication.java:399-413`. The card-tap path (`WKThreadCreatedProvider`) calls `getThreadDetail` and handles `status==3` / errors before navigating, but the long-press menu's "enter thread" action `startActivity`s straight to the cached `channel_id`. If the source mapping was populated from an old `threadCreated` message and the thread was closed/deleted from another device (or before a process restart re-decodes the close), this opens a dead `COMMUNITY_TOPIC`. Consider routing this path through the same `getThreadDetail` status check so it can fall back to "Create Thread" on a closed thread.

2. **Owner-search may apply a stale callback** — `TransferGroupOwnerActivity.java:184-187`. `getWithPageOrSearch` can fire its callback more than once for external groups (local DB then remote). `resortData` reads the *current* mutable `page`/`searchKey`, so if the user edits the search box or pages before a slower remote result returns, the old list is applied/appended under the new key. Capturing the requested `(searchKey, page)` in the lambda and dropping non-matching callbacks would make this robust. (Same pattern exists in sibling member-list screens, so this is pre-existing behavior, not a regression introduced here.)

3. **Webhook name cannot be cleared on edit** — `ChannelWebhookEditActivity.java:160`. `hasChange()` only treats the name as changed when `!TextUtils.isEmpty(n)`, so blanking an existing name to let the server re-derive a default is a no-op. Renaming to any non-empty value works; only the clear-to-empty case is affected.

4. **Pull-to-refresh spinner is time-boxed, not callback-driven** — `ChannelWebhookActivity.java:128-130`. `finishRefresh(800)` runs immediately after `loadList(false)` rather than from the load callback, so on a slow/failed network the spinner ends before data arrives. Cosmetic.

5. **Usage meta line can show an empty timestamp** — `WebhookListAdapter.java:142`. When `callCount > 0` but `lastUsedAt <= 0` (malformed/partial API response), `str_webhook_meta_usage` is formatted with an empty "last used" value. Minor.

## Nits

- `ChatActivity.parseMentionEntitiesFromJson` is now orphaned after the caption-launcher removal — safe to delete.
- `BubblePulseHelper` holds a view reference for up to 350ms if the activity is destroyed mid-pulse; negligible, and the tag is nulled on end/cancel.

## Coverage note

Reviewed the full textual diff plus the surrounding implementations of the touched classes in the checked-out tree. Not assessed: binary asset swaps (`action_celebrate.mp4`, `classy_celebrate.mp4`), the actual on-device R8/`assembleDevRelease` build (author reports it passes locally with minify + shrinkResources), server-side webhook/transfer endpoint contracts, and real-device behavior of the items in the PR's unchecked Test Plan (notification cooldown on Huawei/HarmonyOS, inline image tray, group management flows).

PR #73, opened 2026-06-27
ci(octo-issue-feed): opt into webhook-only triage

lml2468:
Adds enable_im_notify: false to opt this repo out of the redundant IM ping (multica triage agent creates a tracked issue). Backed by Mininglamp-OSS/.github#80.

### DIFF
--- a/.github/workflows/octo-issue-feed.yml
+++ b/.github/workflows/octo-issue-feed.yml
@@ -16,6 +16,7 @@ jobs:
       issue_url: ${{ github.event.issue.html_url }}
       issue_author: ${{ github.event.issue.user.login }}
       event_action: ${{ github.event.action }}
+      enable_im_notify: false
     secrets:
       OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}
       TRIAGE_WEBHOOK_URL: ${{ secrets.TRIAGE_WEBHOOK_URL }}


mochashanyao (changes requested):
[Octo-Q · automated review]

---

# octo-android PR#73 Review Report
**Head SHA:** `501833925d258758a39cdc8d23eda49aa74dcc28`
**PR State:** MERGED
**File changed:** `.github/workflows/octo-issue-feed.yml` (+2 / −1)

---

## 1. Verification Summary

| Item | Status | Evidence |
|------|--------|----------|
| `enable_im_notify: false` input passed to reusable workflow | ⚠️ Silently ignored | Reusable workflow `octo-issue-notify.yml@v1` does NOT declare this input |
| `v1` tag includes `enable_im_notify` support | ❌ | `v1` → commit `135e280f`, which is **3 commits behind** `.github#80`'s merge commit `8ade879f` |
| Trailing newline removal | ✅ Cosmetic only | No functional impact |

## 2. Findings

### P1 — `enable_im_notify: false` is inert; PR goal not achieved

**What:** The PR adds `enable_im_notify: false` as an input to the reusable workflow call `octo-issue-notify.yml@v1`. However, the reusable workflow at the `v1` tag (commit `135e280f`) does **not** declare `enable_im_notify` in its `workflow_call.inputs` section. GitHub Actions silently ignores undeclared inputs passed via `with:` to reusable workflows.

**Data-flow trace:**
- Caller (`.github/workflows/octo-issue-feed.yml` line 22): passes `enable_im_notify: false`
- Reusable workflow (`octo-issue-notify.yml@v1`, inputs section lines 26–46): declares only `repo_name`, `issue_number`, `issue_title`, `issue_url`, `issue_author`, `event_action`, `api_base_url`, `feed_group_id` — no `enable_im_notify`
- The `Send` step (line 92) always receives `im-message: ${{ steps.build.outputs.im_message }}` which is always populated when `event_action == 'opened'` — no conditional on `enable_im_notify`
- **Result:** IM notification fires for every `opened` issue, identical to pre-PR behavior

**Root cause:** `.github#80` (which adds the `enable_im_notify` input to the reusable workflow) was merged to main at 2026-06-27T07:23:41Z, but the `v1` tag was **never bumped** from `135e280f` to include the new commit (`8ade879f`). The `v1` tag is 3 commits behind the merge commit of `.github#80`.

**Diff-scope:**
1. **New** — the `enable_im_notify: false` line is added by this PR
2. The line is newly introduced dead configuration — it looks functional but is silently ignored
3. Does NOT match base-branch behavior — base branch has no such line

**Impact:**
- The stated PR goal ("opt this repo out of the redundant IM ping") is **not achieved**
- IM notifications continue to fire alongside the triage webhook, producing the exact "redundant IM ping" the PR intended to eliminate
- Future readers of the workflow file will incorrectly assume IM is disabled, making debugging harder
- No runtime breakage — behavior is identical to before, but the intended improvement is absent

**Fix:** Bump the `v1` tag in `Mininglamp-OSS/.github` to the merge commit of `.github#80` (`8ade879f`) or later (e.g., current `v1.1.2` at `2f6346c5`, if it includes the change — verified: v1.1.2 also does NOT include `enable_im_notify`, so the tag needs to be moved to or past `8ade879f`).

### P2 — Missing trailing newline

**What:** The diff removes the trailing newline from the file (`\ No newline at end of file`).

**Impact:** Cosmetic. POSIX convention prefers a trailing newline. Some editors will auto-add one on save, causing noise in future diffs. Not blocking.

## 3. Suggestions

1. **Immediate:** Move the `v1` tag in `Mininglamp-OSS/.github` to the commit that includes `.github#80`'s changes. Verify with `gh api "repos/Mininglamp-OSS/.github/contents/.github/workflows/octo-issue-notify.yml?ref=v1"` that `enable_im_notify` appears in the inputs section.
2. **Consider:** Bumping to `@v1.1.2` or a newer tag that includes the feature, rather than relying on a floating `v1` tag.
3. **Add trailing newline** back to the file to follow POSIX convention.

## 4. Additional Observations

- The `.github#80` test plan explicitly lists "In a follow-up, opt one caller" as a step — this PR is that follow-up. However, the dependency chain (v1 tag must be bumped first) was not verified before merging the caller change.
- The PR is already **MERGED** — the P1 finding needs immediate remediation via a v1 tag bump, otherwise every new issue in `octo-android` will continue to produce redundant IM pings.

## 5. Data-Flow Trace

| Consumed Data | Source | Runtime Value at Consumer |
|---|---|---|
| `enable_im_notify: false` (caller input) | `.github/workflows/octo-issue-feed.yml:22` | **Silently dropped** — not declared in reusable workflow's `inputs:` |
| `im-message` (Send step) | `steps.build.outputs.im_message` | Always populated when `event_action == 'opened'` — no gate on `enable_im_notify` |
| `webhook-payload` (Send step) | `steps.build.outputs.webhook_payload` | Correctly built and passed via `TRIAGE_WEBHOOK_URL` secret |
| `event_action` filter | `github.event.action` | Only `opened` proceeds; `reopened` skipped — working as designed |

## 6. Blind-Spot Checklist (R5)

- **C1** Dual-path parity: N/A — no symmetric add/remove paths touched.
- **C2** Control-flow ordering / nested reuse: N/A — single workflow call, no nested composite action ordering issue.
- **C3** Authorization boundary: N/A — no permission/capability changes.
- **C4** Authorization lifecycle: N/A — no auth changes.
- **C5** Build ≠ runtime: **HIT** — analogous to C5 pattern. The PR's YAML change looks correct statically, but at GitHub Actions runtime the reusable workflow at `@v1` doesn't recognize the input. The "build" (YAML syntax) passes, but the runtime behavior is unchanged. Same class as C5 (static-correct, runtime-incorrect).
- **C6** Governance/policy docs: N/A — no policy document changes.

## 7. Cross-Round Blocker Recheck (R6)

N/A — first review.

---

**[Octo-Q] verdict: REQUEST_CHANGES**

P1: `enable_im_notify: false` is silently ignored because the `v1` tag in `Mininglamp-OSS/.github` was not bumped after `.github#80` merged. The PR's stated goal (disable redundant IM notifications) is not achieved. Fix: bump `v1` tag to include `.github#80`'s changes, then re-verify.

yujiawei (changes requested):
# Code Review — PR #73 (octo-android)

## Summary

This PR adds `enable_im_notify: false` to the `octo-issue-feed.yml` caller so the repo opts out of the redundant IM ping and relies on webhook-only triage. The intent is sound, but the change is **built against a reusable-workflow input that does not exist on the ref this caller pins to**, which will break the workflow at startup.

## Spec compliance

The diff matches the stated 1-line intent (add the input, no over/under-build). The problem is not *what* was added but *that the dependency it relies on is not available on the consumed ref* — covered below.

## Findings

### P1 (blocker) — `enable_im_notify` is not a declared input on `octo-issue-notify.yml@v1`

`.github/workflows/octo-issue-feed.yml` calls:

```yaml
jobs:
  notify:
    uses: Mininglamp-OSS/.github/.github/workflows/octo-issue-notify.yml@v1
    with:
      ...
      enable_im_notify: false   # <- added by this PR
```

The caller pins `@v1`. I checked the reusable workflow's `on.workflow_call.inputs` across every released ref:

| ref | `enable_im_notify` declared? |
|-----|------------------------------|
| `v1` | ❌ no |
| `v1.0.0` | ❌ no |
| `v1.1.0` | ❌ no |
| `v1.1.1` | ❌ no |
| `v1.1.2` | ❌ no |
| `main` | ✅ yes |

The input was added by `Mininglamp-OSS/.github#80` (*"feat(octo-issue-notify): add enable_im_notify input"*, merged to **main** at 07:23:41Z). However, the `v1` tag still points to commit `135e280` (dated 2026-06-09), which is **3 commits behind** the merge of #80 — the tag was never moved, and no new `v1.x` release tag carrying the input was cut.

**Consequence:** GitHub Actions validates `with:` inputs against the *called* workflow's declared inputs. Passing an input the referenced workflow does not define produces a startup failure:

> `Invalid input, enable_im_notify is not defined in the referenced workflow`

This fails the entire `notify` job before any step runs. Critically, IM notify **and** the triage webhook forwarding live in that same reusable job — so this change does not merely "skip the IM ping," it takes down the whole notify+triage path. The PR's own goal (webhook-only triage) is defeated: the webhook is forwarded from inside the job that now errors out.

**Fix options:**
1. Cut a release tag of `.github` that includes #80 (e.g. `v1.2.0`, or move the `v1` tag forward) **before** this caller pins to it, then bump the `uses:` ref here to that tag.
2. Or temporarily pin this caller to a commit SHA / `main` that contains #80 (less ideal for supply-chain hygiene).

Either way, the `uses:` ref must point at a revision where the input is declared. As written against `@v1`, this PR breaks `Octo Issue Feed`.

### P2 (nit) — missing trailing newline

The last line `TRIAGE_WEBHOOK_URL: ${{ secrets.TRIAGE_WEBHOOK_URL }}` no longer ends with a newline (`\ No newline at end of file`). Minor POSIX hygiene; worth restoring.

## Verdict

The opt-out approach is correct, but it is wired to an input that is not present on the pinned `@v1` reusable workflow, which will fail the workflow run rather than silence the IM ping. Requesting changes: align the `uses:` ref with a revision that declares `enable_im_notify` (i.e. release/advance `.github` first), then re-pin here.

PR #74, opened 2026-06-27
ci(octo-pr-review-feed): opt into webhook-only triage

lml2468:
Adds enable_im_notify: false + wires PR_TRIAGE_WEBHOOK_URL secret. Backed by Mininglamp-OSS/.github#81 (which added the input + secret to the reusable). PR fan-out triage agent now files a tracked review issue; the IM ping becomes redundant.

### DIFF
--- a/.github/workflows/octo-pr-review-feed.yml
+++ b/.github/workflows/octo-pr-review-feed.yml
@@ -17,6 +17,8 @@ jobs:
       pr_url: ${{ github.event.pull_request.html_url }}
       pr_author: ${{ github.event.pull_request.user.login }}
       event_action: ${{ github.event.action }}
+      enable_im_notify: false
       project_group_id: '84667e59ee9544e6b31a2fb6d3a21b23'
     secrets:
       OCTO_BOT_TOKEN: ${{ secrets.OCTO_BOT_TOKEN }}
+      PR_TRIAGE_WEBHOOK_URL: ${{ secrets.PR_TRIAGE_WEBHOOK_URL }}


yujiawei (approved):
# Code Review — PR #74 (octo-android)

## Summary
Two-line change to `.github/workflows/octo-pr-review-feed.yml`: passes `enable_im_notify: false` and wires the `PR_TRIAGE_WEBHOOK_URL` secret into the reusable `Mininglamp-OSS/.github/.github/workflows/octo-pr-review-feed.yml@v1`. The intent is to stop the duplicate Octo IM ping now that an external PR triage agent files a tracked review issue, while forwarding the PR event to that agent via webhook.

## Spec compliance
The diff matches the stated intent exactly — nothing missing, nothing extra, no deviation.
- The reusable workflow at `@v1` declares `enable_im_notify` (`type: boolean`, default `true`) and the `PR_TRIAGE_WEBHOOK_URL` secret (`required: false`), so this caller-side change is correctly typed against the contract added by Mininglamp-OSS/.github#81.
- Downstream behavior is correct: in the reusable, `im-message` is gated by `inputs.enable_im_notify && ... || ''`, so `false` yields an empty IM message → the `octo-notify` action skips IM mode, while webhook mode still fires because `PR_TRIAGE_WEBHOOK_URL` is supplied. Net effect = webhook-only triage, as described.

## Security review (security-sensitive)
No issues found.
- The webhook URL is treated as a capability secret and passed through `secrets:` → `env:` in the shared action; it is never interpolated into a shell or JSON string. This follows the action's documented safe-handling contract.
- This PR does not alter the `pull_request_target` trigger or top-level `permissions: {}`. The job calls a pinned reusable workflow and does not check out or execute untrusted PR head code, so the well-known `pull_request_target` privilege-escalation risk is not introduced here.
- I confirmed `PR_TRIAGE_WEBHOOK_URL` is configured at the repo level, so the secret reference will resolve at runtime rather than POSTing to an empty URL.

## Nits (non-blocking)
- The file ends without a trailing newline (`\ No newline at end of file`). Harmless, but worth a one-character fix for cleanliness / POSIX-friendliness on the next touch.

## Verdict
No P0/P1 issues. The change is minimal, correctly typed against the reusable's contract, and safe. Approving.

PR #75, opened 2026-06-27
fix: space 校验 + 冷启动 + UI 优化批次

### LINKED ISSUE #77: Space 串消息问题
## 背景

Space 模式下出现跨 Space 的会话 / 消息串台展示，影响场景包括：

- 冷启动后 `WKConnection` 首次同步成功，但 `iSyncConversationChat` 监听尚未注册，`SpaceFilter` 缓存还没装载，最近会话只能看到 SYSTEM_BOTS。
- 群成员变更（被加入 / 被踢）时 `SpaceFilter` 的 `space_memberships` 缓存来不及刷新，新群在错的 Space 出现 / 旧 Space 仍展示已退出的群。
- PERSONAL 频道（真人私聊 / 普通 AI 机器人）没有自己的 Space 归属，未做兜底过滤，所有 Space 都会展示。

## 修复方向

- 冷启动：`initIMListener()` 同步注册到 IM 初始化同帧，SDK 侧 `setSyncConversationListener` 在 listener 缺失 / sync drop 路径补 `onBack(null)` 释放 permit。
- 群成员变更：1002 / 1003 通过 `isMyUidInGroupMemberMsgExtra` 精准判定是不是自己，`needsMembershipRefresh` 标志在循环外触发一次 `triggerSpaceMembershipsRefresh()`，复用 `SpaceSyncCoordinator` 500ms debounce。
- PERSONAL 频道：`SpaceFilter` 新增 `getLastMsgSpaceId()`，按 last-msg `content.space_id` 跟当前 Space 比对，系统 Bot 白名单短路放行，读不到 space_id 时 fail-open 等数据补齐。

依赖：server PR #154 `space_memberships` 作为权威源、未来 `avatar_cache_key` 字段。

mzanalyticst:
Closes #77

## Summary

本分支累积了 11 个 commit, 涵盖以下主题:

- **冷启动**: 同步化 WKIM listener 注册, 解决冷启只显示 SYSTEM_BOTS
- **Space 修复**: 群成员变更刷新 space_memberships; 1003 群减人事件防御性 refresh
- **诊断模式**: Space 串消息异步采集 + 关于页隐藏入口 / dialog UX / 清理临时 Log / 内部 API 标记
- **关注 Tab**: @badge 暂时下线 (等 server reminder.version 修复)
- **最近 Tab**: 放开群聊 / 子区 3 天活跃过滤, 全部展示
- **群头像**: 点击群聊顶部主动刷新头像缓存, 对齐 iOS
- 版本号 bump 到 1.3.4 (35)

## Test plan

- [ ] 冷启动: 杀进程冷启, 列表正常显示所有会话 (非只 SYSTEM_BOTS)
- [ ] Space: 1003 群减人后列表正常刷新, 跨 Space 切换无 race
- [ ] 诊断模式: 关于页隐藏入口可进入, dialog UX 正常
- [ ] 最近 Tab: 超过 3 天未活跃的群也展示, 角标包含老群未读
- [ ] 群头像: 群主换头像 → 点击群聊顶部头像/标题 → 返回最近 Tab 头像刷新
- [ ] 私聊 / 子区 / 其他 channel 类型回归无异常

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -37,8 +37,8 @@ android {
         applicationId appId
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
-        versionCode 34
-        versionName "1.3.3"
+        versionCode 35
+        versionName "1.3.4"
 
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
         vectorDrawables {
--- a/app/src/main/java/com/octoim/app/TSApplication.kt
+++ b/app/src/main/java/com/octoim/app/TSApplication.kt
@@ -61,6 +61,9 @@ class TSApplication : MultiDexApplication() {
             val defaultProcess = processName == getAppPackageName()
             if (defaultProcess) {
                 initAll()
+                // Space 串消息排查: 启动初始上下文 dump. 仅当用户开启过诊断模式才会真正落盘
+                // (DiagSink.isEnabled() 内部 gate, 未开启即纳秒级 no-op).
+                logBootDiagnostics(processName)
             }
         }
         registerActivityLifecycleCallbacks(object : ActivityLifecycleCallbacks {
@@ -266,6 +269,49 @@ class TSApplication : MultiDexApplication() {
                 NotificationManager::class.java
             )
             notificationManager.createNotificationChannel(channel)
+        }
+    }
+
+    /**
+     * Space 串消息排查: 启动时把决策上下文 dump 到诊断日志, 让我们 review 时不用问
+     * 用户"重启时 currentSpaceId 是什么".
+     *
+     * <p>DiagSink 内部 gate, 未开启诊断模式时这些调用是纳秒级 no-op, 不影响冷启动耗时.
+     */
+    private fun logBootDiagnostics(processName: String) {
+        try {
+            com.chat.base.utils.DiagSink.write(
+                "BOOT",
+                "process=" + processName +
+                    " sdk=" + Build.VERSION.SDK_INT +
+                    " device=" + Build.MANUFACTURER + "/" + Build.MODEL +
+                    " appPkg=" + getAppPackageName()
+            )
+        } catch (t: Throwable) {
+            // 防御性: BOOT 路径不能因诊断日志失败而崩
+        }
+        try {
+            val uid = WKConfig.getInstance().uid ?: ""
+            val currentSpaceId = com.chat.base.space.SpaceFilter.getCurrentSpaceId()
+            val currentSpaceName = com.chat.base.space.SpaceNameLookup.nameOf(currentSpaceId)
+            // appconfig 可能还未加载, 容错读
+            val sysBotsCount = try {
+                WKConfig.getInstance().appConfig?.system_bot_uids?.size ?: -1
+            } catch (t: Throwable) {
+                -1
+            }
+            com.chat.base.utils.DiagSink.write(
+                "BOOT",
+                "uid.len=" + uid.length +
+                    " currentSpaceId='" + currentSpaceId + "'" +
+                    " currentSpaceName='" + (currentSpaceName ?: "") + "'" +
+                    " systemBotsCount=" + sysBotsCount
+            )
+        } catch (t: Throwable) {
+            com.chat.base.utils.DiagSink.write(
+                "BOOT",
+                "read uid/space failed: " + t.javaClass.simpleName + ":" + t.message
+            )
         }
     }
 
--- a/wkbase/src/main/java/com/chat/base/WKBaseApplication.java
+++ b/wkbase/src/main/java/com/chat/base/WKBaseApplication.java
@@ -103,6 +103,14 @@ public class WKBaseApplication {
         // AES256-GCM / KeyStore 握手（50-150ms）搬出主线程。
         WKSharedPreferencesUtil.prewarm();
 
+        // Space 串消息排查诊断器(2026-06): 早初始化, 让后续启动链路上的 SF / BOOT 埋点
+        // 一旦上一次会话开启了诊断模式(SP 持久化), 就能立刻开始采集. 不会触发 IO, 仅起
+        // HandlerThread + 读 SP, 不阻塞主线程.
+        com.chat.base.utils.DiagSink.init(context);
+        // 桥接 wkim SDK 内部诊断埋点 → DiagSink. wkim 不能反向依赖 wkbase, 通过 install-pattern
+        // 让 ConversationManager 等 SDK 内部模块的 MEMBERSHIP 状态变更也能写到统一诊断文件.
+        com.xinbida.wukongim.diag.WKDiagWriter.install(com.chat.base.utils.DiagSink::write);
+
         float density = context.getResources().getDisplayMetrics().density;
         AndroidUtilities.setDensity(density);
 
--- a/wkbase/src/main/java/com/chat/base/common/WKCommonModel.java
+++ b/wkbase/src/main/java/com/chat/base/common/WKCommonModel.java
@@ -160,11 +160,14 @@ public class WKCommonModel extends WKBaseModel {
         WKChannel localChannel = WKIM.getInstance().getChannelManager().getChannel(entity.channel.channel_id, entity.channel.channel_type);
         boolean isRefreshContacts = false;
         if (localChannel != null && !TextUtils.isEmpty(localChannel.channelID)) {
-            // 复用已有的 avatarCacheKey；若为空则首次生成，作为 MyGlideUrlWithId 的稳定 cache key。
+            // avatarCacheKey 优先级：服务端 > 本地已有 > 新生成兜底。
+            // 服务端在头像变更时翻 avatar_cache_key 版本号，客户端透传即可让 URL 末尾的
+            // ?v=<key> 变化，绕开 Glide 缓存，实现自动刷新。
             // 不能每次都生成新 key：saveChannel 触发 refreshChannelInfo → adapter 重绘 →
             // fetchChannelInfo → saveChannel → 无限循环。只在首次为空时生成一次即可打破。
-            // -P-03: avatarCacheKey 改由服务端头像变更时翻版本号驱动；本地只负责在缺失时兜底生成。
-            if (!TextUtils.isEmpty(localChannel.avatarCacheKey)) {
+            if (!TextUtils.isEmpty(entity.avatar_cache_key)) {
+                wkChannel.avatarCacheKey = entity.avatar_cache_key;
+            } else if (!TextUtils.isEmpty(localChannel.avatarCacheKey)) {
                 wkChannel.avatarCacheKey = localChannel.avatarCacheKey;
             } else {
                 wkChannel.avatarCacheKey = UUID.randomUUID().toString().replaceAll("-", "");
@@ -174,8 +177,12 @@ public class WKCommonModel extends WKBaseModel {
             if (wkChannel.follow != entity.follow || wkChannel.status != entity.status)
                 isRefreshContacts = true;
         } else {
-            // 全新 channel，生成初始 avatarCacheKey
-            wkChannel.avatarCacheKey = UUID.randomUUID().toString().replaceAll("-", "");
+            // 全新 channel，优先用服务端值，缺失时兜底
+            if (!TextUtils.isEmpty(entity.avatar_cache_key)) {
+                wkChannel.avatarCacheKey = entity.avatar_cache_key;
+            } else {
+                wkChannel.avatarCacheKey = UUID.randomUUID().toString().replaceAll("-", "");
+            }
         }
         if (hashMap == null)
             hashMap = new HashMap<>();
--- a/wkbase/src/main/java/com/chat/base/entity/ChannelInfoEntity.java
+++ b/wkbase/src/main/java/com/chat/base/entity/ChannelInfoEntity.java
@@ -46,6 +46,7 @@ public class ChannelInfoEntity {
     public int device_flag;
     public String space_id;
     public String bot_creator_uid;
+    public String avatar_cache_key;
     public Map extra;
 
 
--- a/wkbase/src/main/java/com/chat/base/space/SpaceConversationPruner.java
+++ b/wkbase/src/main/java/com/chat/base/space/SpaceConversationPruner.java
@@ -38,8 +38,9 @@ import java.util.Set;
  * <ul>
  *     <li>不改 {@link SpaceFilter} 纯函数——本类只调用它做判定，不重新实现 7 分支逻辑。</li>
  *     <li>不写 WKSDK 持久化层——只裁内存 list / 白名单 Set，DB 回放时 SpaceFilter 会自然二次校验。</li>
- *     <li>PERSONAL 频道交给 {@link SpaceFilter#shouldSkipMessageForSpace} 做消息级过滤，
- *         本类不对私聊做 channel 级剔除，避免误杀跨 Space 共享的系统 Bot（如 botfather）。</li>
+ *     <li>PERSONAL 频道也走 prune（与 GROUP 对称）；{@link SpaceFilter} 的 person 分支
+ *         内部已对系统 Bot（botfather / u_10000 / fileHelper）白名单短路放行，
+ *         真人私聊 / 普通 AI 机器人按 last-msg.content.space_id 比对剔除。</li>
  * </ul>
  *
  * <p>Pattern 来源：iOS WKConversationListVM.pruneNonCurrentSpaceGroups（PR#95）扫
@@ -58,7 +59,9 @@ public final class SpaceConversationPruner {
      * <p>规则（与 iOS 对齐）：
      * <ul>
      *     <li>currentSpaceId 为空（非 Space 模式）→ 永不剔除</li>
-     *     <li>{@link WKChannelType#PERSONAL} → 永不剔除（交给消息级过滤 / SYSTEM_BOTS 兜底）</li>
+     *     <li>{@link WKChannelType#PERSONAL} → 调 {@link SpaceFilter#shouldSkipChannelForSpace}；
+     *         person 分支会先用 SystemBotsFallback 白名单短路（botfather / u_10000 / fileHelper
+     *         全 Space 显示），再读 last-msg.content.space_id 跟当前 Space 比对</li>
      *     <li>{@link WKChannelType#GROUP} → 调 {@link SpaceFilter#shouldSkipChannelForSpace}
      *         判定；Skip == true 则剔除</li>
      *     <li>其他 channelType（COMMUNITY_TOPIC 等）不走 prune 路径（上游已过滤）</li>
@@ -71,7 +74,7 @@ public final class SpaceConversationPruner {
                                       @NonNull SpaceFilter.ChannelInfoProvider provider) {
         if (channelID == null || channelID.isEmpty()) return false;
         if (currentSpaceId == null || currentSpaceId.isEmpty()) return false;
-        if (channelType != WKChannelType.GROUP) return false;
+        if (channelType != WKChannelType.GROUP && channelType != WKChannelType.PERSONAL) return false;
         return SpaceFilter.shouldSkipChannelForSpace(channelID, channelType, currentSpaceId, provider);
     }
 
--- a/wkbase/src/main/java/com/chat/base/space/SpaceFilter.java
+++ b/wkbase/src/main/java/com/chat/base/space/SpaceFilter.java
@@ -672,8 +672,12 @@ public final class SpaceFilter {
 
     /**
      * 私聊消息级 Space 过滤诊断日志. 跨 Space 私聊串消息的决策点必打 — 字段集与
-     * {@link #diagLog} 对齐, 多打消息自身元数据(发件人、内容前 60 字符片段)便于
-     * 用户能告诉我们"是哪条消息"加速定位.
+     * {@link #diagLog} 对齐, 多打消息自身元数据 (msgType / contentLen / contentHash /
+     * fromUid / clientMsgNo / messageSeq) 便于跨端比对定位"是哪条消息".
+     *
+     * <p>隐私: 不记录消息正文; 只记 content 长度与 hashCode (32-bit, 单向不可还原),
+     * 用于区分"同 fromUid+seq 但内容不同"的消息. 对齐 PR#75 review 反馈
+     * (redact-by-default), consent 文案无需声明采集消息正文.
      */
     private static void msgDiagLog(@Nullable WKMsg msg,
                                    @Nullable String currentSpaceId,
@@ -687,19 +691,20 @@ public final class SpaceFilter {
         try {
             String cid = msg == null ? "null" : msg.channelID;
             byte ct = msg == null ? -1 : msg.channelType;
+            int msgType = msg == null ? -1 : msg.type;
             String fromUid = msg == null ? "null" : msg.fromUID;
             String clientMsgNo = msg == null ? "null" : msg.clientMsgNO;
             long msgSeq = msg == null ? 0 : msg.messageSeq;
-            String contentSnippet = "";
-            if (msg != null && msg.content != null) {
-                int len = Math.min(msg.content.length(), 60);
-                contentSnippet = msg.content.substring(0, len).replace('\n', ' ').replace('\r', ' ');
-            }
+            int contentLen = (msg == null || msg.content == null) ? 0 : msg.content.length();
+            String contentHash = (msg == null || msg.content == null)
+                    ? "0"
+                    : Integer.toHexString(msg.content.hashCode());
             String curName = SpaceNameLookup.nameOf(currentSpaceId);
             String msgSpaceName = SpaceNameLookup.nameOf(msgSpaceId);
             String line = "branch=" + branch
                     + " skip=" + skip
                     + " channelID=" + cid + " channelType=" + ct
+                    + " msgType=" + msgType
                     + " fromUid=" + fromUid
                     + " clientMsgNo=" + clientMsgNo
                     + " msgSeq=" + msgSeq
@@ -707,7 +712,8 @@ public final class SpaceFilter {
                     + " currentSpaceName='" + (curName == null ? "" : curName) + '\''
                     + " msgSpaceId=" + msgSpaceId
                     + " msgSpaceName='" + (msgSpaceName == null ? "" : msgSpaceName) + '\''
-                    + " content[0..60]='" + contentSnippet + '\''
+                    + " contentLen=" + contentLen
+                    + " contentHash=" + contentHash
                     + " caller=" + callerOf();
             com.chat.base.utils.WKLogUtils.d(LOG_TAG, "SpaceFilter-Msg# " + line);
             com.chat.base.utils.DiagSink.write("SF-MSG", line);
--- a/wkbase/src/main/java/com/chat/base/space/SpaceFilter.java
+++ b/wkbase/src/main/java/com/chat/base/space/SpaceFilter.java
@@ -483,6 +483,7 @@ public final class SpaceFilter {
      * {@code TextUtils}），在 host-side 单元测试里会抛 "Stub!"；用 try/catch
      * 包住保证测试不被日志破坏（SpaceFilter 的纯函数路径必须保持 JVM-runnable）。
      */
+    @SuppressWarnings("RestrictedApi") // 诊断专用: 跨模块访问 wkim 内部 SpaceCacheStats, 排查结束后整段删除
     private static void diagLog(String channelID,
                                  byte channelType,
                                  @Nullable String currentSpaceId,
--- a/wkbase/src/main/java/com/chat/base/space/SpaceFilter.java
+++ b/wkbase/src/main/java/com/chat/base/space/SpaceFilter.java
@@ -465,8 +465,19 @@ public final class SpaceFilter {
      * <p>debug build 打印过滤决策全部关键变量：群 id / 归属 Space / 我自己的
      * source_space_id / 我的 row 是否已缓存 / 最终分支。用来定位「同账号
      * Web 端 2 个外部群全显示，Android 只显示 1 个」的数据缺失路径——
-     * 不猜逻辑，先让日志说话。Release build 自动 no-op（{@code WKLogUtils.d}
-     * 里判 {@code WKBinder.isDebug}）。
+     * 不猜逻辑，先让日志说话。
+     *
+     * <p><b>Space 串消息排查增强(2026-06)</b>: 除 {@code WKLogUtils.d}(仅 debug 进 logcat)
+     * 之外, 命中 {@link com.chat.base.utils.DiagSink#isEnabled()} 时同步写文件,
+     * release 包用户开启诊断模式后也能采到. 关键增量字段:
+     * <ul>
+     *     <li>{@code currentSpaceName / groupSpaceName} — 反查 Space 名, 不用查表</li>
+     *     <li>{@code channelInfo.space_id} 和 {@code convSync.space_id} — 两个原始源都打,
+     *         一眼看出 groupSpaceId 命中的是哪个</li>
+     *     <li>{@code cacheStats.refreshedAt / size / containsThisGroup / authoritative} —
+     *         判断决策时缓存是否 stale, 上次根因就是这类信号缺失</li>
+     *     <li>{@code caller} — 决策点调用栈一行, 区分 channel-list / msg-arrive 路径</li>
+     * </ul>
      *
      * <p>容错：{@code WKLogUtils.d} 依赖 Android framework（{@code Log}、
      * {@code TextUtils}），在 host-side 单元测试里会抛 "Stub!"；用 try/catch
@@ -481,28 +492,91 @@ public final class SpaceFilter {
                                  @Nullable Boolean myCached,
                                  @NonNull String branch,
                                  boolean skip) {
+        // 性能 gate: 二者全关时直接 return, 避免每次过滤决策都 new Throwable + 字符串拼接.
+        // - WKLogUtils.LOGGABLE: 编译时 = isDebug, debug 包恒 true, release 恒 false
+        // - DiagSink.isEnabled(): 运行时, 用户隐藏入口开启后才 true
+        if (!com.chat.base.utils.WKLogUtils.LOGGABLE && !com.chat.base.utils.DiagSink.isEnabled()) {
+            return;
+        }
         try {
             String channelName = "";
-            if (channelType == WKChannelType.GROUP) {
-                WKChannel ch = com.xinbida.wukongim.WKIM.getInstance().getChannelManager()
-                        .getChannel(channelID, channelType);
-                if (ch != null && ch.channelName != null) channelName = ch.channelName;
+            String channelInfoSpaceId = null;
+            if (channelType == WKChannelType.GROUP || channelType == WKChannelType.COMMUNITY_TOPIC) {
+                try {
+                    WKChannel ch = WKIM.getInstance().getChannelManager()
+                            .getChannel(channelID, channelType);
+                    if (ch != null) {
+                        if (ch.channelName != null) channelName = ch.channelName;
+                        if (ch.remoteExtraMap != null) {
+                            Object v = ch.remoteExtraMap.get(CHANNEL_EXTRA_SPACE_ID);
+                            if (v != null) channelInfoSpaceId = v.toString();
+                        }
+                    }
+                } catch (Throwable ignored) {
+                }
             }
-            String line = "SpaceFilter#"
-                    + " branch=" + branch
-                    + " skip=" + skip
-                    + " channelID=" + channelID
-                    + " name=" + channelName
-                    + " channelType=" + channelType
-                    + " currentSpaceId=" + currentSpaceId
-                    + " groupSpaceId=" + groupSpaceId
-                    + " prefix=" + prefix
-                    + " mySourceSpaceId=" + mySourceSpaceId
-                    + " myMembershipCached=" + myCached;
-            com.chat.base.utils.WKLogUtils.d(LOG_TAG, line);
+            // 两个原始源单独打: 让 review 时一眼看出 groupSpaceId 命中的是 channelInfo 还是 convSync.
+            String convSyncSpaceId = null;
+            com.xinbida.wukongim.manager.ConversationManager.SpaceCacheStats stats = null;
+            try {
+                convSyncSpaceId = WKIM.getInstance().getConversationManager()
+                        .getConvSyncSpaceId(channelID);
+                stats = WKIM.getInstance().getConversationManager()
+                        .getSpaceCacheStats(channelID);
+            } catch (Throwable ignored) {
+            }
+            String curName = SpaceNameLookup.nameOf(currentSpaceId);
+            String grpName = SpaceNameLookup.nameOf(groupSpaceId);
+            String caller = callerOf();
+
+            StringBuilder sb = new StringBuilder(384);
+            sb.append("branch=").append(branch)
+                    .append(" skip=").append(skip)
+                    .append(" channelID=").append(channelID)
+                    .append(" channelType=").append(channelType)
+                    .append(" name='").append(channelName).append('\'')
+                    .append(" currentSpaceId=").append(currentSpaceId)
+                    .append(" currentSpaceName='").append(curName == null ? "" : curName).append('\'')
+                    .append(" groupSpaceId=").append(groupSpaceId)
+                    .append(" groupSpaceName='").append(grpName == null ? "" : grpName).append('\'')
+                    .append(" channelInfo.space_id=").append(channelInfoSpaceId)
+                    .append(" convSync.space_id=").append(convSyncSpaceId)
+                    .append(" prefix=").append(prefix)
+                    .append(" mySourceSpaceId=").append(mySourceSpaceId)
+                    .append(" myMembershipCached=").append(myCached);
+            if (stats != null) {
+                sb.append(" cache.size=").append(stats.spaceMapSize)
+                        .append('/').append(stats.externalMapSize)
+                        .append(" cache.refreshedAt=").append(stats.refreshedAt)
+                        .append(" cache.containsThisGroup=").append(stats.containsChannelId)
+                        .append(" cache.authoritative=").append(stats.authoritative);
+            }
+            sb.append(" caller=").append(caller);
+            String line = sb.toString();
+            com.chat.base.utils.WKLogUtils.d(LOG_TAG, "SpaceFilter# " + line);
+            com.chat.base.utils.DiagSink.write("SF", line);
         } catch (Throwable ignored) {
             // 忽略日志异常（host-side 单测或 Android stub 环境）
         }
+    }
+
+    /**
+     * 取调用 SpaceFilter 决策的栈帧一行(跳过 SpaceFilter 内部帧). 用于区分
+     * "channel-list 路径" 还是 "msg-arrive 路径" 还是 "filter-and-display" 触发的过滤.
+     */
+    private static String callerOf() {
+        try {
+            StackTraceElement[] st = new Throwable().getStackTrace();
+            for (StackTraceElement el : st) {
+                String cls = el.getClassName();
+                if (cls == null) continue;
+                if (cls.equals(SpaceFilter.class.getName())) continue;
+                if (cls.startsWith("com.chat.base.space.SpaceFilter")) continue;
+                return el.getClassName() + "." + el.getMethodName() + ":" + el.getLineNumber();
+            }
+        } catch (Throwable ignored) {
+        }
+        return "unknown";
     }
 
     private static final String LOG_TAG = "SpaceFilter";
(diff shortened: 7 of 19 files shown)


Jerry-Xin (changes requested):
Summary: The PR is in scope and the cold-start and Space-refresh fixes look solid, but the new diagnostic mode logs private message content into an exportable file while the in-app consent prompt never discloses that message text is collected. That consent/redaction gap is the one blocking item; everything else is non-blocking.

**🔴 Blocking**
- 🔴 Critical: Diagnostic logging records private message content snippets into an exportable log file. In `wkbase/src/main/java/com/chat/base/space/SpaceFilter.java:639` the new `msgDiagLog()` copies up to 60 characters of raw `msg.content`, and `wkbase/src/main/java/com/chat/base/space/SpaceFilter.java:656` writes that snippet (alongside `fromUid`, `channelID`, `clientMsgNo`) into `DiagSink`, which is persisted to a user-exportable diagnostic file. The about-page opt-in dialog at `wkuikit/src/main/java/com/chat/uikit/setting/WKAboutActivity.java:144` only says it will "collect diagnostics" and "export the diagnostic log to share with us" — it does not disclose that message text is included. This applies on private-chat / cross-space paths, so exported logs can leak the contents of real conversations.
  Fix: remove message-body logging entirely and log only non-content metadata (message type, length, sequence, channel ID, a content hash). If user content truly must be captured, the consent dialog must explicitly state that message contents will be recorded and exported; redaction remains the safer default.

**💬 Non-blocking**
- 🟡 Warning: `DiagSink.init()` at `wkbase/src/main/java/com/chat/base/utils/DiagSink.java:88` unconditionally starts a dedicated `HandlerThread` (line 91) on every app startup, including when diagnostics are disabled, and it is called from `wkbase/src/main/java/com/chat/base/WKBaseApplication.java:109`. Since this PR is partly about cold start, consider reading the persisted TTL first and creating the writer thread lazily only when diagnostics are actually enabled.
- 🔵 Suggestion: Add focused tests for the membership-refresh trigger path, especially `isMyUidInGroupMemberMsgExtra()` parsing in `wkuikit/src/main/java/com/chat/uikit/chat/manager/WKIMUtils.java` (the "my uid present" vs "other user present" branches).

**✅ Highlights**
- Listener registration was moved back into synchronous startup in `wkuikit/src/main/java/com/chat/uikit/WKUIKitApplication.java`, which directly addresses the cold-start race described in the PR (only SYSTEM_BOTS showing).
- The `setSyncConversationListener()` null-listener fallback in `wkim/src/main/java/com/xinbida/wukongim/manager/ConversationManager.java` now releases upstream callbacks instead of leaving sync gates stuck.
- Server-provided `avatar_cache_key` is now preferred when available in `wkbase/src/main/java/com/chat/base/common/WKCommonModel.java`, which is the right direction for cache invalidation.

Verification note: a full `./gradlew :app:assembleDebug` was not run in this review environment (no Java runtime available), so the build-pass claim in the PR was not independently confirmed.

OctoBoooot (changes requested):
## Review: fix: space 校验 + 冷启动 + UI 优化批次 (#75)

**Verdict:** Request changes.
CI is red — `check-sprint` is failing, which is a required check and a merge precondition.
Risk tier: normal, but flagged **needs-human-review** — 18 files / 1086 lines across 6+ distinct concerns (size gate: >10 files). A human maintainer should confirm the bundling is acceptable.

### Blockers
- **CI red — `check-sprint` FAILURE.** The job exits with `No linked issue found. Please add a 'Closes #<issue>' reference to the PR description, then re-run this check.` The PR body has no issue link and no `linked_issues`. Same gate that #73/#69 hit — add `Closes #<n>` referencing the Space/cold-start issue (and ensure that issue is on the board with a Sprint set) and re-run. Verdict cannot be APPROVE while a required check is red.

### Major
- **New behavior-bearing logic ships with zero tests** (CONTRIBUTING: "if you're fixing a bug or adding a feature, please add tests"). Coverage ratio 0.06 across 18 files. The highest-value gap is `WKIMUtils.java:~1254 isMyUidInGroupMemberMsgExtra()` — a pure-ish JSON parser (`content.extra[].uid` vs my uid) that gates the entire space-membership refresh path. It is host-side unit-testable (feed it the documented `{type,content,extra:[{uid,name}]}` payloads incl. malformed/missing-extra/uid-not-present) and is the load-bearing decision for the "方案 B" fix. `SpaceFilter.shouldSkipMessageForSpace(msg, currentSpaceId)` is already `@VisibleForTesting` — the new branch labels (`space-empty-pass` / `fail-open-no-msg-space` / `msg-space-mismatch`) are equally cheap to pin. Failure mode of the parser is benign (returns false → no refresh, other sync paths backfill), so this is a test-debt major, not a correctness defect.

### Questions
- `WKIMUtils.java:~1302 triggerSpaceMembershipsRefresh()` — it calls `setSyncConversationListener(result -> complete())`, which writes the single shared one-shot `iSyncConversationChatBack` slot also used by `ChatFragment.performSpaceSwitch` (the back that hides the space-switch overlay). The `SpaceSyncCoordinator` permit looks like it prevents the two from overlapping (membership-refresh `tryBegin` drops while a switch holds the permit), and the 10s hard-timeout covers a stuck overlay — but I couldn't construct the exact interleaving to rule out a membership-refresh back clobbering an in-flight switch back. Can you confirm the permit fully serializes these two callers, or does the last `setSyncConversationListener` win regardless of permit state?

### Praise
- The cold-start fix is well-reasoned: the double guard — SDK-side `onBack(null)` to release the upstream permit when `iSyncConversationChat == null` (`ConversationManager.java:436`) **plus** moving `initIMListener()` back to synchronous so the listener is registered before connect (`WKUIKitApplication.java:239`) — addresses both the symptom and the root cause, not just one. The `ChatFragment` sync back already null-checks `result`, so the new `onBack(null)` path is safe.
- The membership-refresh trigger rules (1002 → check my uid in `extra`; 1003 → don't trigger, with a documented defensive fallback; 1008 → skip) are each grounded in specific server source line refs. That's the right level of rigor for a race-window fix.

- This PR introduces a sizable temporary diagnostic framework (`DiagSink`, `WKDiagWriter`, `SpaceNameLookup`, `ConversationManager.SpaceCacheStats`, the hidden 5-tap entry) that the author explicitly marks for deletion after the investigation closes. It's gated (no-op unless user-enabled, `RestrictTo` on the SDK API) and ships dark, so not blocking — but it's a lot of release-bound scaffolding riding on a "fix" PR. Consider tracking its removal with a follow-up issue so it doesn't outlive the investigation.
- `DiagnosticLogFile.read()` now buffers up to the full 5 MB file into a single `String` for "export diagnostic log" (was 1 MB). Bounded, but worth a streamed share if export is ever hit on low-memory devices.

*18 files / +992 −94, 11 commits. needs-human-review applied (size gate).*

lml2468:
## Dispatch — PR Review Fan-out

Review initiated for this PR by `coding-team` leader. Three parallel review slots:

| Slot | Reviewer | dispatch_id |
|------|----------|-------------|
| QA | qa-engineer | `octo-android#75\|review\|qa` |
| Security | security-engineer | `octo-android#75\|review\|security` |
| Code | code-reviewer | `octo-android#75\|review\|code` |

⏳ Waiting on all three verdicts. No merge until all slots report.
---
*Automated dispatch by Multica leader agent. Do not merge until all reviews clear.*

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# octo-android PR#75 Review Report
**Head SHA:** `d1b1b22758e04b15110cc3c3cb700d82780b05b0`
**Scope:** 18 files, +992/−94, 11 commits — space filtering + cold-start + diagnostic mode + UI batch

---

## 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| Cold-start fix (sync initIMListener) | ✅ Correct | `WKUIKitApplication.java` L235–237: moved out of `postPhaseB`; `ConversationManager.java` L434–439: `onBack(null)` fallback |
| Space membership refresh (方案 B) | ✅ Correct | `WKIMUtils.java` L235–337: `needsMembershipRefresh` flag + `triggerSpaceMembershipsRefresh()` via SpaceSyncCoordinator debounce |
| 1003 defensive refresh | ✅ Correct | `WKIMUtils.java` L264–271: defensive only, gated by `isMyUidInGroupMemberMsgExtra` |
| Avatar cache key fix | ✅ Correct | `WKCommonModel.java` L160–190: server > local > UUID priority; `ChatActivity.java` L912–922: manual refresh workaround |
| Recent tab 3-day filter removal | ✅ Correct | `ChatFragment.java` L1419–1430, L2861–2864, L2880–2886: consistent removal across badge, display list, and isInactiveGroup |
| @ badge temporary disable | ✅ Correct | `ChatFragment.java` L1477–1484: clean no-op, per-row `[有人@你]` tag independent path preserved |
| DiagSink / diagnostic framework | ✅ Sound | `DiagSink.java`: AtomicBoolean gate + HandlerThread async + TTL auto-expire + SP persistence |
| DiagnosticLogFile ring buffer | ✅ Sound | `DiagnosticLogFile.java`: 5MB cap → keep 2.5MB tail, proper rotation with banner |
| Version bump | ✅ | `build.gradle`: 1.3.3→1.3.4, versionCode 34→35 |

---

## 2. Findings

### P2-1: Dual-layer stuck-reset timing gap (10s vs 15s)

**diff-scope:** pre-existing (both timers existed before this PR; PR adds `triggerSpaceMembershipsRefresh` as a new consumer of the same path)

`SpaceSyncCoordinator.STUCK_RESET_MS = 10_000` vs `ConversationManager.SYNC_STUCK_RESET_MS = 15_000`. At the 10–15s window, the coordinator allows a new `tryBegin`, but the SDK's `syncInFlight` is still true and drops the sync, calling `onBack(null)` → coordinator `complete()`. Net effect: wasted sync attempts for 5s, then self-heals. Not a deadlock, but an efficiency gap.

**Recommendation:** Align both constants or document the intentional gap.

### P2-2: Diagnostic `callerOf()` creates Throwable per filter decision when enabled

**diff-scope:** new (this PR)

`SpaceFilter.callerOf()` calls `new Throwable().getStackTrace()` on every `shouldSkipChannelForSpace` invocation when DiagSink is enabled. On a busy channel list this can fire hundreds of times. Acceptable for a 2-hour-TTL diagnostic tool, but worth noting.

**Recommendation:** Consider sampling (e.g., every Nth call) or caching the caller for the same code path.

### P2-3: `iSyncConversationChat` field not volatile

**diff-scope:** pre-existing (PR adds `onBack(null)` defense-in-depth which mitigates the visibility gap)

`ConversationManager.iSyncConversationChat` is a plain instance field. Written on the thread calling `initIMListener()`, potentially read on a different thread when sync is triggered from `WKConnection`. On ARM there's no guaranteed visibility without a happens-before edge. In practice Android's startup sequence provides sufficient ordering, but technically fragile.

**Recommendation:** Add `volatile` to `iSyncConversationChat` for correctness.

---

## 3. Recommendations

1. **Align stuck-reset timers** (P2-1): Set both to 15s or add a comment explaining why they differ.
2. **Add `volatile`** to `iSyncConversationChat` (P2-3): Low-cost correctness fix.
3. **Avatar cache key**: The `ChatActivity` title-click workaround is necessary today because server doesn't return `avatar_cache_key`. Track server-side fix to remove the workaround.

---

## 4. Extra Observations

- **SpaceFilter authoritative-cache vs newly-joined groups:** Once `authoritative=true`, `SpaceFilter` skips groups not in cache rather than failing open. The new `triggerSpaceMembershipsRefresh` (方案 B) closes this race window to ~620ms (HTTP RTT). This is the correct mitigation.
- **`setSyncConversationListener` naming:** Method name suggests "register a listener" but it actually triggers a sync. The `onBack(null)` fallback correctly releases the permit without starting a sync. Well-handled.
- **Diagnostic mode removal plan is well-documented:** Every diagnostic commit includes explicit removal steps. `grep "DiagSink.write"` + delete 3 files. Clean temporary code.

---

## 5. Data-Flow Tracing

### Cold-start sync chain
```
WKUIKitApplication.init() [sync]
  → initIMListener() [sync, main thread]
    → addOnSyncConversationListener(callback)
      → iSyncConversationChat = callback  ← field set BEFORE WKConnection
  → WKConnection.connectSuccess() [async, later]
    → SyncGate.allow → tryBegin → setSyncConversationListener(onBack)
      → iSyncConversationChat != null ✅
      → CAS permit → syncConversationChat() → HTTP sync
      → saveSyncChatImpl → applySpaceMemberships → cache populated
      → onBack(result) → SyncGate.done → complete()
```

**Key verification:** `iSyncConversationChat` is set synchronously during startup, before any WKConnection callback can fire. The `onBack(null)` fallback in ConversationManager is defense-in-depth for edge cases where the listener might not yet be registered.

### 方案 B membership refresh chain
```
WKIMUtils.onSyncMsgListener [SDK msg thread]
  → for each msg: type check (1002/1003/1005)
    → isMyUidInGroupMemberMsgExtra: parse msg.content JSON → extra[].uid == myUid
    → needsMembershipRefresh = true (if hit)
  → if needsMembershipRefresh:
    → triggerSpaceMembershipsRefresh()
      → SpaceSyncCoordinator.tryBegin("groupMembershipChange") [CAS + 500ms debounce]
      → IO thread: setSyncConversationListener(complete callback)
        → CAS permit → syncConversationChat() → HTTP conv/sync
        → saveSyncChatImpl → applySpaceMemberships(memberships)
        → onBack → complete()
```

**Key verification:** `msg.content` is the server-serialized JSON payload for system messages. Structure `{type, content, extra:[{uid,name}]}` confirmed by server `modules/group/event.go:645-651` and iOS `WKSystemMessageCell.m:80-102`. The `needsMembershipRefresh` flag is local to each callback invocation (no shared state). SpaceSyncCoordinator prevents burst syncs.

### Avatar cache refresh chain
```
ChatActivity titleView click [main thread]
  → AvatarView.clearCache(channelId, channelType) → delete disk file
  → ch.avatarCacheKey = UUID.randomUUID() → new Glide cache key
  → setRefreshChannel(ch, true) → notify observers → adapter rebind
  → Glide: memory miss (new key) + disk miss (deleted) → network fetch
```

**Key verification:** `entity.avatar_cache_key` (from server) takes priority in `saveChannel`, but server currently returns null, so the manual UUID flip in ChatActivity is the active path. Future-proofed for server-side field.

---

## 6. Blind-spot Checklist (R5)

- **C1 Dual-path parity:** N/A — no add↔remove / subscribe↔unsubscribe symmetry in this PR. The 1002/1003 split is asymmetric by server design (1003 never reaches the kicked user), and the PR correctly handles each path independently.
- **C2 Control-flow ordering / nested reuse:** CLEAR — `triggerSpaceMembershipsRefresh` → `setSyncConversationListener` → `SpaceSyncCoordinator.complete()` is called exactly once per sync (CAS permit ensures single entry; `onBack` callback fires once after save). No double-complete risk.
- **C3 Authorization boundary:** N/A — no permission/jail/tool exposure changes.
- **C4 Authorization lifecycle / container-member state:** N/A — no auth changes.
- **C5 Build ≠ runtime:** N/A — standard Android app build, no extension/CLI/heredoc/relative-URL concerns.
- **C6 Governance / policy docs:** N/A — no policy/governance documents in this PR.

---

## 7. Cross-round Blocker Re-check (R6)

N/A — first review of this PR.

---

**[Octo-Q] verdict: APPROVE**

This is a well-structured batch PR. The cold-start fix is correct (sync listener registration + defense-in-depth null callback). 方案 B (membership refresh trigger) correctly closes the cross-Space race window with proper debouncing. All three fail-open paths in SpaceFilter remain intact and are now better instrumented. Diagnostic tooling is gated, async, TTL-limited, and has a clear removal plan. Three P2 observations (stuck-reset timer gap, diagnostic Throwable cost, non-volatile field) are noted but none meets the P1 bar per R1 (no path becomes unusable, no user-visible incorrect data, no regression vs. pre-PR state).

lml2468 (commented):
### QA review — Verdict: PASS-WITH-RISK (functional/quality), merge gated by the existing privacy blocker

**Method / limitation:** ran static analysis + symbol/regression review; the CI test suite was not executed in my env (no JDK/Android SDK). Per fail-closed I do not certify "all green" — please run `:wkbase:testDebugUnitTest :wkim:testDebugUnitTest :wkuikit:testDebugUnitTest` before merge. (The red `check-sprint` is the project-board Sprint-assignment gate, not a build/test failure.)

**Verified safe:** all new symbol references resolve (compiles); `SpaceFilterTest` not regressed (return logic of `shouldSkipMessageForSpace` unchanged, diag gate `WKLogUtils.LOGGABLE←BuildConfig.DEBUG` is JVM-safe and short-circuits in host tests); cold-start `onBack(null)` permit-release is self-consistent; avatar cacheKey path is loop-safe; Recent-tab 3-day filter removal correctly leaves the Follow-tab filter intact.

**🔴 Blocking (concur with @Jerry-Xin, not re-adjudicating):** diagnostic mode writes 60 chars of private `msg.content` to an exportable log without consent disclosure. QA agrees this gates merge — and note this path has **no test/redaction coverage**.

**🟡 QA primary ask — zero tests for new logic:** this PR adds **no test files**, yet `isMyUidInGroupMemberMsgExtra()` (JSON gate that decides whether the core membership-sync fix fires; impact×reach ≈ 3×3), the `DiagSink` TTL state machine, and `DiagnosticLogFile.rotateIfOversized()` ring buffer are all pure-JVM-testable. Recommend adding unit tests (esp. for `isMyUidInGroupMemberMsgExtra` malformed-JSON / missing-`extra` / uid-match cases) before merge or as a tracked follow-up.

**🟢 minor/nit:** `triggerSpaceMembershipsRefresh` shares the `SpaceSyncCoordinator` permit with `performSpaceSwitch` (a Space-switch can be dropped while a membership refresh is in flight — self-heals via STUCK_RESET; worth a regression case); `DiagnosticLogFile.read()` loads the whole 5MB ring buffer into memory on export; tapping the group header regenerates the avatar key + refresh on every tap.

lml2468:
## Code Review Verdict: APPROVE_WITH_NITS

> 功能性代码评审（正确性 / 可读性 / 可维护性 / 设计契合），与 QA、Security 两个 slot 独立。范围内**无 Must Fix**。唯一阻断项（诊断日志写入私聊消息内容）属 security slot 判定，已由 @Jerry-Xin 提出 → 我标 `security-followup`，不在 code slot 重复阻断。最终能否 merge 取决于 security verdict。

### 正确性（逐项核对，均通过）

1. **冷启动同步化** — `WKUIKitApplication` 把 `initIMListener()` 从 `postPhaseB` 异步改回与 `initIM()` 同帧同步注册，根因正确（listener 必须先于 WKConnection 首次连接成功就位）。`ConversationManager.setSyncConversationListener` 在 `iSyncConversationChat==null` 及"sync 已 in-flight / drop"两条短路路径上都补了 `onBack(null)`，释放上层 `SpaceSyncCoordinator/SyncGate` permit——验证过：`addOnSyncConversationListener` 设的是常驻 listener，`setSyncConversationListener` 只传一次性回调，二者不互相覆盖，无 listener 被顶掉的问题。`syncInFlight` CAS + `SYNC_STUCK_RESET_MS` 兜底，无死锁。

2. **群成员变更刷新 + 1003 防御性 refresh** — `WKIMUtils` 用 `needsMembershipRefresh` 标志在循环外只触发一次 sync，事件批次自然合并，设计干净。1002/1003 通过 `isMyUidInGroupMemberMsgExtra` 精准判定"是不是我自己"，JSON 解析全程 try/catch fallback false。**未引入新 race**：`triggerSpaceMembershipsRefresh` 复用 `SpaceSyncCoordinator.tryBegin` 的 500ms debounce + 全局重入守卫；与 cold-start/spaceSwitch 抢同一 permit 时被 drop，但 drop 路径的 `onBack(null)→complete()` 保证 permit 必释放。1003 分支当前 server 实现下不会进，进了也只是多一次幂等 sync，对抗自检后确认零副作用。

3. **最近 Tab 放开 3 天过滤** — `isInactiveGroup` 简化为仅 `lastMsgTimestamp<=0`；`buildRecentDisplayList`（展示）与 `setAllCount`（角标）**两处同步**去掉 3 天阈值，展示与未读计数语义一致，无割裂。

4. **群头像缓存刷新** — `ChatActivity` 点击群聊顶部时 `AvatarView.clearCache` + 翻 `avatarCacheKey`(UUID) + `setRefreshChannel`。`AvatarView.clearCache` / `setRefreshChannel` 签名均已确认存在。`WKCommonModel.saveChannel` 新优先级（server > 本地已有 > 新生成）下，下一次 `fetchChannelInfo→saveChannel` 会保留刚翻的本地 key（server 当前返回 null），不会触发"每次 save 都翻 key"的无限重绘循环——这一点处理正确。

5. **诊断基建（DiagSink / WKDiagWriter / SpaceNameLookup / DiagnosticLogFile）** — 工程质量高：install-pattern 桥正确规避了 wkbase↔wkuikit / wkim 的反向依赖；热路径 gate (`!LOGGABLE && !isEnabled()`) 先于任何 `new Throwable`/字符串拼接，性能正确；TTL 自动失效 + 跨重启恢复；全程 try/catch 不污染业务。`DiagnosticLogFile` 从"超 1MB 直接 delete"改成 5MB ring-buffer（保留尾部 2.5MB + 跳半行 + banner），并用 try-with-resources 修了旧代码的流泄漏——是实打实的改进。

### Should Fix

- **诊断代码生命周期治理** `DiagSink.java` / `WKDiagWriter.java` / `SpaceNameLookup.java` + 各埋点 ~600 行均标注"排查结束后删除"，分散在注释里。**建议**：开一个 tracking issue（关联 server PR #154 闭环），把各文件已写好的"移除步骤"汇总进去并在代码里留 `// TODO(#<issue>)` 锚点，避免临时诊断代码永久沉淀、release 包长期携带。`SpaceCacheStats` 已用 `@RestrictTo(LIBRARY_GROUP_PREFIX)` + `@hide` 标记内部 API，避免业务依赖，这点处理得对。

### Nit

- `ChatActivity` 头像刷新：每次点击群聊顶部都翻新 UUID，即使头像未变也会强制 Glide cache-miss 重新下载一次。属对齐 iOS 的有意取舍（server 不下发 `avatar_cache_key`），可接受；待 server 补字段后走 `entity.avatar_cache_key` 路径即自然消除。
- `DiagnosticLogFile.MAX_SIZE` 1MB→5MB 对 release 包的 crash/ANR 日志同样生效（诊断文件最多占 5MB）。已在注释说明，属合理取舍。

### Question
- 无阻断性疑问。

### 转交其它 slot（不在 code 范围内代办）
- `security-followup`：`SpaceFilter.msgDiagLog` 把私聊 `msg.content` 前 60 字符写入可导出诊断文件（`SpaceFilter.java` 约 :639/:656），而 `WKAboutActivity` 开启 dialog 文案未披露会采集消息正文。**这是本 PR 唯一阻断项，由 security slot 判定**（@Jerry-Xin 已提出）。从纯功能正确性看该代码按预期工作，故不计入 code slot 的 Must Fix。

—— code-reviewer (coding-team 评审三件套之一)

yujiawei (dismissed):
# Code Review — PR #75 (octo-android)

Independent review at head `d1b1b227`. 18 files, +992/−94, 11 commits: space-membership race fix, cold-start sync fix, a gated diagnostic-logging framework, and a UI batch (avatar cache key, recent-tab filter, @-badge). I independently traced the load-bearing changes and verified every cross-file symbol the diff depends on. **Verdict: Approve** — no blocking (P0/P1) issues; three non-blocking P2 notes below.

## Spec / scope

The diff matches the stated intent of the 11 commits and adds nothing out of scope. The diagnostic framework (`DiagSink`, `WKDiagWriter`, `SpaceNameLookup`, `getSpaceCacheStats`) is default-off, gated behind a hidden 5-tap entry with a 2-hour TTL, and ships with an explicit, greppable removal plan documented in each new file. No production behavior changes when diagnostics are off.

## Correctness — verified

**Cold-start fix (the highest-risk change).** `initIMListener()` was moved out of `AppStartup.postPhaseB` to run synchronously in `WKUIKitApplication.init()` (`WKUIKitApplication.java:235-239`). I traced the body of `initIMListener()` (`WKIMUtils.java:148-531`): it only registers listeners and endpoint methods — the DB/HTTP calls live inside lambdas that fire later on events, not at registration time. So running it synchronously adds negligible main-thread cost and **no ANR risk**, while guaranteeing `addOnSyncConversationListener` has set `iSyncConversationChat` before any `WKConnection` success callback can trigger a sync. The paired `onBack(null)` defense in `ConversationManager.setSyncConversationListener` (`ConversationManager.java:441`) correctly releases the upstream permit on the null-listener path so the coordinator can't stay locked for the full stuck-reset window. This is a clean two-layer fix.

**Membership-refresh trigger (方案 B).** `WKIMUtils.java:235-337` accumulates a `needsMembershipRefresh` flag across the message batch and fires one `triggerSpaceMembershipsRefresh()` after the loop, behind `SpaceSyncCoordinator.tryBegin` (CAS + per-path debounce). The add-member (1002) path is correctly gated by `isMyUidInGroupMemberMsgExtra`, and the remove-member (1003) path is gated the same way as documented defense-in-depth (the kicked user is unsubscribed before the 1003 is sent, so it should never self-hit today; if it does, it debounces to a no-op). I confirmed the permit lifecycle is leak-free: every exit path of `setSyncConversationListener` calls `onBack` → `complete()`, so `complete()` runs exactly once per `tryBegin`.

**Avatar cache key.** `WKCommonModel.java` now applies server > local > UUID priority and preserves the original anti-infinite-loop comment; the `ChatActivity` title-click path (`ChatActivity.java:914-922`) is the active workaround while the server omits `avatar_cache_key`. `AvatarView.clearCache(channelID, channelType)` exists (`AvatarView.java:205`). Correct.

**Recent-tab 3-day filter removal.** Removed consistently across all three sites (badge count, display list, `isInactiveGroup`) — no orphaned half-applied filter.

**Cross-file symbols verified present:** `getConvSyncSpaceId`, `getSpaceCacheStats`, `WKLogUtils.LOGGABLE`, `SpaceFilter.getCurrentSpaceId`, `CHANNEL_EXTRA_SPACE_ID`, `SpaceModel.getCachedSpaces`/`cachedSpaces`, `SpaceEntity.name`/`space_id`, `WKChannelType.COMMUNITY_TOPIC`.

## Non-blocking findings (P2)

- **P2-1 — Stuck-reset timer mismatch.** `SpaceSyncCoordinator.STUCK_RESET_MS = 10s` vs `ConversationManager.SYNC_STUCK_RESET_MS = 15s`. In the 10–15s window the coordinator admits a new `tryBegin` while the SDK still holds `syncInFlight`, so the sync is dropped and self-heals via `onBack(null)`. Pre-existing; this PR only adds a new consumer of the path. Worth aligning the constants or documenting the intentional gap.
- **P2-2 — `iSyncConversationChat` not `volatile`.** Written on the init thread, read when a sync is triggered from a connection callback. The new synchronous registration plus the `onBack(null)` fallback make this safe in practice, but adding `volatile` would make the visibility guarantee explicit at near-zero cost.
- **P2-3 — `callerOf()` allocates a `Throwable` per filter decision when diagnostics are enabled.** Fine for a 2-hour TTL diagnostic tool, but on a busy channel list this fires frequently; consider sampling if it ever proves noisy. No impact when diagnostics are off (the gate returns before `callerOf()`).

## Coverage / blind spots

Scope is client-side only — no auth, permission, or build-vs-runtime surface in this diff. The 1002/1003 asymmetry is by server design and handled independently per path, so there is no dual-path-parity defect. The main residual dependency is server PR #154 (`space_memberships` as the authoritative source) and a future server `avatar_cache_key`; both are documented as the conditions under which the diagnostic/workaround code is removed.

yujiawei (changes requested):
# Code Review — PR #75 (octo-android)

> Correction to my earlier note on this PR: I initially approved at this head SHA, then re-traced the diagnostic data flow against the consent text and confirmed a privacy blocker that I had missed. Dismissing the prior approval and submitting this corrected verdict at the same head `d1b1b227`. **Verdict: Request changes.**

Independent review at head `d1b1b227`. 18 files, +992/−94, 11 commits: space-membership race fix, cold-start sync fix, a gated diagnostic-logging framework, and a UI batch. I independently traced the load-bearing changes and verified the cross-file symbols the diff depends on.

## Blocking (P1)

**Diagnostic mode writes private message content into an exportable log without consent disclosure.**

Data flow, all on the message-filter path:
- `wkbase/.../space/SpaceFilter.java:636-639` — `msgDiagLog()` copies up to 60 chars of raw `msg.content`.
- `SpaceFilter.java:653,656` — writes `content[0..60]='…'` together with `fromUid` and `clientMsgNo` into `DiagSink`.
- `DiagSink.write` → `DiagnosticLogFile.append` persists this to `diagnostic.log`.
- `DiagnosticLogFile.share()` (`DiagnosticLogFile.java:138-147`) exports that file to any app via `FileProvider` + `ACTION_SEND`.

`msgDiagLog()` is reachable on every branch of `shouldSkipMessageForSpace` (`SpaceFilter.java:599-613`), including private-chat / cross-space paths, so an exported log can contain the text, sender, and message identifiers of real conversations.

The opt-in dialog at `WKAboutActivity.java:144-145` says only that it will "开启 2 小时诊断采集来定位 Space 消息串台问题" and asks the user to "导出诊断日志分享给我们" — it never states that **message contents** are captured and exported. That is a consent/redaction gap, and it is the one item that blocks merge.

**Fix (preferred):** drop the body snippet entirely and log only non-content metadata — message type, `length`, `messageSeq`, `channelID`, and (if a body discriminator is genuinely needed) a hash of the content rather than the plaintext. The branch label + `msgSpaceId`/`currentSpaceId` fields already carry the diagnostic signal needed for the串台 investigation; the 60-char plaintext adds little and carries the privacy cost. If user content truly must be captured, the consent dialog must explicitly disclose that message contents will be recorded and exported — but redaction is the safer default.

## Correctness — verified (the rest of the PR is sound)

- **Cold-start fix.** `initIMListener()` moved out of `postPhaseB` to run synchronously (`WKUIKitApplication.java:235-239`). I traced its body (`WKIMUtils.java:148-531`): it only registers listeners/endpoint methods — the DB/HTTP calls live inside lambdas that fire later on events, not at registration time. So synchronous execution adds negligible main-thread cost and **no ANR risk**, while guaranteeing `iSyncConversationChat` is set before any connect callback triggers a sync. The paired `onBack(null)` defense (`ConversationManager.java:441`) correctly releases the upstream permit on the null-listener path. Clean two-layer fix.
- **Membership-refresh trigger (方案 B).** `WKIMUtils.java:235-337` accumulates `needsMembershipRefresh` and fires one `triggerSpaceMembershipsRefresh()` after the batch, behind `SpaceSyncCoordinator.tryBegin` (CAS + debounce). 1002 add gated by `isMyUidInGroupMemberMsgExtra`; 1003 remove documented as defensive-only. Permit lifecycle is leak-free — every exit of `setSyncConversationListener` calls `onBack` → `complete()` exactly once per `tryBegin`.
- **Avatar cache key.** server > local > UUID priority preserved with the anti-loop comment; `AvatarView.clearCache` exists (`AvatarView.java:205`).
- **Recent-tab 3-day filter removal** applied consistently across all three sites.
- Cross-file symbols verified present: `getConvSyncSpaceId`, `getSpaceCacheStats`, `WKLogUtils.LOGGABLE`, `getCurrentSpaceId`, `CHANNEL_EXTRA_SPACE_ID`, `getCachedSpaces`, `SpaceEntity.name`/`space_id`, `WKChannelType.COMMUNITY_TOPIC`.

## Non-blocking (P2)

- **Stuck-reset timer mismatch.** `SpaceSyncCoordinator.STUCK_RESET_MS = 10s` vs `ConversationManager.SYNC_STUCK_RESET_MS = 15s`; in the 10–15s window the coordinator admits a new `tryBegin` while the SDK still holds `syncInFlight`, so the sync is dropped and self-heals via `onBack(null)`. Pre-existing; align the constants or document the gap.
- **`iSyncConversationChat` not `volatile`.** Safe in practice given the new synchronous registration + `onBack(null)` fallback, but `volatile` would make the visibility guarantee explicit at near-zero cost.
- **`callerOf()` allocates a `Throwable` per filter decision** when diagnostics are enabled. Fine for a 2-hour-TTL tool; consider sampling if noisy. No cost when diagnostics are off.
- **`DiagSink.init()` starts a `HandlerThread` on every startup** (`DiagSink.java:88-91`, called from `WKBaseApplication.java:109`) even when disabled — since this PR touches cold start, consider creating the writer thread lazily only when diagnostics are actually enabled.
- **Test debt.** New behavior-bearing logic ships with no unit tests; `isMyUidInGroupMemberMsgExtra()` (the JSON gate for the whole membership-refresh fix) and `DiagnosticLogFile.rotateIfOversized()` are pure-JVM testable — worth adding before merge or as a tracked follow-up.

## Notes

- A separate required check, `check-sprint`, is reported red (missing `Closes #<issue>` link). That is a project-board gate enforced independently of this review, not a build/test failure — flagging for the author/merger to resolve.
- The diagnostic framework is a sizable release-bound scaffold riding on a fix PR; it ships dark and gated with a documented removal plan, so it is not itself blocking, but tracking its removal with a follow-up issue is advisable so it doesn't outlive the investigation.

OctoBoooot (changes requested):
## Re-review (delta): fix: space 校验 + 冷启动 + UI 优化批次 (#75) @ 31189844

**Verdict:** Request changes. None of the three prior blocking items are resolved at this head; the delta adds new (untested) PERSONAL-channel space filtering rather than addressing the open blockers.

Delta from d1b1b227 → 31189844: 2 files (`SpaceFilter.java` +61/−4, `SpaceConversationPruner.java` +7/−4).

### Prior items → status at this head
- ✗ **CI still red.** `check-sprint` FAILURE (run 28286536388) — PR body still has no `Closes #<issue>`. Same merge precondition as before; unchanged.
- ✗ **Privacy 🔴 (raised by Jerry-Xin / yujiawei, third reviewer converged).** `SpaceFilter.java:696,710,713` — `msgDiagLog` still captures `msg.content.substring(0,60)` ("content[0..60]") and writes it via `DiagSink.write("SF-MSG", …)` into the user-exportable diagnostic file, while the `WKAboutActivity` consent dialog still never discloses that message content is recorded. Untouched in this delta.
- ✗ **Test debt.** Still no tests for the membership-refresh parser or the SpaceFilter branch labels. The delta *adds* a new untested decision path (see below), widening the gap.

### Major (new in this delta)
- `SpaceFilter.java:399-423` — the PERSONAL branch changed from unconditional pass (`person-pass`) to active filtering: system-bot whitelist short-circuit, then `getLastMsgSpaceId()` vs `currentSpaceId`, else fail-open. This is a real behavior change to channel-list visibility (a private chat can now be *hidden* from a Space), and it ships with no test. Construct: a real-person PERSONAL conversation whose last msg carries `space_id=A` is now pruned when `currentSpaceId=B` — correct per intent, but the `isBlank(msgSpaceId)` fail-open vs mismatch boundary is exactly the kind of logic that needs a pinned test. `shouldSkipChannelForSpace` is testable via the `ChannelInfoProvider` seam (stub `getLastMsgSpaceId`).

### Questions
- `SpaceConversationPruner.java:77` — PERSONAL channels now go through `shouldPrune`. Is there a guarantee `getLastMsgSpaceId` is cheap on the channel-list render path? The javadoc says lazy-load from `MsgDbManager` on first access then cached — but on a cold list with N personal chats, the first prune pass could be N synchronous DB lookups on whatever thread prune runs. Worth confirming prune isn't on the main thread, or that the lookup is already warm.

### Praise
- The PERSONAL-filter design is well-grounded: system-bot whitelist short-circuit (botfather / u_10000 / fileHelper) before the space comparison correctly avoids hiding shared system bots, and it cites the iOS parity path (`WKConversationListVM.m:952-979`). The fail-open-on-missing-space_id choice is the right default for not nuking history.

- Carryover: the temporary diagnostic framework is still release-bound scaffolding marked for later deletion — track its removal with a follow-up issue.

*Gate at this head: reviewDecision=CHANGES_REQUESTED, mergeState=BLOCKED. needs-human-review applied (19 files). The three blocking items above must each be addressed before merge.*

Jerry-Xin (dismissed):
This PR is relevant to `Mininglamp-OSS/octo-android`, but one Space-cache edge case can preserve the cross-Space leakage the PR is meant to fix.

**🔴 Blocking**
- 🔴 Critical: `applySpaceMemberships(emptyList)` incorrectly marks the cache as non-authoritative. In `wkim/src/main/java/com/xinbida/wukongim/manager/ConversationManager.java:735`, a non-null empty `space_memberships` response is treated as "cleared, authoritative=false" at lines 737-741. But `null` already represents the old-backend/no-field case (handled by the early `if (memberships == null) return;`); an empty list from the new field is an authoritative "no memberships" result. Because `wkbase/src/main/java/com/chat/base/space/SpaceFilter.java:466` only disables the `my-row-not-cached-fail-open` path when `isSpaceCacheAuthoritative()` is true, this empty-cache path can still fail-open and show stale/cross-Space group conversations after the server has explicitly returned no memberships — defeating the cross-Space isolation this PR is meant to enforce. Fix: set the empty snapshot's `authoritative` flag to `true` and persist that value; keep only `memberships == null` as the non-authoritative legacy path.

**💬 Non-blocking**
- 🟡 Warning: The diagnostics consent copy in `wkuikit/src/main/java/com/chat/uikit/setting/WKAboutActivity.java:135` does not disclose that exported logs include identifiers such as `channelID`, `fromUid`, `clientMsgNo`, and Space names written at `wkbase/src/main/java/com/chat/base/space/SpaceFilter.java:704`. Note: the prior message-content (`content[0..60]`) capture was the original blocking concern; if it has been removed/redacted, the residual identifier/name disclosure should still be reflected in the consent dialog or redacted before export.
- 🔵 Suggestion: The tests still document "PERSONAL always kept" in `wkbase/src/test/java/com/chat/base/space/SpaceConversationPrunerTest.java:115`, while the production code now prunes non-system PERSONAL conversations by last-message `space_id`. Add tests for `person-space-match`, `person-space-mismatch`, and `person-fail-open-no-msg-space`.

**✅ Highlights**
- The listener registration move in `wkuikit/src/main/java/com/chat/uikit/WKUIKitApplication.java` addresses the cold-start race in the right layer.
- The `WKDiagWriter`/`DiagSink` bridge keeps dependency direction intact.

Verification note: Gradle tests were not run in this review environment (no Java runtime available), so the build/test-pass claim was not independently confirmed.

Jerry-Xin (changes requested):
Re-review at head b3c0f698. The prior blocking privacy issue (raw message-content capture in the diagnostic log) is now RESOLVED — `msgDiagLog` no longer records message body, only `contentLen` + a 32-bit one-way `contentHash`, so the consent dialog no longer needs to disclose content collection. However, a new blocking correctness issue in the Space membership cache can preserve the cross-Space conversation leakage this PR is meant to fix.

🔴 Blocking

🔴 Cross-Space leak: an authoritative empty `space_memberships` response is wrongly flagged non-authoritative. In `wkim/src/main/java/com/xinbida/wukongim/manager/ConversationManager.java:735-741`, `applySpaceMemberships(emptyList)` builds `new SpaceCacheSnapshot(..., false)` (authoritative=false) and the diag line self-documents `authoritative=false`. But `memberships == null` already represents the legacy/no-field case (handled by the early `return` at line 736); a non-null empty list is an authoritative "no memberships" result from the new backend. Because `SpaceFilter.shouldSkipChannelForSpace` only disables the `my-row-not-cached-fail-open` branch when `isSpaceCacheAuthoritative()` is true (`wkbase/src/main/java/com/chat/base/space/SpaceFilter.java:466`), an authoritative-empty snapshot falls into the non-authoritative path (`wkbase/src/main/java/com/chat/base/space/SpaceFilter.java:482-487`) and fails open, so stale/cross-Space group conversations can still be shown after the server explicitly returned no memberships. Fix: mark the empty snapshot authoritative=true and persist that flag; keep only `memberships == null` as the non-authoritative legacy path.

🟡 Non-blocking

🟡 Perf: the new PERSONAL pruning path adds a per-conversation DB read on a UI-thread code path. `SpaceConversationPruner.shouldPrune` now routes PERSONAL channels into `SpaceFilter.shouldSkipChannelForSpace` (`wkbase/src/main/java/com/chat/base/space/SpaceConversationPruner.java:77-78`), whose PERSONAL branch calls `provider.getLastMsgSpaceId(...)` (`wkbase/src/main/java/com/chat/base/space/SpaceFilter.java:414`); the default provider resolves it via `getUIConversationMsg → getWkMsg` lazy DB load (`wkbase/src/main/java/com/chat/base/space/SpaceFilter.java:253-261`). `shouldPrune` is invoked per conversation inside `pruneNonCurrentSpaceGroupsInMemoryOnly()` (`wkuikit/src/main/java/com/chat/uikit/fragment/ChatFragment.java:2242`), which runs in the `sortMsg(...)` first-frame `runOnUIThread` path (`wkuikit/src/main/java/com/chat/uikit/fragment/ChatFragment.java:2123`). First access for each PERSONAL row can hit the DB on the main thread (cached afterward). Please confirm the first-pass cost over a large conversation list is acceptable or move the lookup off the UI thread.

🟡 Tests: the new PERSONAL pruning contract is not directly covered. Tests still document "PERSONAL always kept" and only pass because the stub provider returns no last-message Space ID (`wkbase/src/test/java/com/chat/base/space/SpaceFilterTest.java:206`). Add cases for `person-space-match` (keep), `person-space-mismatch` (skip), `person-fail-open-no-msg-space` (keep), and a system-bot channel (whitelist keep). Also add an empty-membership authoritative case once the 🔴 above is fixed.

🔵 Suggestions

🔵 Some Javadocs still say PERSONAL channels are never channel-level filtered, which is now stale (`wkbase/src/main/java/com/chat/base/space/SpaceFilter.java:339-342`, `wkuikit/src/main/java/com/chat/uikit/fragment/ChatFragment.java:2214-2215`).

✅ Highlights

- Privacy fix landed correctly: diagnostic logging now records `contentLen` + a one-way 32-bit `contentHash` instead of raw `content[0..60]`, so message body is no longer exported (`wkbase/src/main/java/com/chat/base/space/SpaceFilter.java:698-719`).
- Cold-start listener registration is now synchronous before connection sync can fire (`wkuikit/src/main/java/com/chat/uikit/WKUIKitApplication.java`, `wkim/src/main/java/com/xinbida/wukongim/manager/ConversationManager.java`).
- Avatar cache-key priority respects the server-provided `avatar_cache_key` while preserving the local fallback (`wkbase/src/main/java/com/chat/base/common/WKCommonModel.java`).
- PERSONAL cross-Space filtering uses a server-driven SystemBot whitelist (appconfig `system_bot_uids` with a local fallback) rather than caller-supplied ids, and blank `msgSpaceId` fails open for legacy messages — reasonable trust/compat choices.

Verification note: build/tests were not run in this review environment (no Java runtime); findings are from source inspection at head b3c0f698.

yujiawei (approved):
# Code Review — PR #75 (octo-android)

Reviewed at head `103ca1c563f64527edb8b0605c2c0c17f6a391fd` (note: head advanced 2 commits past the originally-dispatched `31189844`; both new commits respond to earlier review feedback — message-body redaction in `msgDiagLog`, and `authoritative=true` on empty `space_memberships` — and are sound). Base merge-base `6d896f19`.

## 1. Scope check

This is a batch fix PR (Closes #77). Every change maps to a stated bullet in the PR description: cold-start listener sync, Space PERSONAL filtering, the diagnostic subsystem, avatar cache-key refresh, recent-tab 3-day filter removal, follow-tab `@` badge takedown, version bump. Nothing lands outside the declared scope. The diagnostic subsystem (DiagSink / WKDiagWriter / SpaceCacheStats) is large but explicitly declared and carries documented removal steps. **Scope: OK.**

All new cross-module symbol references resolve (`WKLogUtils.LOGGABLE`, `AvatarView.clearCache`, `getConvSyncSpaceId`, `getCurrentSpaceId`, color resources `reminderColor`/`colorAccent`/`color999`). No build break.

## 2. Findings (all P2 / advisory — none block merge)

**[P2] `ChatFragment.java:2254` — `pruneWhitelist` uses a provider adapter that doesn't implement `getLastMsgSpaceId()`.**
`SpaceConversationPruner` now prunes `PERSONAL` channels, and the new PERSONAL branch in `SpaceFilter.shouldSkipChannelForSpace` relies on `provider.getLastMsgSpaceId(...)`. The main list prune path uses the production `DEFAULT_PROVIDER` (which overrides it correctly), but `pruneWhitelist(spaceConversationKeys, …, SpaceFilter_DefaultProviderAdapter.INSTANCE)` passes the local adapter, which inherits the interface default `null`. Result: off-space PERSONAL keys hit `person-fail-open-no-msg-space` and are never removed from the whitelist Set. Impact is bounded (stale tracking-key, not a display or data bug — the conversation list itself prunes correctly), but the adapter should override `getLastMsgSpaceId` to delegate to `SpaceFilter`’s static path for parity.

**[P2] PERSONAL filtering has no unit-test coverage for the new behavior.** `SpaceFilterTest` / `SpaceConversationPrunerTest` were not touched; the existing tests still assert the old "PERSONAL always kept / person-pass" semantics (`personPass_privateChat_doesNotSkipAtChannelLevel`, etc.) and `StubProvider` returns `null` for `getLastMsgSpaceId`. The new mismatch/match/fail-open branches are therefore unprotected against regression. Add cases for: system-bot pass, space-match keep, space-mismatch skip, and null-msg fail-open.

**[P2] `DiagnosticLogFile.rotateIfOversized()` does up to ~2.5MB of blocking I/O inside the `synchronized append()`.** This file is shared with `CrashHandler`/`ANRWatchdog`, which append synchronously on the dying thread. If a crash append crosses the 5MB boundary, the rotate copy runs on the crash path and the process may be killed before the crash line lands. Consider rotating off the crash path (e.g. only from the async DiagSink writer, or size-check without inline copy when called from crash handlers).

**[P2] `SpaceFilter.callerOf()` builds a full stack trace per filtered item when diagnostics are enabled.** `new Throwable().getStackTrace()` is expensive on Android. It's correctly gated behind `WKLogUtils.LOGGABLE || DiagSink.isEnabled()`, so it only fires for a user who has manually opted into the 2-hour diagnostic window — acceptable for a debugging tool, but during that window a 100+ item filter pass on the main thread can stutter. Worth a note in the diag-mode UX or a cheaper caller hint.

**[P2] `getLastMsgSpaceId()` → `getUIConversationMsg()` can do a synchronous SQLite read on the UI thread.** When the lazy `wkMsg` cache misses (cold start, many PERSONAL rows), `shouldSkipChannelForSpace` during list rendering will hit the DB sequentially. Mitigated by the lazy cache after first access, but cold-start PERSONAL-heavy lists may jank.

**[P2] `WKAbout` diagnostic `AlertDialog` is not lifecycle-dismissed.** `showDiagDialog()` shows without dismissing in `onDestroy()`; back/rotate before choosing can leak a Window. Minor, but trivial to guard.

**[P2] `DiagnosticLogFile`: `append()` writes UTF-8 but `read()` uses `FileReader` (platform default charset).** On Android the default charset is guaranteed UTF-8, so Chinese in exported logs won't actually garble today; still worth pinning `InputStreamReader(..., StandardCharsets.UTF_8)` to be explicit and future-proof.

**Note on `applySpaceMemberships(emptyList) → authoritative=true` (new commit):** semantically correct for a new backend authoritatively reporting zero memberships, and `null` still preserves the old value for backward compat. One residual risk: a spurious/transient empty response now actively prunes cross-space groups that previously fail-open-displayed. The tradeoff is reasoned in the new javadoc and acceptable; flagging only so it's on record.

## 3. Verdict

No correctness, security, data-loss, or build-breaking issues. The cold-start `onBack(null)` permit-release fix and the synchronous `initIMListener` move are well-reasoned and defensively coded. All findings above are advisory quality/coverage items that can be addressed in follow-ups; the most worthwhile are the `pruneWhitelist` adapter parity and the missing PERSONAL test coverage.

Approving. Please consider the P2s, especially test coverage for the new PERSONAL filter branches.

Jerry-Xin (approved):
Re-review at head 103ca1c5. Both prior blocking items are now resolved — approving.

✅ Resolved blockers
- Privacy (prior 🔴): `msgDiagLog` no longer captures raw message body; it now records only `contentLen` + a one-way `contentHash`, so exported diagnostics no longer leak private-message content (`wkbase/src/main/java/com/chat/base/space/SpaceFilter.java`).
- Cross-Space cache (prior 🔴): `applySpaceMemberships(emptyList)` now builds the empty snapshot with `authoritative=true` and persists it (`wkim/src/main/java/com/xinbida/wukongim/manager/ConversationManager.java:746-754`). A non-null empty list from the new backend is correctly treated as an authoritative "no memberships" result, so `SpaceFilter`'s `my-row-not-cached-fail-open` path no longer shows stale/cross-Space conversations after the server returns no memberships; `memberships == null` remains the only non-authoritative legacy path. The delta is surgical (only this fix) and the diagnostic line now self-documents `authoritative=true`.

CI `check-sprint` is now green (the PR links its issue).

💬 Non-blocking (follow-ups, not gating)
- 🟡 Perf: the new PERSONAL prune path does a per-conversation lazy DB read (`getLastMsgSpaceId` → `getUIConversationMsg`/`getWkMsg`) reachable from the UI-thread `sortMsg`/`pruneNonCurrentSpaceGroupsInMemoryOnly` path; first access per row can hit the DB on the main thread (cached afterward). Worth confirming the first-pass cost over a large conversation list or moving the lookup off the UI thread.
- 🟡 Tests: the new PERSONAL filtering contract (`person-space-match` / `person-space-mismatch` / `person-fail-open-no-msg-space` / system-bot whitelist) plus the empty-membership authoritative case are still uncovered; please add focused tests.
- 🔵 Some Javadocs still describe PERSONAL channels as never channel-level filtered, which is now stale.

✅ Highlights
- Empty-membership authoritative semantics fixed exactly at the boundary that mattered for cross-Space isolation.
- Cold-start listener registration is synchronous before connection sync can fire.
- PERSONAL cross-Space filtering uses a server-driven SystemBot whitelist (appconfig `system_bot_uids`, not caller-supplied ids) and fails open for legacy messages without a space_id.

Verification note: Gradle build/tests were not run in this review environment (no Java runtime); findings are from source inspection at head 103ca1c5.

OctoBoooot (approved):
## Re-review (delta): fix: space 校验 + 冷启动 + UI 优化批次 (#75) @ 103ca1c

**Verdict:** Approve. All three of my prior blockers are resolved at this head, and the cross-Space empty-list blocker (raised by Jerry-Xin) is fixed. Remaining gap is test debt — a follow-up, not a merge blocker.

Delta from 31189844 → 103ca1c: `SpaceFilter.java` (+14/−8), `ConversationManager.java` (+12/−3).

### Prior blocking items → resolved
- ✓ **CI green.** `check-sprint` pass (`Closes`-issue gate satisfied) + `build` pass (3m2s). Red-CI precondition cleared.
- ✓ **Privacy (msgDiagLog content capture).** `SpaceFilter.java:691-716` no longer captures `msg.content.substring(0,60)`. It now logs `contentLen` + `contentHash` (`Integer.toHexString(content.hashCode())`, 32-bit one-way) — enough to distinguish "same fromUid+seq, different body" without writing recoverable message text to the exportable diagnostic file. Redact-by-default; consent dialog no longer needs a message-content disclosure. This was the privacy 🔴 Jerry-Xin, yujiawei, and I each flagged.
- ✓ **Cross-Space empty-list authoritative bug** (`ConversationManager.applySpaceMemberships`, Jerry-Xin's 🔴). `:746` empty-list branch now sets `authoritative=true` (was `false`), so a new-backend authoritative "zero memberships" response stops SpaceFilter's `my-row-not-cached-fail-open` path from showing stale/cross-Space conversations. The `memberships == null` (old backend) case correctly still skips, preserving backward compat. The three-way null / empty / non-empty semantics are now documented at the method.

### Minor (follow-up, non-blocking)
- The membership-refresh parser (`WKIMUtils.isMyUidInGroupMemberMsgExtra`) and the new PERSONAL space-filter branch (`SpaceFilter.shouldSkipChannelForSpace` person path) still ship without unit tests, despite being host-side testable via the `ChannelInfoProvider` seam. Both peer reviewers and the author acknowledge this; recommend a fast-follow test PR rather than holding this one.

### Praise
- The privacy fix is the right shape: rather than dropping the diagnostic field entirely, it kept the discriminating signal (length + non-reversible hash) and removed only the recoverable content. That preserves the debugging value the field existed for while closing the leak.
- Good handling of the empty-list edge: distinguishing "backend doesn't support this field" (null → skip) from "backend authoritatively says zero" (empty → authoritative clear) is exactly the right semantic split, and it's now documented so the next reader doesn't re-introduce the fail-open.

- Carryover perf notes from peers (getLastMsgSpaceId potential UI-thread DB read on cold list; `callerOf()` full-stack cost; `rotateIfOversized` big-file I/O on the synchronous crash-path append) are worth a look but are diagnostic-mode / cold-start paths, not merge blockers. The crash-path rotation I/O is the one I'd prioritize — blocking the crash handler on multi-MB I/O can truncate the crash log it's writing.

*Retires my prior CR at 31189844. Gate at this head: both reviewers who raised blockers (Jerry-Xin, yujiawei) have approved; CI green. needs-human-review remains (19 files) for a maintainer's final merge call.*

PR #83, opened 2026-07-07
feat(channel-search): OpenSearch 频道搜索 + 子区查找入口 + 图片搜索切新接口 (v1.3.5)

### LINKED ISSUE #84: 频道内搜索能力 + Bot 进群成员列表不刷新问题
## 背景

当前 App 缺少完整的**频道内搜索**能力（只能按图片搜索），且**群设置里邀请 Bot 后成员列表不更新**，需要手动退出重进才能看到，体验较差。本 issue 汇总两块相关改动，作为 v1.3.5 的一次性交付。

## 需求 1：频道内搜索（OpenSearch）

进入会话 → 查找聊天记录，提供 4 个 Tab 的搜索结果：

- **全部**：消息 / 媒体 / 文件混合命中，按时间倒序
- **消息**：文本消息命中，头像展示为**发送人**（不是频道头像）
- **媒体**：图片 / 视频宫格，视频需要有 play 角标，格子尺寸与 iOS 对齐
- **文件**：文件列表，点击可跳转原消息

交互要求：

- [ ] 关键词输入实时驱动结果刷新
- [ ] 清空关键词时结果列表也清空
- [ ] 支持分页 / 游标
- [ ] 子区详情页也要有"查找聊天记录"入口，走同一套搜索流程
- [ ] 图片搜索老入口切换到新的 `_search_media` 接口，废弃老接口

## 需求 2：Bot 进群后成员列表不刷新

**复现路径：** 群设置 → 邀请 Bot → 成员列表页无变化，需要退出重进才能看到新加入的 Bot。

**期望：** 邀请成功后成员列表立即出现新 Bot，无需手动刷新。

**方案：** 走 C+（本地兜底刷新）逻辑，避免依赖服务端推送时序。

## 验收

- [ ] 4 个 Tab 结果正确加载 & 分页
- [ ] 媒体 Tab 视频有 play 角标，格子尺寸正确
- [ ] 命中消息头像为发送人
- [ ] 清空关键词 → 结果清空
- [ ] 子区入口可用
- [ ] 邀请 Bot 后成员列表立即刷新
- [ ] 单测覆盖搜索数据层（BodyBuilder / ErrorParser / Outcome / Rfc3339 / MediaHitAdapter）

## 目标版本

v1.3.5 (versionCode 36)

mzanalyticst:
Closes #84

## Summary

- **feat(channel-search)**: 新增 OpenSearch 驱动的频道内搜索数据层 + 4-Tab 搜索结果 UI（全部 / 消息 / 媒体 / 文件），支持关键词输入实时驱动
- **feat(thread)**: 子区详情页新增"查找聊天记录"入口
- **refactor(image-search)**: 老入口切换至 `_search_media` 接口，媒体格子尺寸与视频 play 角标 UI 优化
- **fix(channel-search)**: 搜索结果头像改为发送人；清空关键词时清空浏览
- **fix(group-members)**: bot invite 后 UI 无变化，走 C+ 兜底
- **chore(release)**: 版本号升至 `1.3.5` (versionCode 36)

## 主要改动范围

- 新增 `wkbase/src/main/java/com/chat/base/search/channel/` — OpenSearch DTO / Body Builder / Service / Error Parser / Outcome / RFC3339 工具，含单测
- 新增 `wkuikit/.../chat/search/channel/` — 4 个 Fragment（All / Message / Media / File）+ ViewModel + Adapter
- 改造 `MessageRecordActivity` / `SearchWithImgActivity` 走新接口
- `ThreadDetailActivity` 增加查找聊天记录入口
- 群成员 `GroupModel` bot invite 兜底逻辑

## Test plan

- [ ] 打开会话 → 长按/入口进入查找聊天记录 → 输入关键词，4 个 Tab 结果正确加载 & 分页
- [ ] 媒体 Tab：图片/视频格子尺寸正确，视频有 play 角标
- [ ] 文件 Tab：文件项展示正常，点击可跳转
- [ ] 消息 Tab：命中消息头像为发送人头像，点击定位到原消息
- [ ] 清空关键词输入 → 结果列表清空
- [ ] 子区详情页 → "查找聊天记录" 入口可用，进入后走新搜索流程
- [ ] 群设置邀请 bot → 成员列表立即更新（无需刷新）
- [ ] 单测：ChannelSearchBodyBuilderTest / ChannelSearchErrorParserTest / ChannelSearchOutcomeTest / Rfc3339Test / ChannelMediaHitAdapterTest 全部通过
- [ ] 打 release 包，版本号显示 1.3.5 (36)

### DIFF
--- a/wkbase/src/main/java/com/chat/base/search/channel/ChannelSearchModel.kt
+++ b/wkbase/src/main/java/com/chat/base/search/channel/ChannelSearchModel.kt
@@ -1,0 +1,117 @@
+/*
+ * Copyright 2026-present OctoIM contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.chat.base.search.channel
+
+import com.chat.base.base.WKBaseModel
+import com.chat.base.net.IRequestResultErrorInfoListener
+import com.chat.base.search.channel.dto.AroundRequest
+import com.chat.base.search.channel.dto.AroundResult
+import com.chat.base.search.channel.dto.ChannelSearchReq
+import com.chat.base.search.channel.dto.CombinedHit
+import com.chat.base.search.channel.dto.CursorList
+import com.chat.base.search.channel.dto.FileHit
+import com.chat.base.search.channel.dto.MediaHit
+import com.chat.base.search.channel.dto.MessageHit
+import com.chat.base.search.channel.dto.SearchErrorCode
+import io.reactivex.rxjava3.core.Observable
+
+/**
+ * 频道内搜索数据编排层。负责：
+ *  - 序列化请求体（[ChannelSearchBodyBuilder]）
+ *  - 发起 5 个 `/v1/messages/_search*` 请求
+ *  - 把服务端结构化错误码解析进 [ChannelSearchOutcome]
+ *
+ * 本地兜底不在此层处理。UI 层根据 [uiAction] 判断是否调用 IMSDK
+ * `WKIM.msgManager.searchWithChannel` 取本地结果，并在 [ChannelSearchOutcome.success] 时把
+ * `fromLocalFallback=true` 标记上。
+ *
+ * `X-Space-Id` 由 `CommonRequestParamInterceptor` 自动注入。
+ */
+object ChannelSearchModel : WKBaseModel() {
+
+    private val service by lazy { createService(ChannelSearchService::class.java) }
+
+    fun searchMessages(
+        req: ChannelSearchReq,
+        callback: (ChannelSearchOutcome<CursorList<MessageHit>>) -> Unit,
+    ) {
+        dispatch(service.searchMessages(ChannelSearchBodyBuilder.build(req)), callback)
+    }
+
+    fun searchAll(
+        req: ChannelSearchReq,
+        callback: (ChannelSearchOutcome<CursorList<CombinedHit>>) -> Unit,
+    ) {
+        dispatch(service.searchAll(ChannelSearchBodyBuilder.build(req)), callback)
+    }
+
+    fun searchMedia(
+        req: ChannelSearchReq,
+        callback: (ChannelSearchOutcome<CursorList<MediaHit>>) -> Unit,
+    ) {
+        // 服务端：_search_media keyword 必须为空，否则 400 VALIDATION_FAILED。
+        dispatch(service.searchMedia(ChannelSearchBodyBuilder.build(req, allowKeyword = false)), callback)
+    }
+
+    fun searchFiles(
+        req: ChannelSearchReq,
+        callback: (ChannelSearchOutcome<CursorList<FileHit>>) -> Unit,
+    ) {
+        dispatch(service.searchFiles(ChannelSearchBodyBuilder.build(req)), callback)
+    }
+
+    fun searchAround(
+        req: AroundRequest,
+        callback: (ChannelSearchOutcome<AroundResult>) -> Unit,
+    ) {
+        dispatch(service.searchAround(ChannelSearchBodyBuilder.buildAround(req)), callback)
+    }
+
+    private fun <T : Any> dispatch(observable: Observable<T>, callback: (ChannelSearchOutcome<T>) -> Unit) {
+        requestAndErrorBack(observable, object : IRequestResultErrorInfoListener<T> {
+            override fun onSuccess(result: T) {
+                callback(ChannelSearchOutcome.success(result))
+            }
+
+            override fun onFail(code: Int, msg: String, errJson: String) {
+                callback(mapFailure(code, msg, errJson))
+            }
+        })
+    }
+
+    private fun <T : Any> mapFailure(httpStatus: Int, msg: String, errJson: String): ChannelSearchOutcome<T> {
+        val parsed = ChannelSearchErrorParser.parse(errJson)
+        if (parsed != null) {
+            return ChannelSearchOutcome.failure(
+                httpStatus = httpStatus,
+                errorCode = parsed.code,
+                errorMessage = parsed.message.ifBlank { msg },
+                retryAfterSec = parsed.retryAfterSec,
+            )
+        }
+        // 无结构化 error body：按 HTTP 状态码兜底分类。
+        val fallbackCode = when (httpStatus) {
+            in 500..599 -> SearchErrorCode.UPSTREAM_UNAVAILABLE
+            429 -> SearchErrorCode.RATE_LIMITED
+            404 -> SearchErrorCode.NOT_FOUND
+            400 -> SearchErrorCode.VALIDATION_FAILED
+            0 -> ChannelSearchOutcome.LOCAL_ERROR_NETWORK   // ResponseExceptionHandle 对网络异常通常给非 HTTP code
+            else -> SearchErrorCode.INTERNAL
+        }
+        return ChannelSearchOutcome.failure(httpStatus = httpStatus, errorCode = fallbackCode, errorMessage = msg)
+    }
+}
--- a/wkbase/src/main/java/com/chat/base/search/channel/ChannelSearchModel.kt
+++ b/wkbase/src/main/java/com/chat/base/search/channel/ChannelSearchModel.kt
@@ -104,12 +104,14 @@ object ChannelSearchModel : WKBaseModel() {
             )
         }
         // 无结构化 error body：按 HTTP 状态码兜底分类。
-        val fallbackCode = when (httpStatus) {
-            in 500..599 -> SearchErrorCode.UPSTREAM_UNAVAILABLE
-            429 -> SearchErrorCode.RATE_LIMITED
-            404 -> SearchErrorCode.NOT_FOUND
-            400 -> SearchErrorCode.VALIDATION_FAILED
-            0 -> ChannelSearchOutcome.LOCAL_ERROR_NETWORK   // ResponseExceptionHandle 对网络异常通常给非 HTTP code
+        // 注意：ResponseExceptionHandle 对非 HttpException（连接失败、超时、DNS 等传输层异常）分配 code = -1；
+        // 任何非正数都视作传输层错误，走 LOCAL_ERROR_NETWORK → FALLBACK_TO_LOCAL，触发本地 IMSDK 兜底搜索。
+        val fallbackCode = when {
+            httpStatus in 500..599 -> SearchErrorCode.UPSTREAM_UNAVAILABLE
+            httpStatus == 429 -> SearchErrorCode.RATE_LIMITED
+            httpStatus == 404 -> SearchErrorCode.NOT_FOUND
+            httpStatus == 400 -> SearchErrorCode.VALIDATION_FAILED
+            httpStatus <= 0 -> ChannelSearchOutcome.LOCAL_ERROR_NETWORK
             else -> SearchErrorCode.INTERNAL
         }
         return ChannelSearchOutcome.failure(httpStatus = httpStatus, errorCode = fallbackCode, errorMessage = msg)
--- a/wkbase/src/main/java/com/chat/base/search/channel/Rfc3339.kt
+++ b/wkbase/src/main/java/com/chat/base/search/channel/Rfc3339.kt
@@ -23,11 +23,14 @@ import java.util.TimeZone
 /**
  * RFC3339 → epoch 秒。最小可用，不引入 java.time（项目 minSdk=23 且 coreLibraryDesugaring 未开）。
  *
- * 接受形态：`2026-06-29T10:30:00Z` / `2026-06-29T10:30:00+08:00` / 不带时区。
+ * 接受形态：`2026-06-29T10:30:00Z` / `2026-06-29T10:30:00+08:00` / `2026-06-29T10:30:00.123Z`
+ * / `2026-06-29T10:30:00.123456+08:00` / 不带时区。
+ * 分数秒（`.SSS…`）按 RFC3339 规定为任意精度，我们不需要亚秒精度，直接在归一化阶段剥离。
  * 解析失败返回 0。
  */
 object Rfc3339 {
     private val OFFSET_COLON = Regex("([+-])(\\d{2}):(\\d{2})$")
+    private val FRACTION = Regex("\\.\\d+")
     private val PATTERNS = listOf(
         "yyyy-MM-dd'T'HH:mm:ssZ",
         "yyyy-MM-dd'T'HH:mm:ss",
@@ -35,9 +38,11 @@ object Rfc3339 {
 
     fun toEpochSeconds(s: String?): Long {
         if (s.isNullOrEmpty()) return 0L
-        // SimpleDateFormat 的 Z 模式接受 +0000，不接受 Z 字面量或 +08:00。先标准化。
+        // SimpleDateFormat 的 Z 模式接受 +0000，不接受 Z 字面量或 +08:00；也不接受任意精度分数秒。
+        // 依次剥离：Z → +0000，+08:00 → +0800，.SSS… → 去掉（服务端可能发 1-9 位任意精度）。
         val normalized = s.replace("Z", "+0000")
             .let { OFFSET_COLON.replace(it) { m -> "${m.groupValues[1]}${m.groupValues[2]}${m.groupValues[3]}" } }
+            .let { FRACTION.replace(it, "") }
         for (pattern in PATTERNS) {
             val sdf = SimpleDateFormat(pattern, Locale.US).apply {
                 timeZone = TimeZone.getTimeZone("UTC")
--- a/wkbase/src/main/java/com/chat/base/search/channel/Rfc3339.kt
+++ b/wkbase/src/main/java/com/chat/base/search/channel/Rfc3339.kt
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2026-present OctoIM contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.chat.base.search.channel
+
+import java.text.SimpleDateFormat
+import java.util.Locale
+import java.util.TimeZone
+
+/**
+ * RFC3339 → epoch 秒。最小可用，不引入 java.time（项目 minSdk=23 且 coreLibraryDesugaring 未开）。
+ *
+ * 接受形态：`2026-06-29T10:30:00Z` / `2026-06-29T10:30:00+08:00` / 不带时区。
+ * 解析失败返回 0。
+ */
+object Rfc3339 {
+    private val OFFSET_COLON = Regex("([+-])(\\d{2}):(\\d{2})$")
+    private val PATTERNS = listOf(
+        "yyyy-MM-dd'T'HH:mm:ssZ",
+        "yyyy-MM-dd'T'HH:mm:ss",
+    )
+
+    fun toEpochSeconds(s: String?): Long {
+        if (s.isNullOrEmpty()) return 0L
+        // SimpleDateFormat 的 Z 模式接受 +0000，不接受 Z 字面量或 +08:00。先标准化。
+        val normalized = s.replace("Z", "+0000")
+            .let { OFFSET_COLON.replace(it) { m -> "${m.groupValues[1]}${m.groupValues[2]}${m.groupValues[3]}" } }
+        for (pattern in PATTERNS) {
+            val sdf = SimpleDateFormat(pattern, Locale.US).apply {
+                timeZone = TimeZone.getTimeZone("UTC")
+                isLenient = false
+            }
+            runCatching { sdf.parse(normalized)?.time?.div(1000) }.getOrNull()?.let { return it }
+        }
+        return 0L
+    }
+}
--- a/wkuikit/src/main/java/com/chat/uikit/chat/search/SearchMessageAdapter.kt
+++ b/wkuikit/src/main/java/com/chat/uikit/chat/search/SearchMessageAdapter.kt
@@ -30,6 +30,7 @@ import com.chat.base.utils.WKTimeUtils
 import com.chat.uikit.R
 import com.chat.uikit.chat.provider.WKFileProvider
 import com.chat.uikit.search.remote.GlobalAdapter
+import com.xinbida.wukongim.entity.WKChannelType
 
 class SearchMessageAdapter :
     BaseQuickAdapter<GlobalMessage, BaseViewHolder>(R.layout.item_global_message_layout) {
@@ -39,7 +40,10 @@ class SearchMessageAdapter :
     override fun convert(holder: BaseViewHolder, item: GlobalMessage) {
         val avatarView = holder.getView<AvatarView>(R.id.avatarView)
         avatarView.setSize(40f)
-        avatarView.showAvatar(item.channel.channel_id, item.channel.channel_type)
+        // 频道内搜索结果每行展示发送人头像（与 web / iOS 对齐）。
+        // 原本展示 channel 头像在群里是群头像、在子区里无头像，所有行视觉一致但信息量低；
+        // 改为发送人头像后每行明确"谁发的"，子区也不会再出现空头像。
+        avatarView.showAvatar(item.from_uid, WKChannelType.PERSONAL)
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
             holder.setText(
                 R.id.nameTv,
--- a/wkuikit/src/main/java/com/chat/uikit/chat/search/channel/BaseChannelSearchFragment.kt
+++ b/wkuikit/src/main/java/com/chat/uikit/chat/search/channel/BaseChannelSearchFragment.kt
@@ -85,12 +85,13 @@ abstract class BaseChannelSearchFragment : Fragment() {
 
     /** 默认不兜底；消息相关 tab 覆写。 */
     protected open fun executeLocalFallback(keyword: String) {
+        val b = _binding ?: return
         showOfflineBanner(true)
         resetList()
         showEmpty(getString(emptyResultHintRes))
         hasMore = false
         nextCursor = null
-        binding.refreshLayout.setEnableLoadMore(false)
+        b.refreshLayout.setEnableLoadMore(false)
     }
 
     override fun onCreateView(
@@ -158,21 +159,27 @@ abstract class BaseChannelSearchFragment : Fragment() {
         executeSearch(keyword, nextCursor, isReset = false)
     }
 
-    /** 子类在拿到 outcome 时调用，处理通用错误 + 本地兜底；返回 true 表示已处理。 */
+    /**
+     * 子类在拿到 outcome 时调用，处理通用错误 + 本地兜底；返回 true 表示已处理。
+     *
+     * view 已销毁（`_binding == null`）时静默返回 true——异步回调完成后 UI 已不存在，
+     * 任何 toast / binding 访问都是无意义的且会 NPE。
+     */
     protected fun handleNonSuccess(
         outcome: ChannelSearchOutcome<*>,
         isReset: Boolean,
         keyword: String,
     ): Boolean {
-        binding.refreshLayout.finishRefresh()
-        binding.refreshLayout.finishLoadMore()
+        val b = _binding ?: return true
+        b.refreshLayout.finishRefresh()
+        b.refreshLayout.finishLoadMore()
         return when (outcome.uiAction()) {
             ChannelSearchUiAction.FALLBACK_TO_LOCAL -> {
                 if (isReset) {
                     executeLocalFallback(keyword)
                 } else {
                     hasMore = false
-                    binding.refreshLayout.setEnableLoadMore(false)
+                    b.refreshLayout.setEnableLoadMore(false)
                 }
                 true
             }
@@ -196,13 +203,17 @@ abstract class BaseChannelSearchFragment : Fragment() {
         }
     }
 
-    /** 子类拿到 success 后调用，统一处理 cursor + loadMore 开关 + 空态。 */
+    /**
+     * 子类拿到 success 后调用，统一处理 cursor + loadMore 开关 + 空态。
+     * view 已销毁时静默返回，避免 NPE。
+     */
     protected fun updatePaginationState(hasMore: Boolean, nextCursor: String, isEmpty: Boolean, isReset: Boolean) {
-        binding.refreshLayout.finishRefresh()
-        binding.refreshLayout.finishLoadMore()
+        val b = _binding ?: return
+        b.refreshLayout.finishRefresh()
+        b.refreshLayout.finishLoadMore()
         this.hasMore = hasMore
         this.nextCursor = nextCursor.takeIf { it.isNotEmpty() }
-        binding.refreshLayout.setEnableLoadMore(hasMore && !this.nextCursor.isNullOrEmpty())
+        b.refreshLayout.setEnableLoadMore(hasMore && !this.nextCursor.isNullOrEmpty())
         if (isReset && isEmpty) {
             showEmpty(getString(emptyResultHintRes))
         } else if (isReset) {
@@ -211,15 +222,16 @@ abstract class BaseChannelSearchFragment : Fragment() {
     }
 
     protected fun showOfflineBanner(visible: Boolean) {
-        binding.offlineBannerTv.visibility = if (visible) View.VISIBLE else View.GONE
+        _binding?.offlineBannerTv?.visibility = if (visible) View.VISIBLE else View.GONE
     }
 
     protected fun showEmpty(text: String?) {
+        val b = _binding ?: return
         if (text.isNullOrEmpty()) {
-            binding.emptyTv.visibility = View.GONE
+            b.emptyTv.visibility = View.GONE
         } else {
-            binding.emptyTv.text = text
-            binding.emptyTv.visibility = View.VISIBLE
+            b.emptyTv.text = text
+            b.emptyTv.visibility = View.VISIBLE
         }
     }
 
--- a/wkuikit/src/main/java/com/chat/uikit/chat/search/channel/BaseChannelSearchFragment.kt
+++ b/wkuikit/src/main/java/com/chat/uikit/chat/search/channel/BaseChannelSearchFragment.kt
@@ -1,0 +1,240 @@
+/*
+ * Copyright 2026-present OctoIM contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.chat.uikit.chat.search.channel
+
+import android.os.Bundle
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import androidx.fragment.app.Fragment
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.ViewModelProvider
+import androidx.lifecycle.lifecycleScope
+import androidx.lifecycle.repeatOnLifecycle
+import androidx.recyclerview.widget.RecyclerView
+import com.chat.base.search.channel.ChannelSearchOutcome
+import com.chat.base.search.channel.ChannelSearchUiAction
+import com.chat.base.search.channel.uiAction
+import com.chat.base.utils.WKToastUtils
+import com.chat.uikit.R
+import com.chat.uikit.databinding.FragChannelSearchListBinding
+import com.scwang.smart.refresh.layout.api.RefreshLayout
+import com.scwang.smart.refresh.layout.listener.OnRefreshLoadMoreListener
+import com.xinbida.wukongim.entity.WKChannelType
+import kotlinx.coroutines.launch
+
+/**
+ * 频道内搜索 Fragment 通用骨架。各 tab 复用 [FragChannelSearchListBinding]
+ * (refreshLayout + recyclerView + offline banner + empty state)。
+ *
+ * 子类需实现：
+ *  - [setupRecyclerView]：layoutManager + adapter
+ *  - [resetList] / [appendList]：把服务端返回的数据落进自己的 adapter
+ *  - [executeSearch]：发起一次服务端请求；reset=true 表示新关键词，false 表示翻页
+ *  - [executeLocalFallback]：服务端 503 时的本地兜底（仅消息/全部 tab 实现）
+ *
+ * Activity 通过 [ChannelSearchViewModel.queryEvents] 通知所有 Fragment 重置查询。
+ */
+abstract class BaseChannelSearchFragment : Fragment() {
+
+    protected var _binding: FragChannelSearchListBinding? = null
+    protected val binding get() = _binding!!
+
+    protected lateinit var channelID: String
+    protected var channelType: Byte = WKChannelType.PERSONAL
+
+    protected val viewModel: ChannelSearchViewModel by lazy {
+        ViewModelProvider(requireActivity())[ChannelSearchViewModel::class.java]
+    }
+
+    protected var nextCursor: String? = null
+    protected var hasMore: Boolean = true
+    protected var requestSeq: Long = 0L
+
+    /** 关键词为空时的引导文案。子类可覆写。 */
+    protected open val emptyKeywordHintRes: Int = R.string.channel_search_empty_keyword
+
+    /** 命中关键词但服务端返回空时的文案。 */
+    protected open val emptyResultHintRes: Int = R.string.nodata
+
+    protected abstract fun setupRecyclerView(recyclerView: RecyclerView)
+
+    protected abstract fun resetList()
+
+    protected abstract fun executeSearch(keyword: String, cursor: String?, isReset: Boolean)
+
+    /** 默认不兜底；消息相关 tab 覆写。 */
+    protected open fun executeLocalFallback(keyword: String) {
+        showOfflineBanner(true)
+        resetList()
+        showEmpty(getString(emptyResultHintRes))
+        hasMore = false
+        nextCursor = null
+        binding.refreshLayout.setEnableLoadMore(false)
+    }
+
+    override fun onCreateView(
+        inflater: LayoutInflater,
+        container: ViewGroup?,
+        savedInstanceState: Bundle?
+    ): View {
+        _binding = FragChannelSearchListBinding.inflate(inflater, container, false)
+        return binding.root
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        channelID = requireArguments().getString(ARG_CHANNEL_ID).orEmpty()
+        channelType = requireArguments().getByte(ARG_CHANNEL_TYPE, WKChannelType.PERSONAL)
+
+        setupRecyclerView(binding.recyclerView)
+
+        binding.refreshLayout.setEnableRefresh(false)
+        binding.refreshLayout.setEnableLoadMore(false)
+        binding.refreshLayout.setOnRefreshLoadMoreListener(object : OnRefreshLoadMoreListener {
+            override fun onLoadMore(refreshLayout: RefreshLayout) = loadMore()
+            override fun onRefresh(refreshLayout: RefreshLayout) {}
+        })
+
+        observeQueryEvents()
+    }
+
+    private fun observeQueryEvents() {
+        viewLifecycleOwner.lifecycleScope.launch {
+            viewLifecycleOwner.repeatOnLifecycle(Lifecycle.State.STARTED) {
+                viewModel.queryEvents.collect { state ->
+                    triggerReset(state.keyword)
+                }
+            }
+        }
+    }
+
+    private fun triggerReset(keyword: String) {
+        nextCursor = null
+        hasMore = true
+        showOfflineBanner(false)
+        if (keyword.isEmpty()) {
+            // 清空关键词 == 完全重置：列表清空、cursor 归零、loadMore 关闭、回到引导文案。
+            // 与 Activity 顶部"清空回到快捷过滤入口"的语义保持一致；
+            // 即使 [supportsBrowseWithoutKeyword]=true（媒体 / 文件 tab）也不做无关键词浏览，
+            // 避免清空后台仍在拉取造成视觉跳动。
+            resetList()
+            showEmpty(getString(emptyKeywordHintRes))
+            binding.refreshLayout.setEnableLoadMore(false)
+            return
+        }
+        showEmpty(null)
+        binding.refreshLayout.setEnableLoadMore(true)
+        requestSeq++
+        executeSearch(keyword, null, isReset = true)
+    }
+
+    private fun loadMore() {
+        if (!hasMore || nextCursor.isNullOrEmpty()) {
+            binding.refreshLayout.finishLoadMore()
+            binding.refreshLayout.setEnableLoadMore(false)
+            return
+        }
+        val keyword = viewModel.state.value.keyword
+        requestSeq++
+        executeSearch(keyword, nextCursor, isReset = false)
+    }
+
+    /** 子类在拿到 outcome 时调用，处理通用错误 + 本地兜底；返回 true 表示已处理。 */
+    protected fun handleNonSuccess(
+        outcome: ChannelSearchOutcome<*>,
+        isReset: Boolean,
+        keyword: String,
+    ): Boolean {
+        binding.refreshLayout.finishRefresh()
+        binding.refreshLayout.finishLoadMore()
+        return when (outcome.uiAction()) {
+            ChannelSearchUiAction.FALLBACK_TO_LOCAL -> {
+                if (isReset) {
+                    executeLocalFallback(keyword)
+                } else {
+                    hasMore = false
+                    binding.refreshLayout.setEnableLoadMore(false)
+                }
+                true
+            }
+            ChannelSearchUiAction.RATE_LIMITED -> {
+                toast(R.string.channel_search_rate_limited)
+                true
+            }
+            ChannelSearchUiAction.FEATURE_DISABLED -> {
+                toast(R.string.channel_search_disabled)
+                true
+            }
+            ChannelSearchUiAction.BLOCK_NOT_FOUND -> {
+                toast(R.string.channel_search_not_found)
+                true
+            }
+            ChannelSearchUiAction.VALIDATION_ERROR,
+            ChannelSearchUiAction.GENERIC_ERROR -> {
+                toast(R.string.channel_search_generic_error)
+                true
+            }
+        }
+    }
+
+    /** 子类拿到 success 后调用，统一处理 cursor + loadMore 开关 + 空态。 */
+    protected fun updatePaginationState(hasMore: Boolean, nextCursor: String, isEmpty: Boolean, isReset: Boolean) {
+        binding.refreshLayout.finishRefresh()
+        binding.refreshLayout.finishLoadMore()
+        this.hasMore = hasMore
+        this.nextCursor = nextCursor.takeIf { it.isNotEmpty() }
+        binding.refreshLayout.setEnableLoadMore(hasMore && !this.nextCursor.isNullOrEmpty())
+        if (isReset && isEmpty) {
+            showEmpty(getString(emptyResultHintRes))
+        } else if (isReset) {
+            showEmpty(null)
+        }
+    }
+
+    protected fun showOfflineBanner(visible: Boolean) {
+        binding.offlineBannerTv.visibility = if (visible) View.VISIBLE else View.GONE
+    }
+
+    protected fun showEmpty(text: String?) {
+        if (text.isNullOrEmpty()) {
+            binding.emptyTv.visibility = View.GONE
+        } else {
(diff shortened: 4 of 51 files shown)


Jerry-Xin (changes requested):
The PR is relevant to `octo-android`, but there is a blocking error-handling regression in the new channel search fallback path.

**🔴 Blocking**
- 🔴 Critical: Network failures will not trigger the intended local search fallback. `ResponseExceptionHandle` assigns non-HTTP/network exceptions code `-1`, but `ChannelSearchModel.mapFailure()` only maps code `0` to `LOCAL_ERROR_NETWORK`; `-1` falls through to `INTERNAL`, so `uiAction()` becomes `GENERIC_ERROR` instead of `FALLBACK_TO_LOCAL`. This breaks offline/timeout fallback for the Messages tab and any code relying on the documented network fallback behavior. Fix by mapping `-1` or all non-positive transport errors to `LOCAL_ERROR_NETWORK`.
  - [ResponseExceptionHandle.java](/Users/mlamp/.openclaw/tmp/pr-review-krneg_5n/repo/wkbase/src/main/java/com/chat/base/net/ResponseExceptionHandle.java:110)
  - [ChannelSearchModel.kt](/Users/mlamp/.openclaw/tmp/pr-review-krneg_5n/repo/wkbase/src/main/java/com/chat/base/search/channel/ChannelSearchModel.kt:107)

**💬 Non-blocking**
- 🟡 Warning: Structured search error bodies are only preserved for HTTP 400 today. For 429/404/503, `ResponseExceptionHandle` mostly discards `errorBody`, so `ChannelSearchErrorParser` cannot see `disabled`, `retry_after`, or other server-specific codes. The UI will lose the feature-disabled path and rate-limit retry metadata. Consider preserving `errJson` for all HTTP error statuses.
  - [ResponseExceptionHandle.java](/Users/mlamp/.openclaw/tmp/pr-review-krneg_5n/repo/wkbase/src/main/java/com/chat/base/net/ResponseExceptionHandle.java:84)
  - [ChannelSearchModel.kt](/Users/mlamp/.openclaw/tmp/pr-review-krneg_5n/repo/wkbase/src/main/java/com/chat/base/search/channel/ChannelSearchModel.kt:96)

- 🟡 Warning: `Rfc3339` does not parse fractional seconds, which are common in RFC3339/RFC3339Nano server timestamps. If `sent_at` arrives as `2026-06-29T10:30:00.123Z`, timestamps become `0`, causing blank or incorrect times in search results.
  - [Rfc3339.kt](/Users/mlamp/.openclaw/tmp/pr-review-krneg_5n/repo/wkbase/src/main/java/com/chat/base/search/channel/Rfc3339.kt:31)

**✅ Highlights**
- The new DTO/body-builder layer is well scoped and has useful unit coverage.
- The UI split into tab fragments keeps paging logic relatively contained.
- The bot invite member-sync fallback is conservative and avoids unconditional full syncs.

I could not run Gradle tests because this environment has no Java runtime installed.

wkuikit/src/main/java/com/chat/uikit/group/service/GroupModel.java
```diff
@@ -237,6 +237,77 @@ public void onFail(int code, String msg) {
...
+     */
+    public void groupMembersSyncAndVerify(String groupNo, List<String> expectedUids) {
+        groupMembersSync(groupNo, (code, msg) -> {
+            if (code != HttpResponseCode.success) return;
+            boolean anyMissing;
+            if (expectedUids == null) {
+                anyMissing = true; // 未知邀请对象, 保守走全量
+            } else {
+                anyMissing = false;
+                for (String uid : expectedUids) {
```
OctoBoooot, line 289: **[major]** The bot-invite fallback (`groupMembersSyncAndVerify` + `forceFullMembersSync`) ships with no test, though `GroupModelTest.kt` is the intended home and the PR test-plan lists live member refresh. The null-vs-empty-vs-missing-uid branching here is exactly the kind of pure logic worth a JVM unit test (mock `getMember`, assert `forceFullMembersSync` fires only when a uid is absent/deleted). Not a blocker, but this is the highest-value thing to add before merge.

wkuikit/src/main/java/com/chat/uikit/chat/search/image/SearchWithImgActivity.java
```diff
@@ -230,155 +238,175 @@ private void collect(WKMsg msg) {
...
+        return entity;
+    }
+
+    private WKMessageContent buildContent(MediaHit hit) {
+        if (hit.isVideo()) {
+            WKVideoContent video = new WKVideoContent();
+            video.cover = hit.getThumb_url();
+            video.width = hit.getWidth();
+            video.height = hit.getHeight();
+            video.second = hit.getDuration_ms() / 1000;
```
OctoBoooot, line 395: nit: `hit.getDuration_ms() / 1000` truncates integer seconds — a 1500ms clip shows 1s, a sub-1000ms clip shows 0s. Cosmetic; consider rounding.

wkuikit/src/main/java/com/chat/uikit/chat/search/image/SearchWithImgActivity.java
```diff
@@ -230,155 +238,175 @@ private void collect(WKMsg msg) {
...
+                for (MediaHit hit : hits) {
+                    SearchImgEntity entity = toEntity(hit);
+                    if (entity == null) continue;
+                    addDateHeaderIfNeeded(page, entity.date);
+                    page.add(entity);
+                }
+            }
+
+            hasMore = outcome.getData().getPagination().getHas_more();
+            nextCursor = hasMore ? outcome.getData().getPagination().getNext_cursor() : null;
```
OctoBoooot, line 338: **[minor]** `next_cursor` defaults to `""` (non-null). If the server ever returns `has_more=true` with a blank `next_cursor`, the next `getData()` sends `cursor=""`, is treated as a non-reset load-more, and appends a duplicated first page. Suggest treating a blank cursor as end-of-list (`hasMore = has_more && !next_cursor.isBlank()`).

wkuikit/src/main/java/com/chat/uikit/chat/search/image/SearchWithImgActivity.java
```diff
@@ -230,155 +238,175 @@ private void collect(WKMsg msg) {
...
-                    String date = WKTimeUtils.getInstance().time2YearMonth(msg.getTimestamp() * 1000);
-                    addDateHeaderIfNeeded(merged, date);
-
-                    SearchImgEntity entity = new SearchImgEntity();
-                    entity.date = date;
-                    entity.message = msg;
-                    entity.url = showUrl;
-                    merged.add(entity);
+            if (!outcome.getOk() || outcome.getData() == null) {
+                if (isReset && adapter.getData().isEmpty()) {
```
OctoBoooot, line 319: **[minor]** On a load-more failure or an empty non-reset page, `nodataTv`/`hasMore` are left untouched, so the list silently stops growing with no "no more data" footer unless the server also returns `has_more=false`. Fine if the server contract guarantees that; otherwise consider finishing load-more with no-more-data on the error branch too.

wkuikit/src/main/java/com/chat/uikit/chat/search/image/SearchWithImgActivity.java
```diff
@@ -230,155 +238,175 @@ private void collect(WKMsg msg) {
...
+    }
+
+    private WKMessageContent buildContent(MediaHit hit) {
+        if (hit.isVideo()) {
+            WKVideoContent video = new WKVideoContent();
+            video.cover = hit.getThumb_url();
+            video.width = hit.getWidth();
+            video.height = hit.getHeight();
+            video.second = hit.getDuration_ms() / 1000;
+            // url 字段对视频指向视频文件本身；MediaHit 仅返回封面缩略图，转发时若需要原视频
```
OctoBoooot, line 396: **[minor]** Synthesized `WKVideoContent.url` is set to the thumbnail URL. `forward()` correctly guards videos (re-fetch local msg or toast), so no user-visible bug today — but any future consumer of a video `originalContent` would get a cover URL where a video URL is expected. Consider leaving `url` empty for synthesized video content so misuse fails loudly.

wkuikit/src/main/java/com/chat/uikit/chat/search/SearchMessageAdapter.kt
```diff
@@ -39,7 +40,10 @@ class SearchMessageAdapter :
     override fun convert(holder: BaseViewHolder, item: GlobalMessage) {
         val avatarView = holder.getView<AvatarView>(R.id.avatarView)
         avatarView.setSize(40f)
-        avatarView.showAvatar(item.channel.channel_id, item.channel.channel_type)
+        // 频道内搜索结果每行展示发送人头像（与 web / iOS 对齐）。
```
OctoBoooot, line 43: praise: Switching the row avatar from `channel` to `from_uid` genuinely fixes the empty-avatar-in-thread / group-avatar-per-row issue and aligns with web/iOS. Clear comment explaining the why.

OctoBoooot (commented):
## Review: feat(channel-search): OpenSearch 频道搜索 + 图片搜索切新接口 (#83)

**Verdict:** Approve with comments (scoped to the 8 files I was asked to review, not the whole PR)

Reviewed only the modified files in scope: `MessageRecordActivity.kt`, `SearchWithImgActivity.java` (+entity/adapter), `SearchMessageAdapter.kt`, `GroupModel.java`, `WKIMUtils.java`, `ThreadDetailActivity.java`, `ChooseContactsActivity.java`. Verified the new interfaces/DTOs/SDK methods they call actually exist at this SHA. The refactor and the bot-invite fallback are both fundamentally sound; findings below are edge-cases, not correctness blockers.

CI: only `check-sprint` is red — that's a governance check (maintainer Sprint field), not code CI, so it doesn't bar approval. `mergeStateStatus` stays BLOCKED until it's cleared.

### Verified correct (adversarial checks)
- **Bot-invite fallback actually refreshes the list.** `groupMembersSyncAndVerify` → incremental sync → verify uids → `forceFullMembersSync(v=0)` → `save()` fires `onAddMembers`, and `GroupDetailActivity` subscribes via `addOnAddChannelMemberListener`, so a live-open group settings grid updates without reopen. `v=0` (`WHERE version>0`) does pull the bot's low-version row. No infinite loop: full sync recurses with a **null** listener, so verification runs at most once.
- **Refactor parity:** the -150 deleted lines in `MessageRecordActivity` (local+remote merge, dedup, click-to-jump) moved into the new 4 fragments / ViewModel, not dropped. `initialKeyword` path works: `setText` → TextWatcher → `showResultState` → pager attach → `setKeyword`; `queryEvents` has `replay=1` + fragments `replayCurrent()` on bind.
- All new symbols exist: `getWithMessageID`, `message_idstr`, `WKVideoContent.{cover,width,height,second,url}`, `allowKeyword=false` strips keyword (has a unit test), and `uikit_video_forward_unavailable` + `channel_search_tab_*` are in both `values/` and `values-en/`.

### Major
- `GroupModel.java` / `WKIMUtils.java` / `ChooseContactsActivity.java` — the whole bot-invite fallback (3 changed files, plus `resolveSyncVersion`/`forceFullMembersSync`) ships with **no test**, despite `GroupModelTest.kt` being the intended home and the PR body listing member-refresh in the test plan. The verify-then-fallback branching (null vs empty vs missing uid) is exactly the logic worth a JVM unit test.

### Minor
- `SearchWithImgActivity.java:395` — `video.second = hit.getDuration_ms() / 1000` truncates; a 1500ms clip shows `1s`, a 900ms clip shows `0s`. Cosmetic but visible. Consider rounding.
- `SearchWithImgActivity.java:319-321,343` — on a **load-more** failure or empty page (`isReset==false`), `nodataTv` is never touched and `hasMore` stays true, so the list silently stops growing with no "no more data" affordance unless the server also returns `has_more=false`. Trusts the server contract; fine if guaranteed.
- `SearchWithImgActivity.java:338` — `nextCursor = pagination.next_cursor` where `next_cursor` defaults to `""` (non-null). If server ever returns `has_more=true` with an empty `next_cursor`, the next request sends `cursor=""` treated as `isReset==false` → duplicate first page appended. Guard: treat blank cursor as end-of-list.
- `SearchWithImgActivity.java:389-397 buildContent` — synthesized `WKVideoContent.url` is set to the thumbnail URL. Forward already guards videos (re-fetches local msg or toasts), but any other consumer of `originalContent` on a video entity would get a cover-as-video URL. Contained today; brittle if reused.

### Questions
- `MediaHit` fields `message_id`/`media_kind`/`sender_id`/`sent_at`/`month_bucket` are Kotlin `lateinit`/non-null but populated by Gson reflection. Java callers guard with `!= null` / `TextUtils.isEmpty`, so no NPE today — but is Gson the deserializer here, and is there any Kotlin caller that would dereference these non-null and NPE on an omitted field? (Out of my file scope; flagging because the media path constructs from it.)

### Praise
- `SearchMessageAdapter.kt:43` — switching the row avatar from channel to `from_uid` genuinely fixes the empty-avatar-in-thread case; good cross-platform alignment call, well-commented.
- The verify-before-fallback design in `groupMembersSyncAndVerify` (only pay for a full sync when a uid is actually missing) is the right cost trade-off, and reusing it for the "duplicate invite → silent success, no 1002" case in `ChooseContactsActivity` is a nice catch beyond the literal bug.

- The new `channel/` fragments, ViewModel, adapters, and `wkbase/.../search/channel/**` data layer were not reviewed here (not in the requested file set) — the parity of the deleted `MessageRecordActivity` logic depends on them.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review: feat(channel-search): OpenSearch 频道搜索 + 子区查找入口 + 图片搜索切新接口 (v1.3.5)

**PR:** Mininglamp-OSS/octo-android #83
**Head SHA:** `3e889b40ca18d983bb4a918eef0d00babadda7b3`
**Merge-base:** `9b5990f9e91f62b55ba4d78076cfbb47d393aa04`
**Files changed:** 51
**Scope:** New OpenSearch-backed channel search (service + DTO + ViewModel + 4-tab Fragments + adapters), image search refactored from local DB to `_search_media` API, group member sync hardening, thread "find content" entry, version bump 1.3.4 → 1.3.5.

---

## Summary

This PR introduces a comprehensive channel search feature powered by OpenSearch, replacing local DB queries for message/media/file search across the Android client. The architecture is clean: a singleton `ChannelSearchModel` orchestrates Retrofit calls, a shared `ChannelSearchViewModel` coordinates keyword/filter state across 4 tab Fragments via SharedFlow, and adapters render highlighted HTML snippets from the API.

Additionally, the PR hardens group member sync with a verify-and-fallback pattern (incremental sync + local count check → full sync fallback), adds a "find content" entry from thread detail, and bumps to v1.3.5.

**Overall assessment:** Well-structured feature PR. Two P2 issues found in error-handling paths; no P0/P1 blockers.

---

## Verification

### ✅ Service Layer (`ChannelSearchService.kt`, `ChannelSearchModel.kt`)
- 5 Retrofit POST endpoints under `/v1/messages/_search*` correctly defined with `@Body JSONObject`.
- `requestAndErrorBack` wrapper properly maps HTTP failures to `ChannelSearchOutcome` with error codes.
- `ChannelSearchErrorParser` correctly extracts structured error envelope `{ "error": { "code", "message", "details" } }` with `retry_after` from details map.
- `ChannelSearchModel.searchMedia()` correctly forces `allowKeyword=false`, preventing keyword injection into media queries.

### ✅ Data Flow: Keyword → API Request
- `MessageRecordActivity` TextWatcher → `viewModel.setKeyword()` → `MutableSharedFlow<State>(replay=1)` with 300ms debounce → all 4 Fragments collect via `repeatOnLifecycle(STARTED)` → `triggerReset()` clears pagination and calls `executeSearch()`.
- Verified: keyword flows correctly from UI to ViewModel to API body via `ChannelSearchBodyBuilder`.

### ✅ Data Flow: Cursor Pagination
- API response `CursorList<T>.pagination.next_cursor` → Fragment `nextCursor` field → next `ChannelSearchReq.cursor` → API.
- `hasMore` flag correctly gates `finishLoadMoreWithNoMoreData()` on SmartRefreshLayout.
- `requestSeq` monotonically increases to discard stale responses (race condition protection).

### ✅ ViewModel State Management
- `State(keyword, filters, tabIndex)` with `MutableSharedFlow` for query events ensures all Fragments react to keyword changes.
- `offscreenPageLimit = fragments.size` (4) keeps all tabs alive — acceptable for search UX.

### ✅ Group Member Sync Hardening (`GroupModel.java`)
- `groupMembersSyncAndVerify()`: incremental sync → verify local count ≥ expected → fallback to `forceFullMembersSync()` (v=0).
- Data flow verified: 1002 message handler → `groupMembersSyncAndVerify()` → `doMembersSyncWithVersion()` → `GroupService.getGroupMembers()` → local DB upsert.
- `parseGroupMemberAddExtraUids()` correctly extracts invited UIDs from message extra JSON for targeted sync.

### ✅ RFC3339 Parser (`Rfc3339.kt`)
- Normalizes `Z` → `+0000`, `+08:00` → `+0800` for `SimpleDateFormat` compatibility (minSdk=23, no `java.time`).
- `isLenient=false` prevents silent date overflow. Returns 0L on parse failure — safe fallback.

### ✅ Adapter Layer
- `ChannelCombinedHitAdapter`: multi-type (message + file) with HTML `<mark>` → `<font color=#7761F4>` conversion for highlighted snippets.
- `ChannelMediaHitAdapter`: month-bucket headers via `Entry.Header` insertion on `month_bucket` change. Cell sizing via `cellSize` parameter.
- `ChannelFileHitAdapter`: file icon via `WKFileProvider`, HTML highlighting for file names.

### ✅ Image Search Refactor (`SearchWithImgActivity.java`)
- Correctly migrated from local DB `msgManager.searchWithChannel()` to `_search_media` API.
- `toEntity()` properly builds `SearchImgEntity` with `GlobalMessage` metadata and `WKImageContent`/`WKVideoContent`.
- `collect()`/`collectFromEntity()` handle favorites with local DB fallback for `WKMsg` lookup.
- Video forward path correctly falls back to local DB for full video URL (MediaHit only has `thumb_url`).

### ✅ Thread "Find Content" Entry (`ThreadDetailActivity.java`)
- Click handler launches `MessageRecordActivity` with `WKChannelType.COMMUNITY_TOPIC` and compound `channelId` format `{groupNo}____{shortId}`.
- Verified: compound channelId correctly parsed by downstream search API.

---

## Findings

### Finding 1: P2 — Error path leaves blank screen on initial search failure

**File:** `wkuikit/src/main/java/com/chat/uikit/chat/search/channel/BaseChannelSearchFragment.kt`
**Line:** 144
**Diff-scope:** New (introduced in this PR)

**Description:**
`triggerReset()` calls `showEmpty(null)` before `executeSearch()`. If the initial API call fails with RATE_LIMITED / FEATURE_DISABLED / NOT_FOUND / VALIDATION_ERROR / GENERIC_ERROR, `handleNonSuccess()` shows a toast but never restores the empty-state text. The user sees a blank screen with only a toast.

**Data flow trace:**
```
triggerReset() [line 144]
  → showEmpty(null)  // sets emptyTv text to null, visibility GONE
  → executeSearch()
    → API callback: outcome.ok=false
      → handleNonSuccess(outcome)
        → shows toast for RATE_LIMITED/FEATURE_DISABLED/etc.
        → does NOT call showEmpty() with error message
        → emptyTv remains GONE, recyclerView remains empty
```

**Severity:** P2 (UX degradation, not data loss or crash). The toast provides feedback, but the blank screen is confusing.

**Suggestion:** In `handleNonSuccess()`, after showing the toast, call `showEmpty()` with an appropriate error message string (e.g., "Search temporarily unavailable") so the empty state is restored.

---

### Finding 2: P2 — Load-more failure path silently swallows errors

**File:** `wkuikit/src/main/java/com/chat/uikit/chat/search/image/SearchWithImgActivity.java`
**Line:** 273
**Diff-scope:** New (introduced in this PR — refactored from local DB to API)

**Description:**
In `getData()`, when a non-initial page (load-more) fails (`!outcome.getOk()`), the callback resets `loading=false` and calls `finishLoadMore()` but does NOT set `hasMore=false` and does NOT show a toast/error message. The user can repeatedly scroll to trigger infinite silent failing requests.

**Data flow trace:**
```
getData() [line 248]
  → isReset = (nextCursor == null)  // false for load-more
  → API callback: outcome.ok=false [line 268]
    → if (isReset && adapter.getData().isEmpty()) { ... }  // skipped for load-more
    → return null  // no toast, no hasMore=false
  → loading=false, finishLoadMore() called
  → hasMore still true, nextCursor unchanged
  → user scrolls → onLoadMore() → getData() → same failure repeats
```

**Severity:** P2 (silent failure loop, not crash or data corruption). Network waste and poor UX.

**Suggestion:** On load-more failure, either: (a) set `hasMore=false` + `finishLoadMoreWithNoMoreData()` to stop further requests, or (b) show a toast and keep `hasMore=true` so the user can retry by scrolling. Option (a) is safer to prevent infinite loops.

---

## Blind Spot Checklist (C1–C6)

**C1 — Dual-path parity:** N/A. No symmetric add/remove or subscribe/unsubscribe paths in this PR.

**C2 — Control-flow ordering / nested reuse:** Clear. `BaseChannelSearchFragment` is reused across 4 tab Fragments; verified `triggerReset()` and `loadMore()` are not double-called. `requestSeq` monotonic counter prevents stale response processing.

**C3 — Authorization boundary:** Clear. All 5 search endpoints are behind the same auth token as existing chat APIs. No new endpoints exposed without auth.

**C4 — Authorization lifecycle:** N/A. No container/member status cascade relevant to search.

**C5 — Build/note through ≠ runtime correct:** Clear. Verified layout XMLs (`frag_channel_search_list.xml`, `act_message_record_layout.xml`) correctly wire ViewPager2 + TabLayout + SmartRefreshLayout. No relative URL or heredoc issues.

**C6 — Governance/security doc:** N/A. No policy or governance docs changed.

---

## Extra Observations (non-blocking)

1. **`ChannelSearchMediaFragment`** maintains `allMedia` ArrayList and rebuilds entries via `toEntries(allMedia)` on each page. For very large result sets (1000+ items), this could cause memory pressure. Consider lazy entry generation or pagination-aware adapter. Not blocking — acceptable for v1.

2. **`SearchWithImgActivity.buildContent()`** sets `video.url = hit.getThumb_url()` (cover, not video file). The comment correctly notes this is a known limitation and forward path falls back to local DB. Acceptable for v1.

3. **`MessageRecordActivity`** uses `offscreenPageLimit = fragments.size` (4) to keep all tabs alive. This is intentional for search UX (tab switching without re-query) but increases memory footprint. Acceptable trade-off.

4. **`GroupModel.groupMembersSyncAndVerify()`** does a local count check after sync. If the count is still low, it falls back to full sync (v=0). This is a defensive pattern for bot-invite edge cases. The fallback is correct but adds one extra network round-trip in the failure path. Acceptable — the verify step is cheap.

---

## Verdict

**[Octo-Q] verdict: APPROVE**

No P0/P1 blockers found. Two P2 issues in error-handling paths (blank screen on initial search failure; silent load-more failure loop) are UX degradations but do not cause data loss, crashes, or security issues. The core feature (OpenSearch-backed channel search, 4-tab UI, image search refactor, group member sync hardening) is correctly implemented with clean data flow from UI → ViewModel → API → adapter.

The P2 findings are actionable and can be addressed in a follow-up PR.

wkbase/src/main/java/com/chat/base/search/channel/Rfc3339.kt
```diff
@@ -0,0 +1,50 @@
...
+
+    fun toEpochSeconds(s: String?): Long {
+        if (s.isNullOrEmpty()) return 0L
+        // SimpleDateFormat 的 Z 模式接受 +0000，不接受 Z 字面量或 +08:00。先标准化。
+        val normalized = s.replace("Z", "+0000")
+            .let { OFFSET_COLON.replace(it) { m -> "${m.groupValues[1]}${m.groupValues[2]}${m.groupValues[3]}" } }
+        for (pattern in PATTERNS) {
+            val sdf = SimpleDateFormat(pattern, Locale.US).apply {
+                timeZone = TimeZone.getTimeZone("UTC")
+                isLenient = false
```
OctoBoooot, line 49: **[major]** Fractional-second timestamps parse to epoch 0. The patterns are only `…ss'Z'`-equivalent and `…ss` with `isLenient=false`, so `toEpochSeconds("2026-06-29T10:30:00.123Z")` returns `0L`. Any server `sent_at` with `.SSS` flows through `MessageHitMapper.toGlobalMessage` into `timestamp=0` → hits sort to Jan-1970 and month-bucket headers corrupt, with no error signal. `Rfc3339Test` omits this exact input. Add a `.SSS` pattern variant before the plain ones (or strip the fraction), and confirm whether octo-server emits millis.

wkuikit/src/main/java/com/chat/uikit/chat/search/channel/ChannelSearchMediaFragment.kt
```diff
@@ -0,0 +1,119 @@
...
+        ChannelSearchModel.searchMedia(req) { outcome ->
+            if (seqSnapshot != requestSeq) return@searchMedia
+            if (!outcome.ok) {
+                handleNonSuccess(outcome, isReset, keyword)
+                return@searchMedia
+            }
+            val list = outcome.data!!
+            if (isReset) allMedia.clear()
+            allMedia.addAll(list.data)
+            adapter.setList(ChannelMediaHitAdapter.toEntries(allMedia))
```
OctoBoooot: **[major]** Media pagination re-`setList`s the full cumulative list every page (`allMedia.addAll` → `setList(toEntries(allMedia))`), which triggers a full rebind + Glide reload + scroll/sticky-header reset on each load-more — O(n²) main-thread work. The three sibling tabs all use `addData` for append (File:74, Message:80, All:77). The full rebuild is here because month-header bucketing needs cross-page context, so the fix must append only the new tail while carrying `lastBucket` across pages (otherwise a page starting in the previous page's month emits a duplicate header). `ChannelMediaHitAdapterTest` covers one-shot bucketing but not this cross-page boundary.

wkuikit/src/main/java/com/chat/uikit/chat/search/channel/BaseChannelSearchFragment.kt
```diff
@@ -0,0 +1,244 @@
...
+        executeSearch(keyword, nextCursor, isReset = false)
+    }
+
+    /** 子类在拿到 outcome 时调用，处理通用错误 + 本地兜底；返回 true 表示已处理。 */
+    protected fun handleNonSuccess(
+        outcome: ChannelSearchOutcome<*>,
+        isReset: Boolean,
+        keyword: String,
+    ): Boolean {
+        binding.refreshLayout.finishRefresh()
```
OctoBoooot: **[minor]** Network callbacks touch `binding` (`_binding!!`) with no post-`onDestroyView` guard. The `requestSeq` snapshot discards stale responses but not post-destroy ones: closing the fragment while a request is in flight fires the callback with `requestSeq` unchanged and `_binding` null → NPE. Null-check `_binding` at callback entry, or route callbacks through `viewLifecycleOwner.lifecycleScope`.

wkuikit/src/main/java/com/chat/uikit/chat/search/image/SearchWithImgActivity.java
OctoBoooot, line 338: **[minor]** `next_cursor` defaults to `""` (non-null). If the server ever returns `has_more=true` with a blank `next_cursor`, the next request sends `cursor=""`, is treated as a non-reset load-more, and appends a duplicated first page. Treat a blank cursor as end-of-list (`hasMore = has_more && !next_cursor.isBlank()`).

wkuikit/src/main/java/com/chat/uikit/chat/search/image/SearchWithImgActivity.java
OctoBoooot, line 319: **[minor]** On a load-more failure or empty non-reset page, `nodataTv`/`hasMore` are left untouched, so the list silently stops growing with no "no more data" footer unless the server also returns `has_more=false`. Fine if that server contract is guaranteed; otherwise finish load-more with no-more-data on the error branch too.

wkbase/src/main/java/com/chat/base/search/channel/ChannelSearchModel.kt
```diff
@@ -0,0 +1,117 @@
...
+                retryAfterSec = parsed.retryAfterSec,
+            )
+        }
+        // 无结构化 error body：按 HTTP 状态码兜底分类。
+        val fallbackCode = when (httpStatus) {
+            in 500..599 -> SearchErrorCode.UPSTREAM_UNAVAILABLE
+            429 -> SearchErrorCode.RATE_LIMITED
+            404 -> SearchErrorCode.NOT_FOUND
+            400 -> SearchErrorCode.VALIDATION_FAILED
+            0 -> ChannelSearchOutcome.LOCAL_ERROR_NETWORK   // ResponseExceptionHandle 对网络异常通常给非 HTTP code
```
OctoBoooot: **[minor]** `ChannelSearchOutcome.LOCAL_ERROR_PARSE` is declared but never assigned anywhere in the data layer. Either wire it up (body present but unparseable) or remove it as dead code.

wkuikit/src/main/java/com/chat/uikit/chat/search/image/SearchWithImgActivity.java
OctoBoooot, line 395: nit: `getDuration_ms() / 1000` truncates integer seconds — a 900ms clip shows 0s, a 1500ms clip shows 1s. Cosmetic; consider rounding.

wkuikit/src/main/java/com/chat/uikit/chat/search/SearchMessageAdapter.kt
OctoBoooot, line 43: praise: Switching the row avatar from channel to `from_uid` genuinely fixes the empty-avatar-in-thread case and aligns with web/iOS. Clear comment on the why.

wkuikit/src/main/java/com/chat/uikit/group/service/GroupModel.java
OctoBoooot, line 289: praise: The bot-invite fix (issue #84) is verified correct — `groupMembersSyncAndVerify` → incremental sync → verify uids → `forceFullMembersSync(v=0)` → `save()` fires `onAddMembers`, which `GroupDetailActivity` subscribes to, so a live-open member grid refreshes without reopen. Verify-before-fallback is the right cost trade-off. One gap: this path ships with no test though `GroupModelTest.kt` is the intended home — the null/empty/missing-uid branching is the highest-value JVM unit test to add.

OctoBoooot (changes requested):
## Review: feat(channel-search): OpenSearch 频道搜索 + 子区查找入口 + 图片搜索切新接口 (#83)

**Verdict:** Request changes. Two real correctness/perf majors (timestamp parsing, media pagination); no blockers. The feature is well-structured and the data layer is genuinely well-tested — these are fixable edge cases, not a design problem.

CI: only `check-sprint` shows a stale red; the latest run is green and all code CI (build / secret-scan / osv-scan / pr-title-lint) passes. `Closes #84` present. No CI blocker.
Risk tier: normal, but flagged **needs-human-review** — 51 files / 3407 lines (size gate: both thresholds). A maintainer should own the final merge call given the scope.

Reviewed across three layers (data / UI / refactor) with each major adversarially verified against source at head 3e889b40.

### Major
- `wkbase/.../search/channel/Rfc3339.kt:31-48` — **fractional-second timestamps parse to epoch 0.** Patterns are only `…ssZ` and `…ss` with `isLenient=false`, so `toEpochSeconds("2026-06-29T10:30:00.123Z")` → `0L` (verified). Any server `sent_at` carrying `.SSS` flows through `MessageHitMapper.toGlobalMessage` into `gm.timestamp=0` → hits sort to Jan-1970 and month-bucket headers corrupt, with no error signal. `Rfc3339Test` is thorough but omits exactly this input. Fix: add a `.SSS` pattern variant before the plain ones, or strip the fraction. Confirm whether octo-server emits millis.
- `wkuikit/.../chat/search/channel/ChannelSearchMediaFragment.kt:85-87` — **media pagination re-`setList`s the full cumulative list every page** (`allMedia.addAll` → `setList(toEntries(allMedia))`), O(n²) main-thread rebind + Glide reload + scroll/sticky-header reset on each load-more. All three sibling tabs use `addData` for append (File:74, Message:80, All:77). The full rebuild exists because month-header bucketing needs cross-page context — so the fix must append only the new tail while carrying `lastBucket` across pages (else a page starting in the previous page's month emits a duplicate header). `ChannelMediaHitAdapterTest` covers the one-shot bucketing but not the cross-page continuation boundary that this bug lives in.

### Minor
- `BaseChannelSearchFragment.kt:167-222` — network callbacks touch `binding` (`_binding!!`) with no post-`onDestroyView` guard. The `requestSeq` snapshot discards *stale* responses but not *post-destroy* ones: close the fragment while a request is in flight → callback fires → `_binding` is null → NPE. Null-check `_binding` at callback entry.
- `SearchWithImgActivity.java:338` — `nextCursor = pagination.next_cursor` defaults to `""` (non-null); a server returning `has_more=true` with a blank cursor sends `cursor=""`, treated as load-more → duplicate first page appended. Treat blank cursor as end-of-list.
- `SearchWithImgActivity.java:319-321` — on load-more failure/empty non-reset page, `nodataTv`/`hasMore` are left untouched; list silently stops growing unless server also returns `has_more=false`. Fine if that contract is guaranteed.
- `ChannelSearchModel.kt:112` — `ChannelSearchOutcome.LOCAL_ERROR_PARSE` is declared but never assigned — dead code, or wire it to the unparseable-body path.

### Nits (≤3)
- `SearchWithImgActivity.java:395` — `getDuration_ms()/1000` truncates (900ms → "0s"); round instead.
- `ChannelCombinedHitAdapter.bindFile:98-106` and `ChannelFileHitAdapter.convert:42-50` — identical `<mark>`→`<font>` highlight logic; extract to a shared util.

### Questions
- `ChannelSearchBodyBuilder.build()` writes `sort`/`page_size` for all four endpoints including `_search_media`/`_search_files` (`buildAround` correctly omits `sort`). Do the media/file endpoints accept `sort`, or will it 400? Not covered by the body-builder tests.
- `MediaHit` non-null (`lateinit`/non-null Gson-populated) fields — is any Kotlin caller dereferencing these directly such that an omitted server field NPEs? Java callers guard; Kotlin path unverified.

### Praise
- **Data layer is the model to follow:** `ChannelSearchBodyBuilder` builds a typed FastJSON object (no string-concatenated query → no injection), the omit-empty-keyword/filters/cursor logic matches the server's `keywordOrFilterRequired` contract with a matching test per branch, and `ChannelSearchErrorParserTest` / `ChannelSearchOutcomeTest` genuinely exercise every error code and envelope shape — not trivially-passing tests.
- The `requestSeq` snapshot pattern (All:68-70 et al.) cleanly discards out-of-order paginated responses — the classic search-UI bug, handled correctly.
- **Bot-invite fix (issue #84) verified correct:** `groupMembersSyncAndVerify` → incremental sync → verify uids → `forceFullMembersSync(v=0)` → `save()` fires `onAddMembers`, and `GroupDetailActivity` subscribes, so a live-open member grid refreshes without reopen. `v=0` (`WHERE version>0`) does pull the bot's low-version row, and the full-sync recursion carries a null listener so verification runs at most once — no loop. Verify-before-fallback is the right cost trade-off, and reusing it for the "duplicate invite → silent success" case is a catch beyond the literal bug.
- `SearchMessageAdapter.kt:43` — switching the row avatar from channel to `from_uid` fixes the empty-avatar-in-thread case; good iOS/web alignment, well-commented.

- Test coverage across the whole PR is ~0.11 — the data layer is well covered, but the entire UI layer (6 fragments + ViewModel + 3 adapters) and the bot-invite fallback (`GroupModel` / `WKIMUtils` / `ChooseContactsActivity`) ship with no tests, despite `GroupModelTest.kt` being the intended home and the PR test-plan listing member refresh. Not a blocker, but the bot-invite branching (null vs empty vs missing-uid) is the highest-value JVM unit test to add.

*51 files / +3072 −335. needs-human-review applied (size). Two majors → REQUEST_CHANGES; both are contained edge cases on an otherwise solid, well-tested feature.*

yujiawei (changes requested):
# Code Review — PR #83 (octo-android)

Reviewed at head SHA `3e889b40ca18d983bb4a918eef0d00babadda7b3` against `main` (merge-base `9b5990f9`). Scope: 51 files, +3072/-335 — OpenSearch-backed channel search (data layer + 4-tab UI), thread "find messages" entry, image-search endpoint switch, group bot-invite fallback, version bump to 1.3.5 (36).

This review has two parts: **(1) scope/spec compliance**, **(2) code quality**. Both must pass to approve.

---

## 1. Scope & spec compliance — ✅ (with one note)

The diff implements what the PR describes and nothing contradictory:
- `_search`/`_search_all`/`_search_media`/`_search_files` data layer with strongly-typed DTOs, body builder, structured error parser, RFC3339 util — all present and unit-tested.
- 4-tab result UI (All / Message / Media / File) driven by a shared `ChannelSearchViewModel` with 300ms debounce.
- Thread detail "find chat history" entry reusing `MessageRecordActivity` with `channel_type=5`.
- Image search switched from local+GlobalSearch to `_search_media` cursor pagination, with video play badge and forward/favorite fallbacks.
- Group bot-invite `groupMembersSyncAndVerify` fallback in both the 1002-message path and the manual add/invite callbacks.
- `versionCode 35→36`, `versionName 1.3.4→1.3.5`.

**Note (not blocking):** the `_search_around` endpoint, `AroundRequest`/`AroundResult`, `buildAround`, and `ChannelSearchModel.searchAround` are fully implemented and tested but have **no callers anywhere in the diff** (verified: no references outside the search package). This is unused scaffolding. Either wire it to a "jump to message context" flow or drop it to avoid shipping dead code — your call, it doesn't block merge.

---

## 2. Code quality — Changes Requested

### P1 — Fragment can crash (NPE) on a late search callback after view is destroyed

`BaseChannelSearchFragment` holds a nullable binding:

```kotlin
protected var _binding: FragChannelSearchListBinding? = null
protected val binding get() = _binding!!          // line 55
...
override fun onDestroyView() { super.onDestroyView(); _binding = null }   // line 232
```

The search requests go through `WKBaseModel.requestAndErrorBack`, whose RxJava subscription is **not** disposed on fragment lifecycle (no `CompositeDisposable`, no `takeUntil`). The result is delivered on the main thread whenever the network completes, regardless of fragment state. The per-callback `requestSeq` guard only suppresses *stale* results — it does not protect against the view being gone:

- User enters channel search, types a keyword → request in flight.
- Before it returns (slow network), user presses back → activity finishes → `onDestroyView()` → `_binding = null`.
- Response arrives → `requestSeq` guard passes (seq unchanged) → `handleNonSuccess(...)` / `updatePaginationState(...)` touch `binding.refreshLayout` → `_binding!!` throws `NullPointerException` → crash.

This is reachable with a normal "search, then navigate away" gesture. Notably, `SearchWithImgActivity` already guards its async path with a `loading` flag — the fragments have no equivalent view-liveness guard.

**Fix options:** switch the callback bodies to `_binding?.let { ... }` / early-return when `_binding == null`, or tie the subscription to `viewLifecycleOwner` and dispose in `onDestroyView`.

### P2 — Findings (non-blocking, but worth addressing)

1. **Empty-keyword reset does not bump `requestSeq` → stale results can repopulate a cleared list.** In `triggerReset` (line 136), the `keyword.isEmpty() && !supportsBrowseWithoutKeyword` branch calls `resetList()` and returns **without** incrementing `requestSeq`. If a prior keyword query is still in flight when the user clears the box, its response passes the `seqSnapshot == requestSeq` check and re-fills the list the user just emptied. Bump `requestSeq` in that branch too.

2. **`loadMore()` has no in-flight guard → possible duplicate same-cursor page.** `loadMore` (line 150) gates on `hasMore`/`nextCursor`, both of which only update when a response returns. Two rapid load-more triggers can issue the same `nextCursor` twice and append the page twice. SmartRefreshLayout mitigates this in practice, but the image-search path already uses an explicit `loading` flag — the fragments should match it for robustness. (Independently flagged by two review passes.)

3. **`MediaFragment` rebuilds the whole list on every page.** `executeSearch` accumulates into `allMedia` and calls `adapter.setList(toEntries(allMedia))` on each page, re-grouping and full-rebinding the entire dataset per page instead of appending. For channels with lots of media this is O(n²) over pagination plus a full RecyclerView rebind each page — visible jank on long histories. Consider appending only the new page's entries.

4. **Deserialized DTOs use `lateinit var` for server-controlled fields.** `CombinedHit.result_type`, `MediaHit.month_bucket`/`sent_at`, `MessageHit.message_id`, etc. are `lateinit`. If the server ever omits one, FastJSON leaves it uninitialized and the first access (e.g. `isMessage()`, `toEntries` reading `month_bucket`) throws `UninitializedPropertyAccessException`. This matches the repo's existing DTO convention (e.g. `GlobalMessage`), so it's not newly wrong — but it's a latent crash surface that hardens cheaply with `String = ""` defaults or nullables.

5. **`Rfc3339.toEpochSeconds` allocates `SimpleDateFormat` per call.** Called per row during list binding (`MessageHitMapper`, both file/combined adapters). `SimpleDateFormat` construction/parsing is expensive; cache via `ThreadLocal` to reduce GC churn while scrolling.

### Notes on adjudication

I checked, and **rejected as not-real**, a couple of null-crash concerns raised during review: `outcome.data!!` in the fragments is safe because `ChannelSearchOutcome.success` is only constructed from a non-null RxJava emission (`onNext(@NotNull T)`), and `CursorList.data`/`pagination` carry non-null default initializers, so an omitted `pagination` does not NPE. These would only fail if the server sent an explicit JSON `null` for `pagination`, which is a low-likelihood edge — covered by adding defensive handling if you touch the area, but not a bug today.

Cross-file dependencies verified as sound: `getWithMessageID`, `GlobalMessage.message_idstr`, `WKVideoContent.second`/`cover`, `WKContentType.WK_IMAGE=2`/`WK_VIDEO=5`, all referenced layout IDs (`playIv`, `thumbIv`, `monthHeaderTv`, `frag_channel_search_list`, `item_channel_search_file`), `icon_play` mipmap, and all new string resources (zh + en) resolve. No stale references to the removed `GlobalSearchModel`/`seenSeqs`/`page` code paths remain.

---

## Verdict

**Spec: ✅** — implements the stated scope; only observation is the unused `_search_around` scaffolding.
**Quality: Changes Requested** — one reachable NPE crash (P1) on late callback after view destruction.

Requesting changes for the P1. The P2 items are recommended follow-ups; the stale-`requestSeq` and `loadMore` guards are quick and worth folding into the same pass.

yujiawei:
### Addendum to my review — one more P1 (network-fallback regression)

After a second pass and cross-checking with the other reviews, I confirmed a blocking issue I under-weighted in my main review. Worth calling out explicitly because it defeats the central design goal of the new error layer:

**P1 — Network/timeout failures never trigger the intended local fallback.**

`ResponseExceptionHandle.handleException` assigns **`-1`** to any non-`HttpException` (connection failure, timeout, DNS, etc.):

```java
// ResponseExceptionHandle.java:110
responeThrowable = new ResponseThrowable(e, -1);
```

But `ChannelSearchModel.mapFailure` only maps code `0` to the local-network error:

```kotlin
// ChannelSearchModel.kt (fallback branch)
0 -> ChannelSearchOutcome.LOCAL_ERROR_NETWORK
else -> SearchErrorCode.INTERNAL
```

So a real transport failure arrives as `-1`, falls through to `INTERNAL`, and `uiAction()` returns `GENERIC_ERROR` instead of `FALLBACK_TO_LOCAL`. Net effect: the offline banner + IMSDK local search fallback documented in `ChannelSearchOutcome`/`BaseChannelSearchFragment` **never fires when the network is actually down** — the exact scenario it was built for. The user just gets a generic "search failed" toast.

**Fix:** map `-1` (or, more robustly, all non-positive/transport codes) to `LOCAL_ERROR_NETWORK` in `mapFailure`.

**Related P2 (same root):** for `429`/`503` the structured `errorBody` is not preserved into `errJson` by `ResponseExceptionHandle` (only `400` sets it), so `ChannelSearchErrorParser` can't see server codes like `disabled` or `retry_after`. Consequences: a `disabled` (503) is misclassified as `upstream_unavailable → FALLBACK_TO_LOCAL` instead of `FEATURE_DISABLED`, and rate-limit `retry_after` is lost. Consider preserving `errJson` for all HTTP error statuses if you want the structured-code UI paths to work as designed.

This is additive to the P1 (post-destroy binding NPE) and the P2 list in my review above; the overall verdict stands at **Changes Requested**.

OctoBoooot (approved):
## Re-review (delta): feat(channel-search) (#83) @ 14d75cd0

**Verdict:** Approve. Every blocking/major item from the prior round is resolved at this head, each with a matching test. CI is fully green (build + check-sprint).

Delta from 3e889b40 → 14d75cd0: 6 files (2 src fixes + 2 tests + the fallback-mapping + NPE guards).

### Prior findings → resolved (verified against source)
- ✓ **Network→local fallback break** (Steve's blocker) — `ChannelSearchModel.kt:107-114` `mapFailure` rewritten to `httpStatus <= 0 -> LOCAL_ERROR_NETWORK`. Offline/timeout (`ResponseExceptionHandle` code `-1`) now → FALLBACK_TO_LOCAL instead of GENERIC_ERROR. The dead `0`-only branch is gone; the fix covers -1 and any non-positive transport code exactly as prescribed.
- ✓ **Rfc3339 `.SSS` → epoch 0** (my major) — `Rfc3339.kt:33,44` strips fractional seconds via `FRACTION = \.\d+` in the normalization step. Four new tests cover millis, micros, nanos, and fraction-with-offset — including the exact `2026-06-29T10:30:00.123Z → 1782729000` case the prior test omitted.
- ✓ **Media pagination full-rebuild** (my major) — `ChannelSearchMediaFragment.kt:88-103` now `addData`s only the new page tail; `toEntries(list.data, previousBucket = lastMonthBucket)` carries the last month across pages so no duplicate header at a same-month page boundary. `ChannelMediaHitAdapter.toEntries` gained the `previousBucket` param, and `ChannelMediaHitAdapterTest` gained a cross-page continuation case. This is exactly the append-with-carried-bucket fix, and it closes the O(n²) rebind.
- ✓ **View-destroy NPE** (my minor / yujiawei's 🔴) — every async callback in `BaseChannelSearchFragment.kt` (`handleNonSuccess`, `updatePaginationState`, `showEmpty`, `showOfflineBanner`, `executeLocalFallback`) now opens with `val b = _binding ?: return`, so a response landing after `onDestroyView` is a silent no-op.

### Praise
- Clean, surgical fixes — each addresses the root cause (not a symptom patch) and each ships with the test that would have caught it. The media-pagination fix in particular took the harder-but-correct path (carry `lastMonthBucket` across pages) rather than reverting to the full rebuild.

### Out of scope (informational, non-blocking follow-ups)
- The bot-invite fallback (`GroupModel`/`WKIMUtils`/`ChooseContactsActivity`) still has no test — the highest-value remaining follow-up, and a good candidate for a fast-follow PR.
- Minor carryovers not required for merge: `SearchWithImgActivity` blank-cursor / load-more-error affordance and the `duration_ms/1000` truncation; the shared `<mark>`→`<font>` highlight duplication. Author's call.

*Retires my prior CR at 3e889b40. All three reviewers' blockers (mine, Steve's, yujiawei's) resolved at this head. needs-human-review remains (size gate: 51 files) for a maintainer's final merge call.*

Jerry-Xin (dismissed):
Summary: The PR is in scope for octo-android, but it currently has a compile-blocking Java/Kotlin interop error and a functional regression in the stated clear-keyword behavior.

🔴 Blocking

- 🔴 Critical: `SearchWithImgActivity` directly accesses private Kotlin property backing fields on `GlobalMessage`.
  `GlobalMessage` is a Kotlin class with normal `var` properties, so Java must use generated accessors. These lines will not compile:
  `wkuikit/src/main/java/com/chat/uikit/chat/search/image/SearchWithImgActivity.java:193`, `249-253`, `269`.
  Use `getMessage_idstr()` / `getFrom_uid()` or keep the value in `SearchImgEntity`.

- 🔴 Critical: Clearing the search keyword does not clear all result lists as described in the PR/test plan.
  `MessageRecordActivity` keeps the 4-tab result pager open after it has been created, even when the text becomes empty (`MessageRecordActivity.kt:153-162`). Then `BaseChannelSearchFragment` still executes searches for tabs with `supportsBrowseWithoutKeyword=true` (`BaseChannelSearchFragment.kt:137-148`), and media/files explicitly enable that behavior (`ChannelSearchMediaFragment.kt:50-52`, `ChannelSearchFileFragment.kt:41-42`). This means clearing the input re-populates media/file results instead of clearing them.

💬 Non-blocking

- 🟡 Warning: I could not run Gradle verification because this environment has no Java runtime installed: `Unable to locate a Java Runtime.` Please run the listed unit tests plus a debug compile after fixing the blockers.

✅ Highlights

- The new request body builder has focused tests around omitted empty fields and cursor passthrough.
- The media pagination adapter handles cross-page month headers without rebuilding the whole list.

Jerry-Xin (approved):
All four standing blockers from the prior round are genuinely fixed at this head (byte-verified end to end, each with accompanying unit tests).

Verified fixes
- Offline fallback (-1 mapping): ChannelSearchModel.mapFailure now maps httpStatus <= 0 to LOCAL_ERROR_NETWORK. Full chain re-traced: non-HTTP exception -> ResponseExceptionHandle assigns code -1 with empty error body -> BaseObserver.onFail(-1) -> parser returns null -> -1<=0 branch -> LOCAL_ERROR_NETWORK -> uiAction() = FALLBACK_TO_LOCAL. Fires correctly now.
- Rfc3339 fractional seconds: normalize step strips a \.\d+ fraction, so .SSS / .SSSSSS / .SSSSSSSSS and offset/naive forms no longer parse to epoch 0. Four new tests cover ms/us/ns/offset.
- Media pagination: switched from full setList(toEntries(all)) per page to addData(newTail); a lastMonthBucket cursor passed into toEntries avoids duplicate month headers across pages. Matches sibling tabs. Four new adapter tests.
- Fragment view-destroy NPE: handleNonSuccess / updatePaginationState / executeLocalFallback / showEmpty / showOfflineBanner all guard on _binding before use, and _binding is nulled in onDestroyView.

Note on the compile concern raised mid-review
- I initially flagged Java field-style access to GlobalMessage (message_idstr / from_uid) in SearchWithImgActivity.java as a compile break. On verification this is a false positive: the CI build job runs `./gradlew testDebugUnitTest assembleDevDebug`, app depends on :wkuikit, and assembleDevDebug is green at this exact head, so this code compiles. Withdrawn.

Non-blocking (discussion, not gating)
- Clear-keyword behavior: Media/File tabs set supportsBrowseWithoutKeyword=true and intentionally browse-all on empty keyword, so clearing the input re-populates those tabs rather than emptying them. This is a deliberate design choice but differs from the PR test-plan wording ("clear keyword -> results cleared"); worth aligning the copy or the behavior in a follow-up.

Highlights: focused request-body-builder tests, clean tab-fragment UI split, conservative post-invite member sync fallback.

LGTM.

yujiawei (approved):
# Code Review — PR #83 (octo-android)

Re-reviewed at head SHA `14d75cd0` against `main` (merge-base `9b5990f9`). This head addresses the four blocking issues raised in the previous round (prior head `3e889b40`). The incremental diff is 7 files / ~130 lines; I verified each fix against the prior finding and scanned the increment for regressions. All four prior blockers are resolved and no regression was introduced.

This review has two parts: **(1) scope/spec compliance**, **(2) code quality**.

---

## 1. Scope & spec compliance — ✅

The diff implements exactly the stated scope (OpenSearch channel-search data layer + 4-tab UI, thread "find messages" entry, image-search endpoint switch, group bot-invite fallback, version bump 1.3.5/36) with nothing out of scope. The incremental changes are confined to the four review fixes. No missing, extra, or contradictory functionality.

*Note (non-blocking, carried over):* the `_search_around` path (`AroundRequest`/`AroundResult`/`buildAround`/`searchAround`) is fully implemented and tested but still has no callers in the diff — unused scaffolding. Wire it to a "jump to context" flow or drop it; doesn't block merge.

---

## 2. Code quality — Approved

### Verification of the four prior blockers (all resolved)

**1. Network-failure local fallback — FIXED.** `ChannelSearchModel.mapFailure` changed from `when (httpStatus) { 0 -> LOCAL_ERROR_NETWORK }` to a subject-less `when { httpStatus <= 0 -> LOCAL_ERROR_NETWORK }`. `ResponseExceptionHandle` assigns `-1` to transport-layer exceptions (timeout/DNS/connect), which previously fell through to `INTERNAL` and skipped fallback; now any non-positive code routes to `LOCAL_ERROR_NETWORK → FALLBACK_TO_LOCAL`. Branch ordering is sound (500-599/429/404/400 are positive, no overlap). Confirmed the outcome→`uiAction()` chain and the existing `ChannelSearchOutcomeTest` assertion.

**2. RFC3339 fractional seconds — FIXED.** `Rfc3339` now strips `\.\d+` before `SimpleDateFormat`, so `2026-06-29T10:30:00.123Z` (and micro/nanosecond precision) parse correctly instead of returning `0`. The normalization order (`Z→+0000`, `+08:00→+0800`, strip fraction) is safe — dots appear only in the sub-second field, so the regex cannot corrupt the date or offset. Four new `Rfc3339Test` cases assert correct epoch values across precisions and offsets.

**3. Fragment NPE on late callback (my prior P1) — FIXED.** `handleNonSuccess`, `updatePaginationState`, `showEmpty`, `showOfflineBanner`, and the base `executeLocalFallback` now guard with `val b = _binding ?: return` / `_binding?.`. I traced the "search then navigate away" path end-to-end: search callbacks are delivered via `WKBaseModel.requestAndErrorBack`, which `observeOn(AndroidSchedulers.mainThread())` — the same thread as `onDestroyView`. The `_binding` guard in `handleNonSuccess` and the synchronous call into `executeLocalFallback` run in one uninterrupted main-thread turn, so `_binding` cannot be nulled mid-callback. The crash is no longer reachable.

**4. Media pagination O(n²) rebuild — FIXED.** `ChannelSearchMediaFragment` now appends via `adapter.addData(newEntries)` instead of re-running `setList(toEntries(allMedia))` per page, and threads `lastMonthBucket` into `toEntries(previousBucket = …)` to suppress duplicate month headers across page boundaries. I verified: `previousBucket` uses the prior page's last bucket (updated *after* `toEntries`), `isReset` resets the cursor and uses `setList`, empty pages leave `lastMonthBucket` untouched with no phantom header, and `isEmpty` reads `adapter.data` after mutation. Five new `ChannelMediaHitAdapterTest` cases cover the cross-page header cases including the mixed boundary.

### Regression scan of the increment
No regressions. `toEntries`'s new `previousBucket` is an optional named-arg param (only production caller passes it by name; back-compat preserved). `addData(Collection)`/`.data` are provided by BRVAH `BaseMultiItemQuickAdapter`. The `_binding` guards protect UI access only and don't drop any non-UI state update (`hasMore`/`nextCursor` are set inside the guarded helpers after the check).

### Follow-ups (non-blocking, carried from prior round)
These remain optional and do not block merge:
- `triggerReset` empty-keyword branch still returns without bumping `requestSeq`; a still-in-flight prior query could repopulate a just-cleared list. Bump the seq in that branch.
- `loadMore()` has no explicit in-flight guard (relies on SmartRefreshLayout to debounce) — an explicit flag like the image-search path would be more robust.
- Server-controlled DTO fields use `lateinit var`; matches repo convention but hardens cheaply with defaults/nullables.
- `Rfc3339.toEpochSeconds` allocates a `SimpleDateFormat` per call during row binding — a `ThreadLocal` cache would cut GC churn while scrolling.

*Minor maintainability nit on the NPE fix:* the subclass `executeLocalFallback` overrides (All/Message fragments) access `binding` directly and rely on the implicit invariant that the sole caller guards `_binding` on the main thread. It's safe today, but adding a `val b = _binding ?: return` at the top of those overrides would make them self-defensive if a future caller is added.

---

## Verdict

**Spec: ✅** — implements the stated scope; only observation is the unused `_search_around` scaffolding.
**Quality: Approved** — all four prior blockers verified fixed, no regressions; remaining items are non-blocking follow-ups.

Approving.

PR #92, opened 2026-07-15
feat(chat): 批量改进 — 贴纸/预览/名称长度/撤回/release 修复

### LINKED ISSUE #87: [Android] 消息撤回经常无反应 / 要点多次才成功
## 现象
用户实测：撤回消息经常没反应，或要点好几次才成功。

## 根因
**根因 1：clientMsgNO 被塞给 message_id 参数**
- 服务端 `verifyRevokeMessageID` 用 `strconv.ParseInt` 解析 `message_id`，UUID 形态的 clientMsgNO 必失败 → `ErrMessageIDSeqMismatch`
- 服务端本已支持空 `message_id` 反查（见 `octo-server modules/message/api_authorize.go:57-69`），Android 侧一直没利用

**根因 2：只走 CMD → syncExtraMsg 单路径**
- 服务端 `syncMessageExtra` 有 cache bug，`resultSize=0` 概率约 66%
- 单路径下丢一次通知就丢一次撤回

## 修复
- `messageID` 空或 `'0'` 时传空字符串，让服务端自动用 `client_msg_no` 反查真实 messageID
- 撤回改为双路径：HTTP 200 后立即本地 mark，不再只靠 `syncExtraMsg`

## 影响范围
所有单聊 / 群聊 / 子区消息撤回入口。

## 相关 commit
`856bb10 fix(revoke)`

### LINKED ISSUE #88: [Android] 群名/子区名超长静默截断，无用户反馈
## 现象
在编辑群名 / 子区名 / 加好友备注 / Space 名 / 群昵称 等 EditText 已达到 `android:maxLength` 上限后，再输入或粘贴会被 Android 静默吞掉字符，无任何提示，用户困惑（"为什么打不上去 / 粘贴半截"）。

## 修复方案
1. 新增 `ThrottledRunnable`（通用节流 Runnable，2s 内重复调用被吞）
2. `StringUtils.getInputFilter(max, onTruncate)` 重载：InputFilter 截断时回调（覆盖 surrogate pair / 部分截断 / 全拒 三种分支）
3. `StringUtils.attachLengthLimit(EditText, max)` 一行 API：内部装配 filter + 节流 toast，统一 UX（2s 间隔、`str_input_length_limit` 文案）

3 处 EditText 收敛到一行调用：
- `CreateThreadActivity.threadNameEt`（子区名 100 上限）
- `UpdateGroupNameActivity.nameEt`（群名 50 上限）
- `WKDialogUtils.showInputDialog`（惠及 10+ 调用方）

## 顺带上限调整
- 子区名 50 → 100
- 群名 30 → 50

## 行为
- 首次触顶立即弹 toast
- 2s 内额外触发被吞（粘贴超长 / 长按键不刷屏）
- 未触顶不弹 toast

## 相关 commit
`ddee4d0 feat(chat)`

### LINKED ISSUE #89: [Android] release 包频道搜索崩溃 (ClassCastException)
## 现象
release 包打开频道搜索的 4 个 tab（Message / All / Media / File）时抛 ClassCastException，Bugly 上报 UndeliverableException → Caused by CCE。debug 包正常。

## 根因
`com.chat.base.search.channel.dto.**` 未加 proguard keep。R8 minify 后字段名被混淆，FastJson 反序列化产物类型错乱，`val list = outcome.data!!` 处 checkcast 失败。

## 修复
在 `app/proguard-rules.pro` 补 keep 规则保留该包所有类的字段名。

## 相关 commit
`1a91400 fix(channel-search)`

## 备注
同类 proguard 遗漏问题另见 issue（自定义贴纸 release 崩溃），后续可考虑做一次全项目 FastJson DTO keep 规则扫描。

### LINKED ISSUE #90: [Android] 聊天窗口内文件预览 (PDF / xlsx / 文本) + 修长文件名静默下载失败
## 背景
聊天窗口内文件消息只能"打开 / 保存 / 分享"，走系统 App。对 PDF、Excel、文本等常见办公格式，用户希望不切出去也能快速预览一眼。

## 新增：应用内预览
文件消息底部新增"预览"按钮：
- **pdf** → `PdfPreviewActivity`（Android 原生 `PdfRenderer`，零依赖）
- **xlsx** → `XlsxPreviewActivity`（自写零依赖 `XlsxParser` 解析 sharedStrings/workbook/sheet → HTML table → WebView，JS 已关闭）
- **txt / md / json** 等 30+ 文本扩展 → 复用 `TextPreviewActivity`
- 其它扩展名"预览"按钮隐藏，弹框仍是"打开 / 保存 / 分享"
- "打开"路径完全未改，走系统 App，与老行为一致

xls / doc / ppt / pptx 老格式暂不支持 App 内预览，走系统 App 兜底。

## 顺带修 bug：长文件名静默下载失败
现象：文件名超过 Linux `NAME_MAX`（255 字节 UTF-8）时 `FileOutputStream` 抛 `ENAMETOOLONG` 被静默吞掉，用户观感是"点了没反应"。100+ 中文/emoji 文件名很容易触发。

修法：`sanitizeFileName` 加 200 字节按 code point 截断，保留扩展名，避免破坏 UTF-8 surrogate pair。

## 相关 commit
`b275d6b feat(chat)`

### LINKED ISSUE #91: [Android] 自定义贴纸功能（发送 / 接收 / 收藏 / 管理）
## 目标
支持自定义贴纸消息类型 + 用户表情面板管理，覆盖发送 / 接收 / 收藏 / 删除 / 排序 / 上传全链路。

## 消息模型
新增两种 content type：
- `WKVectorStickerContent` (12)
- `WKEmojiStickerContent` (13)

字段：`url / width / height / category / placeholder / format`。

Android 不引入 Lottie SDK：
- 静态图（png / jpg / webp / gif）走 Glide 直显
- `.lim / .json` 走占位图降级（`StickerUrlUtils` 分类）
- 未注册的 content type 会被 SDK 强制降级为 `unknown_msg(-3)` 显示"未知消息"，因此在 `WKUIKitApplication` 里 `registerContentMsg`

## Cell provider
`WKVectorStickerProvider` (12) + `WKEmojiStickerProvider` (13, 仅覆写 itemViewType) —— 独立实现，不继承 `WKImageProvider`（贴图无气泡 / 无大图预览 / 无上传进度 / 无焚阅；`.lim` URL 也会让 Glide 崩）。

长按菜单启用 reply / withdraw / delete / forward / reaction / pin；贴图开子区语义不合理，故不出现子区入口。

## 收藏链路
`WKStickerManager` (Kotlin object) — LiveData 缓存 + Retrofit API 网关：
- `GET /sticker/user` 拉列表（30s 节流，短时间重复调用短路）
- `POST /sticker/user/collect` 收藏（服务端 `SHA256(source_path)` 天然幂等，Android 侧不做前置去重）
- `DELETE /v1/sticker/user/{id}` 删除
- **收藏走 `/sticker/user/collect` 而不是 `/sticker/user`**：后者需要 handle 签名 + path 必须 uid 前缀，收藏他人贴图会 400

`AddToMyStickersMenuProvider` 通过 `EndpointCategory.wkChatPopupItem` 注入长按菜单，过滤：type 必须 12/13、非自己发的、未收藏（URL 用 `StickerUrlUtils.normalizePath` 归一化）。

## 面板 UI（我的贴图 tab）
- 5 列 `GridLayoutManager`，首格 `+` 上传入口常驻
- `CollectStickerAdapter` 重写为原生 `RecyclerView.Adapter` 支持双 viewType
- **编辑态**：长按非编辑态 sticker → 进入编辑态（cells 抖动 + 红 × 徽章 + 中等 haptic），tap × 二次确认删除，tap 贴图/空白/切 tab/tap `+` 退出编辑
- **拖拽排序**：`ItemTouchHelper` 长按接管重排，服务端无重排接口，本地 SP `sticker_order_{uid}` 存顺序
- **长按预览**：编辑态 hold-still ≥500ms 弹预览弹窗
- **详情弹窗**（`StickerDetailPopup`）：消息内点击贴图 → "发送 + 添加到我的表情"；面板长按 → "发送 + 删除"
- 关键交互决策：`EditModeTouchListener` 自己 dispatch touch，在同一次长按里区分 preview 和 drag（`ItemTouchHelper.isLongPressDragEnabled = false`）

## 上传
`WKStickerUploader` + `StickerUploadValidator`，三步链路复用 `WKUploader` 双步预签名：
- 客户端校验：magic bytes（gif / png / jpg / webp）+ ≤1024KB + 长边 ≤512px
- `POST /file/upload/credentials?type=sticker` → `PUT CDN` → `POST /sticker/user`
- path 前缀 `/sticker/{uid}/` 满足服务端上传路径校验

## proguard
`com.chat.uikit.chat.sticker.**` 补 keep，避免 release 包 R8 混淆 `ListStickerResp.list` 字段名后 FastJson 无法解析 `{"list":[...]}` → 用户长按"添加到我的表情"在 release 无反应。

## 测试
- `WKStickerContentTest`：type / encode / decode / parcel 回归（7 case）
- `StickerUrlUtilsTest`：后缀识别 + URL 归一化（13 case）
- `StickerLocalOrderStoreTest`：`applyOrder / pruneOrder` 纯函数版本 empty/full/partial/stale ids（8 case）
- `StickerUploadValidatorTest`：magic bytes 识别覆盖 gif87a/gif89a/png/jpeg/webp + bmp/text/riff-wav/empty/short-prefix 拒绝分支（10 case）

## 相关 commit
- `808dcaf feat(sticker): 自定义贴纸功能（发送/接收/收藏）`
- `b99aff0 feat(sticker): 数据层加删除/重排/上传`
- `114ab84 feat(sticker): 面板 UI 重构（5列 + 编辑态 + 长按预览+拖拽 + 详情弹窗）`
- `652f35a test(sticker): 补 LocalOrderStore + UploadValidator 单测`
- `0b7c19b fix(sticker): 补 proguard keep 修复 release 添加表情失败`

mzanalyticst:
## Summary
本次批量合入 5 组独立可追溯的 chat 侧改动（10 笔 commit / 50 文件 / +3722 −50）：

| Issue | 主题 | 类型 |
|---|---|---|
| #87 | 消息撤回经常无反应 —— 修 messageID 参数 + 双路径 | bug |
| #88 | 群名 / 子区名超长静默截断 —— 节流 toast + 上限调整 | enhancement |
| #89 | release 包频道搜索 ClassCastException —— 补 proguard keep | bug |
| #90 | 聊天窗口内文件预览（PDF / xlsx / 文本）+ 修长文件名下载失败 | feature |
| #91 | 自定义贴纸功能（发送 / 接收 / 收藏 / 删除 / 排序 / 上传）| feature |

Closes #87
Closes #88
Closes #89
Closes #90
Closes #91

## 关键设计决策
- **撤回双路径**：HTTP 200 后立即本地 mark，不再单靠服务端 CMD → syncExtraMsg（后者有 cache bug，miss 率 ~66%）；messageID 空/`'0'` 时传空串让服务端反查
- **贴纸不引入 Lottie SDK**：静态图走 Glide，`.lim / .json` 走占位图降级；两种新 contentType 12/13 走独立 provider，不继承 `WKImageProvider`
- **贴纸编辑态自研 touch dispatch**：`ItemTouchHelper.isLongPressDragEnabled = false`，用 `EditModeTouchListener` 在同一次长按里区分 preview（hold-still ≥500ms 弹预览）和 drag（超 touchSlop 才 startDrag）
- **贴纸重排**：服务端无接口，本地 SP `sticker_order_{uid}` 存顺序；reorder 直接用参数排序不走 SP 读回路径，避免 apply() 异步落盘 race
- **文件预览零依赖**：PDF 走 Android 原生 `PdfRenderer`；xlsx 自写 `XlsxParser` 解析 sharedStrings/workbook/sheet → HTML → WebView（JS 已关闭）
- **名称长度节流**：`ThrottledRunnable` + `StringUtils.attachLengthLimit()` 一行 API，收敛 3 处 EditText + 10+ 弹窗调用方

## 影响 & 风险
- 未改任何消息收发 / 存储 / 同步 / 通知 / 子区 / 转发核心路径；贴纸走 Provider 注入模式与既有消息互不影响
- release proguard 已补两处 keep（`com.chat.uikit.chat.sticker.**` 和 `com.chat.base.search.channel.dto.**`），避免 R8 混淆导致 FastJson 反序列化崩溃
- 单测新增 4 个测试类共 38 个 case（`WKStickerContentTest / StickerUrlUtilsTest / StickerLocalOrderStoreTest / StickerUploadValidatorTest`）

## Test plan
- [ ] 撤回：单聊 / 群聊 / 子区自己发的消息撤回一次成功，不再需要重试
- [ ] 撤回：他人撤回消息 CMD 到达后本地能刷新为"已撤回"
- [ ] 名称长度：群名 / 子区名 / 加好友备注 / Space 名 / 群昵称 打字触顶弹 toast，2s 内粘贴超长不刷屏
- [ ] release 频道搜索：Message / All / Media / File 4 个 tab 不再崩溃
- [ ] 文件预览：pdf / xlsx / txt 消息底部"预览"按钮工作正常，其它扩展名隐藏"预览"
- [ ] 长文件名（100+ 中文/emoji）文件下载不再静默失败
- [ ] 贴纸发送：面板选贴图能正常发送，对端能收到并显示
- [ ] 贴纸收藏：长按他人贴图菜单出现"添加到我的表情"，收藏后重复长按菜单不再出现
- [ ] 贴纸编辑：长按进编辑态，tap × 二次确认删除，长按 hold ≥500ms 弹预览，长按拖动排序，切 tab / tap 空白退出
- [ ] 贴纸上传：+ 号选图，超 1024KB / 长边 >512px / 非 gif/png/jpg/webp 会被客户端校验拦截
- [ ] 贴纸详情弹窗：消息内 tap 贴图 "发送 + 添加"；面板编辑态外长按 "发送 + 删除"
- [ ] **release 包全量回归**：贴纸收藏 / 频道搜索这两个 R8 敏感路径重点验证

### DIFF
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -160,6 +160,9 @@
 -keep class com.chat.base.net.entity.** { *; }
 -keep class com.chat.base.msg.model.** { *; }
 -keep class com.chat.base.msgcontent.** { *; }
+# 频道内搜索 DTO：CursorList/FileHit/MediaHit/MessageHit/CombinedHit/AroundResult/Pagination
+# 走 FastJson 反射按字段名映射 JSON，不 keep 会被 R8 重命名字段导致反序列化产物类型错乱 → checkcast CCE
+-keep class com.chat.base.search.channel.dto.** { *; }
 #----------登录模块---------------
 -keep class com.chat.login.entity.** { *; }
 #----------uikit模块--------------
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -199,6 +199,11 @@
 -keep class com.chat.label.entity.** { *; }
 #----------表情模块------------------
 -keep class com.chat.sticker.entity.** { *; }
+# 自定义贴纸 (uikit) DTO：WKSticker / ListStickerResp / StickerService
+# 走 FastJson 反射按字段名映射 JSON ({"list":[...]})，不 keep 会被 R8 重命名字段
+# 导致 collect / list 反序列化失败 → release 包"添加到我的表情"回调走 onFail
+-keep class com.chat.uikit.chat.sticker.** { *; }
+-keep interface com.chat.uikit.chat.sticker.StickerService { *; }
 #----------客服------------------
 -keep class com.chat.customerservice.entity.** { *; }
 #----------隐私安全------------------
--- a/wkuikit/src/main/java/com/chat/uikit/chat/manager/WKIMUtils.java
+++ b/wkuikit/src/main/java/com/chat/uikit/chat/manager/WKIMUtils.java
@@ -75,6 +75,7 @@ import com.xinbida.wukongim.entity.WKChannel;
 import com.xinbida.wukongim.entity.WKChannelExtras;
 import com.xinbida.wukongim.entity.WKChannelMember;
 import com.xinbida.wukongim.entity.WKChannelStatus;
+import com.xinbida.wukongim.entity.WKSyncExtraMsg;
 import com.xinbida.wukongim.entity.WKChannelType;
 import com.xinbida.wukongim.entity.WKConversationMsg;
 import com.xinbida.wukongim.entity.WKMsg;
@@ -658,6 +659,14 @@ public class WKIMUtils {
                     if (revokeRemind == 1) {
                         if (BuildConfig.DEBUG) android.util.Log.d("RevokeDebug", "[revokeMsg] calling syncExtraMsg for channelID=" + channelID);
                         MsgModel.getInstance().syncExtraMsg(channelID, channelType);
+                        // 竞态修复：服务端撤回操作 + 写 extra 增量 + 推 CMD 三步不是严格原子的，
+                        // CMD 常常先于 extra 增量写入到达客户端。此时立即调 syncExtraMsg 会拿到
+                        // resultSize=0（服务端还没有增量可给），本地永远不知道消息已撤回 →
+                        // 表现为"撤回没反应 / 要撤好几次才成功"（碰运气：偶尔 extra 已就绪则成功）。
+                        // 分别在 500ms / 1500ms 再拉一次，每次先查本地 msg.remoteExtra.revoke
+                        // 是否已 = 1（前一次拉已生效），已生效则跳过（幂等，避免多余请求）。
+                        scheduleRevokeRetry(messageId, channelID, channelType, 1, 500);
+                        scheduleRevokeRetry(messageId, channelID, channelType, 2, 1500);
                     } else {
                         WKMsg wkMsg = WKIM.getInstance().getMsgManager().getWithMessageID(messageId);
                         if (wkMsg != null) {
@@ -679,6 +688,87 @@ public class WKIMUtils {
                 });
             }
         }
+    }
+
+    /**
+     * 撤回 CMD 时序竞态兜底：延迟 {@code delayMs} 后如果本地目标 msg 尚未 mark 撤回，
+     * 再调一次 {@link MsgModel#syncExtraMsg}。查本地状态确保幂等，避免无谓请求。
+     */
+    private void scheduleRevokeRetry(String messageId, String channelID, byte channelType, int attempt, long delayMs) {
+        new Handler(Looper.getMainLooper()).postDelayed(() -> {
+            com.chat.base.utils.WKDbScheduler.get().scheduleDirect(() -> {
+                WKMsg local = WKIM.getInstance().getMsgManager().getWithMessageID(messageId);
+                boolean alreadyRevoked = local != null && local.remoteExtra != null && local.remoteExtra.revoke == 1;
+                if (BuildConfig.DEBUG) android.util.Log.d("RevokeDebug",
+                        "[revokeMsg] retry attempt=" + attempt + " messageId=" + messageId
+                                + " alreadyRevoked=" + alreadyRevoked);
+                if (!alreadyRevoked) {
+                    MsgModel.getInstance().syncExtraMsg(channelID, channelType);
+                }
+            });
+        }, delayMs);
+    }
+
+    /**
+     * 对齐 iOS 自己撤回架构：{@code POST /message/revoke} HTTP 200 后立即本地 mark
+     * 目标消息 {@code remoteExtra.revoke=1, revoker=self.uid}，然后触发 UI 刷新。
+     *
+     * <p>为什么不"欺骗自己"：服务端 HTTP 200 已确认撤回成功，服务端也会推 CMD
+     * 给其它设备 / 其它用户，全局状态一致。本地 mark 只是**同步这个已知事实**到
+     * 当前设备 UI —— 对齐 iOS WKMessageManagerDelegateImp.m:137-138 的做法。
+     *
+     * <p>为什么必须做：服务端 syncMessageExtra 的 cache 逻辑（api.go:1258-1268）
+     * 可能返回 resultSize=0（用户实测 6 次撤回 4 次踩坑），只依赖 CMD → sync 路径
+     * 会永久卡住。iOS 通过"HTTP 200 立即本地 mark"绕过整个 sync 依赖。
+     *
+     * <p>用 clientMsgNo 查本地 msg（比传入 messageId 更稳，处理 messageID 尚未
+     * 从 IM ack 回填的场景）。若本地 msgID 仍为空则跳过（此消息扩展表按 messageID
+     * 索引，无法写入）—— 交给后续 IM ack + CMD 兜底。
+     *
+     * <p>他人撤回场景仍走 CMD → syncExtraMsg 路径（保留 [scheduleRevokeRetry] 兜底）。
+     */
+    public void markMsgRevokedLocallyByClientMsgNO(String clientMsgNo, String revoker) {
+        if (android.text.TextUtils.isEmpty(clientMsgNo)) return;
+        com.chat.base.utils.WKDbScheduler.get().scheduleDirect(() -> {
+            WKMsg wkMsg = WKIM.getInstance().getMsgManager().getWithClientMsgNO(clientMsgNo);
+            if (wkMsg == null) {
+                if (BuildConfig.DEBUG) android.util.Log.w("RevokeDebug",
+                        "[markLocal] msg not found for clientMsgNo=" + clientMsgNo);
+                return;
+            }
+            String messageId = wkMsg.messageID;
+            if (android.text.TextUtils.isEmpty(messageId) || "0".equals(messageId)) {
+                if (BuildConfig.DEBUG) android.util.Log.w("RevokeDebug",
+                        "[markLocal] messageID not yet filled (=" + messageId + ") for clientMsgNo=" + clientMsgNo
+                                + " — skip local mark, rely on CMD retry");
+                return;
+            }
+
+            WKSyncExtraMsg syncExtra = new WKSyncExtraMsg();
+            syncExtra.message_id = messageId;
+            syncExtra.revoke = 1;
+            syncExtra.revoker = revoker;
+            // 保留其它 remoteExtra 字段避免 insertOrReplace 覆盖丢失 (readed/pinned/editedAt 等)。
+            if (wkMsg.remoteExtra != null) {
+                syncExtra.readed = wkMsg.remoteExtra.readed;
+                syncExtra.readed_count = wkMsg.remoteExtra.readedCount;
+                syncExtra.unread_count = wkMsg.remoteExtra.unreadCount;
+                syncExtra.is_mutual_deleted = wkMsg.remoteExtra.isMutualDeleted;
+                syncExtra.is_pinned = wkMsg.remoteExtra.isPinned;
+                syncExtra.extra_version = wkMsg.remoteExtra.extraVersion;
+                syncExtra.edited_at = wkMsg.remoteExtra.editedAt;
+            }
+            List<WKSyncExtraMsg> list = new ArrayList<>();
+            list.add(syncExtra);
+            WKChannel channel = new WKChannel();
+            channel.channelID = wkMsg.channelID;
+            channel.channelType = wkMsg.channelType;
+            WKIM.getInstance().getMsgManager().saveRemoteExtraMsg(channel, list);
+
+            if (BuildConfig.DEBUG) android.util.Log.d("RevokeDebug",
+                    "[markLocal] revoke=1 marked messageID=" + messageId
+                            + " clientMsgNo=" + clientMsgNo + " revoker=" + revoker);
+        });
     }
 
 
--- a/wkuikit/src/main/java/com/chat/uikit/chat/manager/WKIMUtils.java
+++ b/wkuikit/src/main/java/com/chat/uikit/chat/manager/WKIMUtils.java
@@ -738,6 +738,13 @@ public class WKIMUtils {
             }
             String messageId = wkMsg.messageID;
             if (android.text.TextUtils.isEmpty(messageId) || "0".equals(messageId)) {
+                // 已知边界（非主 66% miss 场景）：仅"消息刚发出、IM ack 还没回填 messageID"
+                // 的极小窗口才会命中。此分支跳过本地 mark（extra 表按 messageID 索引，
+                // 没有 messageID 就没法写），实际结果 = 只走服务端 CMD → syncExtraMsg
+                // 兜底。因为 HTTP revoke 本身已用 clientMsgNO 成功（服务端反查），全局
+                // 撤回一致；只是当前设备的 UI 会退回到"等 CMD"这条老路径。
+                // 需要覆盖到这个窗口，需要引入"按 clientMsgNO 索引的 pending revoke 表"，
+                // 待后续独立优化。
                 if (BuildConfig.DEBUG) android.util.Log.w("RevokeDebug",
                         "[markLocal] messageID not yet filled (=" + messageId + ") for clientMsgNo=" + clientMsgNo
                                 + " — skip local mark, rely on CMD retry");
--- a/wkuikit/src/main/java/com/chat/uikit/chat/provider/PdfPreviewActivity.kt
+++ b/wkuikit/src/main/java/com/chat/uikit/chat/provider/PdfPreviewActivity.kt
@@ -21,6 +21,8 @@ import android.graphics.Bitmap
 import android.graphics.Color
 import android.graphics.pdf.PdfRenderer
 import android.os.Bundle
+import android.os.Handler
+import android.os.Looper
 import android.os.ParcelFileDescriptor
 import android.util.DisplayMetrics
 import android.util.TypedValue
@@ -35,19 +37,30 @@ import androidx.recyclerview.widget.RecyclerView
 import com.chat.base.R
 import com.chat.base.utils.WKToastUtils
 import java.io.File
+import java.util.concurrent.ExecutorService
+import java.util.concurrent.Executors
+import kotlin.math.sqrt
 
 /**
  * App 内 PDF 预览。用 [android.graphics.pdf.PdfRenderer]（API 21+ 内置），
  * 每页按屏幕宽度渲染成 Bitmap，RecyclerView 懒加载。
  *
- * 对齐 iOS `WKSafeFilePreviewVC.setupPDFViewInFrame:` 的行为：不走系统外部 App。
+ * 线程模型：`PdfRenderer` 文档要求"所有操作在同一线程"。此处所有 renderer 访问
+ * （open / openPage / render / close）都走 [renderExecutor]（单线程）序列化，
+ * 结果通过 [mainHandler] 回主线程 setImageBitmap。避免主线程解码超大页 → ANR/OOM。
+ *
+ * OOM 防护：大页按 [MAX_BITMAP_PIXELS] 等比缩放；`createBitmap`/`render` 失败
+ * （包括 [OutOfMemoryError]）单页跳过，不影响其它页。
  */
 class PdfPreviewActivity : AppCompatActivity() {
 
     private var filePath: String = ""
-    private var pdfRenderer: PdfRenderer? = null
-    private var fileDescriptor: ParcelFileDescriptor? = null
+    @Volatile private var pdfRenderer: PdfRenderer? = null
+    @Volatile private var fileDescriptor: ParcelFileDescriptor? = null
+    @Volatile private var pageCount: Int = 0
     private var pageWidthPx: Int = 0
+    private val renderExecutor: ExecutorService = Executors.newSingleThreadExecutor()
+    private val mainHandler = Handler(Looper.getMainLooper())
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
@@ -70,23 +83,32 @@ class PdfPreviewActivity : AppCompatActivity() {
         }
         setContentView(recyclerView)
 
-        val opened = openPdf()
-        if (!opened) {
-            WKToastUtils.getInstance().showToastNormal(getString(R.string.str_file_not_exist))
-            finish()
-            return
+        renderExecutor.execute {
+            val ok = openPdfOnExecutor()
+            mainHandler.post {
+                if (isFinishing || isDestroyed) return@post
+                if (!ok || pdfRenderer == null || pageCount <= 0) {
+                    WKToastUtils.getInstance().showToastNormal(getString(R.string.str_file_not_exist))
+                    finish()
+                    return@post
+                }
+                recyclerView.adapter = PdfPageAdapter(
+                    pdfRenderer!!, pageCount, pageWidthPx, renderExecutor, mainHandler
+                )
+            }
         }
-        val renderer = pdfRenderer ?: return
-        recyclerView.adapter = PdfPageAdapter(renderer, pageWidthPx)
     }
 
-    private fun openPdf(): Boolean {
+    /** 必须在 [renderExecutor] 上执行——PdfRenderer 线程绑定。 */
+    private fun openPdfOnExecutor(): Boolean {
         val file = File(filePath)
         if (!file.exists()) return false
         return try {
             val fd = ParcelFileDescriptor.open(file, ParcelFileDescriptor.MODE_READ_ONLY)
             fileDescriptor = fd
-            pdfRenderer = PdfRenderer(fd)
+            val renderer = PdfRenderer(fd)
+            pdfRenderer = renderer
+            pageCount = renderer.pageCount
             true
         } catch (_: Exception) {
             false
@@ -147,16 +169,30 @@ class PdfPreviewActivity : AppCompatActivity() {
 
     override fun onDestroy() {
         super.onDestroy()
-        try { pdfRenderer?.close() } catch (_: Exception) {}
-        try { fileDescriptor?.close() } catch (_: Exception) {}
+        // renderer/fd 必须在 renderExecutor 上关闭（线程绑定），提交后再 shutdown。
+        val r = pdfRenderer
+        val fd = fileDescriptor
+        pdfRenderer = null
+        fileDescriptor = null
+        renderExecutor.execute {
+            try { r?.close() } catch (_: Exception) {}
+            try { fd?.close() } catch (_: Exception) {}
+        }
+        renderExecutor.shutdown()
     }
 
     private class PdfPageAdapter(
         private val renderer: PdfRenderer,
+        private val pageCount: Int,
         private val pageWidthPx: Int,
+        private val renderExecutor: ExecutorService,
+        private val mainHandler: Handler,
     ) : RecyclerView.Adapter<PdfPageAdapter.VH>() {
 
-        class VH(val imageView: ImageView) : RecyclerView.ViewHolder(imageView)
+        class VH(val imageView: ImageView) : RecyclerView.ViewHolder(imageView) {
+            /** 当前请求的页码；异步 render 完成时用于比对是否已被回收/复用。 */
+            @Volatile var boundPage: Int = -1
+        }
 
         override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): VH {
             val marginPx = TypedValue.applyDimension(
@@ -174,29 +210,71 @@ class PdfPreviewActivity : AppCompatActivity() {
         }
 
         override fun onBindViewHolder(holder: VH, position: Int) {
-            val page = try { renderer.openPage(position) } catch (_: Exception) { return }
-            try {
-                val ratio = page.height.toFloat() / page.width.toFloat()
-                val h = (pageWidthPx * ratio).toInt().coerceAtLeast(1)
-                val bmp = Bitmap.createBitmap(pageWidthPx, h, Bitmap.Config.ARGB_8888)
-                bmp.eraseColor(Color.WHITE)
-                page.render(bmp, null, null, PdfRenderer.Page.RENDER_MODE_FOR_DISPLAY)
-                holder.imageView.setImageBitmap(bmp)
+            holder.boundPage = position
+            holder.imageView.setImageDrawable(null)
+            renderExecutor.execute {
+                val bmp = try {
+                    renderPage(position)
+                } catch (_: Exception) {
+                    null
+                }
+                mainHandler.post {
+                    if (holder.boundPage == position && bmp != null) {
+                        holder.imageView.setImageBitmap(bmp)
+                    }
+                    // 若 holder 已被复用/回收，丢弃这张 bitmap（不主动 recycle，让 GC 处理，
+                    // 避免"trying to use a recycled bitmap"在硬件加速 display list 里踩到）。
+                }
+            }
+        }
+
+        /** 在 [renderExecutor] 上执行——PdfRenderer 线程绑定。 */
+        private fun renderPage(position: Int): Bitmap? {
+            val page = try { renderer.openPage(position) } catch (_: Exception) { return null }
+            return try {
+                val ratio = page.height.toFloat() / page.width.toFloat().coerceAtLeast(1f)
+                val h0 = (pageWidthPx * ratio).toInt().coerceAtLeast(1)
+                val (w, h) = clampToMaxPixels(pageWidthPx, h0)
+                // Bitmap.createBitmap 在极端尺寸下会抛 OutOfMemoryError，单页失败不影响
+                // 其它页的渲染——所以 catch 掉返 null。
+                try {
+                    val bitmap = Bitmap.createBitmap(w, h, Bitmap.Config.ARGB_8888)
+                    bitmap.eraseColor(Color.WHITE)
+                    page.render(bitmap, null, null, PdfRenderer.Page.RENDER_MODE_FOR_DISPLAY)
+                    bitmap
+                } catch (_: OutOfMemoryError) {
+                    null
+                } catch (_: Exception) {
+                    null
+                }
             } finally {
                 try { page.close() } catch (_: Exception) {}
             }
         }
 
+        /** 按最大总像素数等比缩放，避免 `createBitmap` 分配几百 MB。 */
+        private fun clampToMaxPixels(w: Int, h: Int): Pair<Int, Int> {
+            val total = w.toLong() * h.toLong()
+            if (total <= MAX_BITMAP_PIXELS) return w to h
+            val scale = sqrt(MAX_BITMAP_PIXELS.toDouble() / total.toDouble())
+            return (w * scale).toInt().coerceAtLeast(1) to (h * scale).toInt().coerceAtLeast(1)
+        }
+
         override fun onViewRecycled(holder: VH) {
-            (holder.imageView.drawable as? android.graphics.drawable.BitmapDrawable)?.bitmap?.recycle()
+            holder.boundPage = -1
+            // 不主动 recycle bitmap：可能仍在 hardware display list 里被引用。
+            // 清除 drawable 引用后交给 GC 回收更安全。
             holder.imageView.setImageDrawable(null)
         }
 
-        override fun getItemCount(): Int = renderer.pageCount
+        override fun getItemCount(): Int = pageCount
     }
 
     companion object {
         private const val MENU_SAVE = 1001
         private const val MENU_SHARE = 1002
+
+        /** 单张 bitmap 允许的最大总像素数。8M px * 4B/px (ARGB_8888) = 32 MB。 */
+        private const val MAX_BITMAP_PIXELS: Long = 8_000_000L
     }
 }
--- a/wkuikit/src/main/java/com/chat/uikit/chat/provider/PdfPreviewActivity.kt
+++ b/wkuikit/src/main/java/com/chat/uikit/chat/provider/PdfPreviewActivity.kt
@@ -1,0 +1,202 @@
+/*
+ * Copyright 2026-present OctoIM contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.chat.uikit.chat.provider
+
+import android.content.Intent
+import android.graphics.Bitmap
+import android.graphics.Color
+import android.graphics.pdf.PdfRenderer
+import android.os.Bundle
+import android.os.ParcelFileDescriptor
+import android.util.DisplayMetrics
+import android.util.TypedValue
+import android.view.Menu
+import android.view.MenuItem
+import android.view.ViewGroup
+import android.widget.ImageView
+import androidx.appcompat.app.AppCompatActivity
+import androidx.core.content.FileProvider
+import androidx.recyclerview.widget.LinearLayoutManager
+import androidx.recyclerview.widget.RecyclerView
+import com.chat.base.R
+import com.chat.base.utils.WKToastUtils
+import java.io.File
+
+/**
+ * App 内 PDF 预览。用 [android.graphics.pdf.PdfRenderer]（API 21+ 内置），
+ * 每页按屏幕宽度渲染成 Bitmap，RecyclerView 懒加载。
+ *
+ * 对齐 iOS `WKSafeFilePreviewVC.setupPDFViewInFrame:` 的行为：不走系统外部 App。
+ */
+class PdfPreviewActivity : AppCompatActivity() {
+
+    private var filePath: String = ""
+    private var pdfRenderer: PdfRenderer? = null
+    private var fileDescriptor: ParcelFileDescriptor? = null
+    private var pageWidthPx: Int = 0
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+
+        val title = intent.getStringExtra("title") ?: "PDF"
+        filePath = intent.getStringExtra("filePath") ?: ""
+
+        supportActionBar?.apply {
+            setDisplayHomeAsUpEnabled(true)
+            this.title = title
+        }
+
+        val metrics = DisplayMetrics()
+        @Suppress("DEPRECATION") windowManager.defaultDisplay.getMetrics(metrics)
+        pageWidthPx = metrics.widthPixels
+
+        val recyclerView = RecyclerView(this).apply {
+            layoutManager = LinearLayoutManager(this@PdfPreviewActivity)
+            setBackgroundColor(Color.parseColor("#EFEFEF"))
+        }
+        setContentView(recyclerView)
+
+        val opened = openPdf()
+        if (!opened) {
+            WKToastUtils.getInstance().showToastNormal(getString(R.string.str_file_not_exist))
+            finish()
+            return
+        }
+        val renderer = pdfRenderer ?: return
+        recyclerView.adapter = PdfPageAdapter(renderer, pageWidthPx)
+    }
+
+    private fun openPdf(): Boolean {
+        val file = File(filePath)
+        if (!file.exists()) return false
+        return try {
+            val fd = ParcelFileDescriptor.open(file, ParcelFileDescriptor.MODE_READ_ONLY)
+            fileDescriptor = fd
+            pdfRenderer = PdfRenderer(fd)
+            true
+        } catch (_: Exception) {
+            false
+        }
+    }
+
+    override fun onCreateOptionsMenu(menu: Menu): Boolean {
+        menu.add(0, MENU_SAVE, 0, R.string.str_file_save_to)
+            .setShowAsAction(MenuItem.SHOW_AS_ACTION_NEVER)
+        menu.add(0, MENU_SHARE, 1, R.string.str_file_share)
+            .setShowAsAction(MenuItem.SHOW_AS_ACTION_NEVER)
+        return true
+    }
+
+    override fun onOptionsItemSelected(item: MenuItem): Boolean {
+        return when (item.itemId) {
+            android.R.id.home -> {
+                finish()
+                true
+            }
+            MENU_SAVE -> {
+                saveFile()
+                true
+            }
+            MENU_SHARE -> {
+                shareFile()
+                true
+            }
+            else -> super.onOptionsItemSelected(item)
+        }
+    }
+
+    private fun saveFile() {
+        if (filePath.isEmpty()) return
+        val file = File(filePath)
+        if (!file.exists()) return
+        val intent = Intent(this, FileSaveActivity::class.java)
+        intent.putExtra("sourceFilePath", file.absolutePath)
+        intent.putExtra("fileName", file.name)
+        startActivity(intent)
+    }
+
+    private fun shareFile() {
+        if (filePath.isEmpty()) return
+        val file = File(filePath)
+        if (!file.exists()) return
+        try {
+            val uri = FileProvider.getUriForFile(this, "$packageName.fileProvider", file)
+            val shareIntent = Intent(Intent.ACTION_SEND).apply {
+                type = WKFileProvider.getMimeType(file.name)
+                putExtra(Intent.EXTRA_STREAM, uri)
+                addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION)
+            }
+            startActivity(Intent.createChooser(shareIntent, file.name))
+        } catch (_: Exception) {
+        }
+    }
+
+    override fun onDestroy() {
+        super.onDestroy()
+        try { pdfRenderer?.close() } catch (_: Exception) {}
+        try { fileDescriptor?.close() } catch (_: Exception) {}
+    }
+
+    private class PdfPageAdapter(
+        private val renderer: PdfRenderer,
+        private val pageWidthPx: Int,
+    ) : RecyclerView.Adapter<PdfPageAdapter.VH>() {
+
+        class VH(val imageView: ImageView) : RecyclerView.ViewHolder(imageView)
+
+        override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): VH {
+            val marginPx = TypedValue.applyDimension(
+                TypedValue.COMPLEX_UNIT_DIP, 8f, parent.resources.displayMetrics
+            ).toInt()
+            val iv = ImageView(parent.context).apply {
+                layoutParams = RecyclerView.LayoutParams(
+                    ViewGroup.LayoutParams.MATCH_PARENT,
+                    ViewGroup.LayoutParams.WRAP_CONTENT,
+                ).apply { setMargins(marginPx, marginPx, marginPx, marginPx) }
+                adjustViewBounds = true
+                setBackgroundColor(Color.WHITE)
+            }
+            return VH(iv)
+        }
+
+        override fun onBindViewHolder(holder: VH, position: Int) {
+            val page = try { renderer.openPage(position) } catch (_: Exception) { return }
+            try {
+                val ratio = page.height.toFloat() / page.width.toFloat()
+                val h = (pageWidthPx * ratio).toInt().coerceAtLeast(1)
+                val bmp = Bitmap.createBitmap(pageWidthPx, h, Bitmap.Config.ARGB_8888)
+                bmp.eraseColor(Color.WHITE)
+                page.render(bmp, null, null, PdfRenderer.Page.RENDER_MODE_FOR_DISPLAY)
+                holder.imageView.setImageBitmap(bmp)
+            } finally {
+                try { page.close() } catch (_: Exception) {}
+            }
+        }
+
+        override fun onViewRecycled(holder: VH) {
+            (holder.imageView.drawable as? android.graphics.drawable.BitmapDrawable)?.bitmap?.recycle()
+            holder.imageView.setImageDrawable(null)
+        }
+
+        override fun getItemCount(): Int = renderer.pageCount
+    }
+
+    companion object {
+        private const val MENU_SAVE = 1001
+        private const val MENU_SHARE = 1002
+    }
+}
--- a/wkuikit/src/main/java/com/chat/uikit/chat/provider/TextPreviewActivity.kt
+++ b/wkuikit/src/main/java/com/chat/uikit/chat/provider/TextPreviewActivity.kt
@@ -18,6 +18,8 @@ package com.chat.uikit.chat.provider
 
 import android.content.Intent
 import android.os.Bundle
+import android.os.Handler
+import android.os.Looper
 import android.view.Menu
 import android.view.MenuItem
 import android.widget.ScrollView
@@ -26,10 +28,14 @@ import androidx.appcompat.app.AppCompatActivity
 import androidx.core.content.FileProvider
 import com.chat.base.R
 import java.io.File
+import java.util.concurrent.ExecutorService
+import java.util.concurrent.Executors
 
 class TextPreviewActivity : AppCompatActivity() {
 
     private var filePath: String = ""
+    private val loadExecutor: ExecutorService = Executors.newSingleThreadExecutor()
+    private val mainHandler = Handler(Looper.getMainLooper())
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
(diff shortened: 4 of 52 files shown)


Jerry-Xin (dismissed):
Summary: The PR is in scope for this Android chat app, but the new xlsx preview parser needs a safety fix before merge.

**Blocking**
- 🔴 Critical: `XlsxParser` trusts worksheet cell coordinates before validating them. In [XlsxParser.kt](/Users/mlamp/.openclaw/tmp/pr-review-qud92k81/repo/wkuikit/src/main/java/com/chat/uikit/chat/provider/XlsxParser.kt:172), `columnIndex(cellRef)` can return `-1` for cells without an `r` attribute, then line 176 calls `fillTo(...)` and line 177 writes `currentRow[-1]`, causing preview failure for otherwise parseable workbooks. The same path calls `fillTo` before checking `MAX_COLS`, so malformed chat attachments with extreme column refs can force large allocations before being marked truncated. Also, shared strings are read with no count/size cap in [XlsxParser.kt](/Users/mlamp/.openclaw/tmp/pr-review-qud92k81/repo/wkuikit/src/main/java/com/chat/uikit/chat/provider/XlsxParser.kt:74). Fix by rejecting `col < 0`, checking `col >= MAX_COLS` before `fillTo`, and adding bounded parsing limits for shared strings/cell text.

**Non-blocking**
- 🟡 Warning: Text preview still reads the full file into memory before truncating at 50k chars in [TextPreviewActivity.kt](/Users/mlamp/.openclaw/tmp/pr-review-qud92k81/repo/wkuikit/src/main/java/com/chat/uikit/chat/provider/TextPreviewActivity.kt:128) and [WKFileProvider.kt](/Users/mlamp/.openclaw/tmp/pr-review-qud92k81/repo/wkuikit/src/main/java/com/chat/uikit/chat/provider/WKFileProvider.kt:339). A bounded streaming read would avoid UI stalls or OOM on large text attachments.
- 🔵 Suggestion: `XlsxPreviewActivity` has several hardcoded user-visible strings in [XlsxPreviewActivity.kt](/Users/mlamp/.openclaw/tmp/pr-review-qud92k81/repo/wkuikit/src/main/java/com/chat/uikit/chat/provider/XlsxPreviewActivity.kt:114) and [XlsxPreviewActivity.kt](/Users/mlamp/.openclaw/tmp/pr-review-qud92k81/repo/wkuikit/src/main/java/com/chat/uikit/chat/provider/XlsxPreviewActivity.kt:220). Move them into resources for localization consistency.

**Highlights**
- ✅ The PR is relevant to `Mininglamp-OSS/octo-android`: it extends chat messages, file previews, stickers, and related UI.
- ✅ Sticker message registration and provider separation align with the existing content-type/provider architecture.
- ✅ The added R8 keep rules and focused unit tests cover several of the new sticker serialization/order/validation utilities.

Verification note: I attempted `./gradlew testDebugUnitTest --no-daemon`, but this environment has no Java runtime installed, so I could not run the build or tests locally.

wkuikit/src/main/java/com/chat/uikit/chat/provider/XlsxParser.kt
```diff
@@ -0,0 +1,240 @@
...
+                        cellValue = (cellValue ?: "") + parser.text
+                    }
+                    XmlPullParser.END_TAG -> when (parser.name) {
+                        "v" -> inV = false
+                        "c" -> {
+                            if (currentRow != null && rows.size < MAX_ROWS) {
+                                val col = columnIndex(cellRef)
+                                val text = resolveCellText(cellType, cellValue, inlineText, shared)
+                                fillTo(currentRow!!, col)
+                                if (col < MAX_COLS) currentRow!![col] = text else truncated = true
```
OctoBoooot: **[major]** A cell `<c>` with no `r` attribute makes `columnIndex()` return `-1` (see :208-215), and `currentRow!![-1] = text` then throws `IndexOutOfBoundsException`, aborting the whole sheet parse. The `r` attribute is optional in OOXML — some legitimate producers omit it — so this breaks preview for valid files, not just malicious ones. Guard `col >= 0` before the assignment (skip or clamp the cell).

wkuikit/src/main/java/com/chat/uikit/chat/provider/XlsxParser.kt
```diff
@@ -0,0 +1,240 @@
...
+                readSheet(zip, entry, shared, name)
+            }
+            return Workbook(sheets)
+        }
+    }
+
+    // ---- sharedStrings.xml ----
+    private fun readSharedStrings(zip: ZipFile, entry: ZipEntry): List<String> {
+        val out = ArrayList<String>()
+        zip.getInputStream(entry).use { input ->
```
OctoBoooot: **[major]** Zip-bomb / decompression DoS: entries are streamed via `zip.getInputStream(entry)` with no cap on decompressed bytes, and `readSharedStrings` accumulates every `<si>` into an unbounded list. A few-KB `xl/sharedStrings.xml` (or a sheet) that decompresses to GBs of repeated `<si><t>A</t></si>` is fully walked → OOM. `MAX_ROWS`/`MAX_COLS` bound output, not input consumed. Wrap the entry stream in a byte-limited reader and cap the shared-string count.

wkuikit/src/main/java/com/chat/uikit/chat/provider/PdfPreviewActivity.kt
```diff
@@ -0,0 +1,202 @@
...
+            }
+            return VH(iv)
+        }
+
+        override fun onBindViewHolder(holder: VH, position: Int) {
+            val page = try { renderer.openPage(position) } catch (_: Exception) { return }
+            try {
+                val ratio = page.height.toFloat() / page.width.toFloat()
+                val h = (pageWidthPx * ratio).toInt().coerceAtLeast(1)
+                val bmp = Bitmap.createBitmap(pageWidthPx, h, Bitmap.Config.ARGB_8888)
```
OctoBoooot: **[major]** `Bitmap.createBitmap(pageWidthPx, h, ARGB_8888)` with `h = pageWidthPx * (page.height/page.width)`: a degenerate PDF page (e.g. 1×100000 pt) yields an `h` in the millions → multi-GB allocation → uncaught `OutOfMemoryError` (the surrounding `try` has only a `finally`, no catch). `coerceAtLeast(1)` guards the zero case, not the huge case. Clamp total pixels (cap ratio) and wrap the createBitmap+render block in try/catch to skip the page gracefully.

wkuikit/src/main/java/com/chat/uikit/chat/provider/TextPreviewActivity.kt
```diff
@@ -117,5 +121,20 @@ class TextPreviewActivity : AppCompatActivity() {
     companion object {
         private const val MENU_SAVE = 1001
         private const val MENU_SHARE = 1002
+
+        /** 与 WKFileProvider.TEXT_PREVIEW_EXTS 对齐的截断阈值（字符数）。 */
+        private const val MAX_CHARS = 50_000
+
+        private fun readFromFilePath(filePath: String): String {
+            if (filePath.isEmpty()) return ""
+            return try {
```
OctoBoooot: **[minor]** `readText()` loads the entire file into memory on the UI thread before truncating to `MAX_CHARS`; a large `.log`/`.txt` → ANR + OOM, defeating the cap. The xlsx path correctly uses a background executor — mirror that here and stream only up to `MAX_CHARS`. Same pattern in `WKFileProvider.openFileDirectlyStatic`.

wkuikit/src/main/java/com/chat/uikit/chat/manager/WKIMUtils.java
```diff
@@ -681,6 +690,87 @@ public void revokeMsg(JSONObject jsonObject) {
...
+            if (wkMsg == null) {
+                if (BuildConfig.DEBUG) android.util.Log.w("RevokeDebug",
+                        "[markLocal] msg not found for clientMsgNo=" + clientMsgNo);
+                return;
+            }
+            String messageId = wkMsg.messageID;
+            if (android.text.TextUtils.isEmpty(messageId) || "0".equals(messageId)) {
+                if (BuildConfig.DEBUG) android.util.Log.w("RevokeDebug",
+                        "[markLocal] messageID not yet filled (=" + messageId + ") for clientMsgNo=" + clientMsgNo
+                                + " — skip local mark, rely on CMD retry");
```
OctoBoooot, line 750: **[question]** When `messageID` is empty or `"0"`, this skips the local revoke-mark and returns — but the PR body says "messageID 空/'0' 时传空串让服务端反查". For a just-sent message whose IM ack hasn't backfilled `messageID`, the local-mark path is skipped and the flow falls back to the CMD path that issue #87 says misses ~66% of the time. Does `MsgModel.revokeMsg` (out of diff) actually perform the empty-string server reverse-lookup? If not, the un-acked-message window #87 describes isn't covered by the local-mark fix.

app/proguard-rules.pro
```diff
@@ -160,6 +160,9 @@
 -keep class com.chat.base.net.entity.** { *; }
 -keep class com.chat.base.msg.model.** { *; }
 -keep class com.chat.base.msgcontent.** { *; }
```
OctoBoooot, line 162: **[question]** The keep `com.chat.base.search.channel.dto.**` correctly covers the PR #83 DTOs and fixes the release ClassCastException. But siblings in `com.chat.base.search.channel.*` (the Retrofit `ChannelSearchService` interface, `MessageHitMapper`, error parsers) aren't kept. Confirm the release CCE was only in `dto` and no reflectively-accessed class outside that subpackage still needs a keep rule.

wkuikit/src/main/java/com/chat/uikit/chat/provider/WKFileProvider.kt
```diff
@@ -152,7 +152,37 @@ class WKFileProvider : WKChatBaseProvider() {
         // Extract basename to prevent path traversal (../ or /)
         val basename = File(name).name
         // Remove any remaining dangerous characters
-        return basename.replace(Regex("[\\\\/:*?\"<>|]"), "_")
+        val cleaned = basename.replace(Regex("[\\\\/:*?\"<>|]"), "_")
+        // Linux 文件系统 NAME_MAX = 255 字节 (UTF-8, 不是字符)。
+        // 长中文/emoji 文件名（3~4 字节/字符）很容易越界，导致 FileOutputStream 抛
+        // ENAMETOOLONG 静默失败，用户观感是"点了没反应"。留 55 字节余量。
+        return truncateToByteLength(cleaned, MAX_FILENAME_BYTES)
+    }
```
OctoBoooot, line 160: **[praise]** `truncateToByteLength` truncates by UTF-8 byte length on code-point boundaries while preserving the extension — a real fix for the `ENAMETOOLONG` download failure on long CJK/emoji names — and `sanitizeFileName` (basename extraction + separator strip) closes path traversal via the message's file name. Well done.

OctoBoooot (changes requested):
## Review: feat(chat): 批量改进 — 贴纸/预览/名称长度/撤回/release 修复 (#92)

**Verdict:** Request changes.
CI: `check-sprint` is red (governance — but the body already has `Closes #87..#91`, so this should re-run green; not a code blocker on its own) and **`build` has not completed** — an APPROVE isn't possible until build is green regardless. The substantive reason for REQUEST_CHANGES is the majors below, all in the new file-preview parsers (untrusted-input crash/DoS) plus one test-debt gap on the sticker feature's headline fix.

This is a large 50-file / 5-concern PR. Reviewed by concern (sticker / file-preview / recall+name-length+proguard), each finding verified against source at 652f35a. Overall the code quality is high and the security instincts are good — the findings are edge-case robustness on parsers fed by attacker-controllable files.

### Major
- `XlsxParser.kt:174-177` — a cell `<c>` with no `r` attribute makes `columnIndex()` return `-1`, then `currentRow!![-1]` throws `IndexOutOfBoundsException` and aborts the whole sheet parse. `r` is optional in OOXML; some legitimate producers omit it, so this isn't only a malicious-file case — those files fail to preview. Guard `col >= 0` before the assignment.
- `XlsxParser.kt:74-93,135-189` — zip-bomb / decompression DoS. Entries are streamed via `zip.getInputStream(entry)` with no cap on *decompressed* bytes; `readSharedStrings` accumulates every `<si>` into an unbounded list. A ~KB-compressed `xl/sharedStrings.xml` expanding to GBs of `<si><t>A</t></si>` → OOM on preview. `MAX_ROWS/MAX_COLS` bound output, not input consumed. Wrap the entry stream in a byte-limited reader and cap shared-string count.
- `PdfPreviewActivity.kt:180-182` — `Bitmap.createBitmap(pageWidthPx, h, ARGB_8888)` with `h = pageWidthPx * (page.height/page.width)`; a degenerate PDF page (e.g. 1×100000 pt) yields a multi-GB allocation → uncaught `OutOfMemoryError` (the `try` around it only has a `finally`, no catch). Clamp total pixels and wrap the render block in try/catch to skip the page.
- `WKStickerManager.reorder()` (sticker feature) — the PR's headline design decision ("reorder sorts params directly, avoiding the SP `apply()` read-back race") is **correct as written but has no test**. `StickerLocalOrderStoreTest` covers only the pure `applyOrder`/`pruneOrder`; reverting `reorder` to a read-back implementation would pass every existing test. Add a test asserting `reorder(ids)` emits `ids`-order with no SP round-trip — this is exactly the flagged-risk path.

### Minor
- `TextPreviewActivity.kt:128-138` (and `WKFileProvider.openFileDirectlyStatic`) — `readText()` loads the entire file into memory on the UI thread *before* truncating to `MAX_CHARS`; a 500 MB `.log` → ANR + OOM, defeating the cap. The xlsx path correctly uses a background executor; mirror that and stream up to `MAX_CHARS`.
- `XlsxParser` — `readSharedStrings` has no count cap (distinct from the sheet's `MAX_ROWS`); millions of distinct strings OOM before any sheet is read.
- `WKFileProvider.kt:209-214` — two downloads differing only past byte-200 truncate to the same filename; the `exists()` short-circuit then serves the first file's stale content for the second. Incorporate `clientMsgNO`/hash into the truncated name.
- `WKIMUtils.java:660` — on a self-revoke, the first `syncExtraMsg` fires unconditionally (only the two retries are gated by `alreadyRevoked`); idempotent, but one redundant network sync per own-revoke.
- `StringUtils.attachLengthLimit` — `setFilters(new InputFilter[]{...})` replaces all existing filters; any caller relying on a coexisting filter (emoji-block, XML `maxLength`) loses it silently. Current 3 callers look clean; worth a doc note.

### Questions
- `WKIMUtils.java:743-746` — when `messageID` is empty/`"0"`, `markMsgRevokedLocallyByClientMsgNO` **skips the local mark and returns**, which contradicts the PR body's "messageID 空/'0' 时传空串让服务端反查." For a just-sent message whose IM ack hasn't backfilled `messageID`, the local-mark path is skipped and the user falls back to the exact unreliable CMD path issue #87 is fixing (~66% miss). Does `MsgModel.revokeMsg` (out of diff) actually do the empty-string server reverse-lookup? If not, the un-acked window #87 describes isn't covered.
- `app/proguard-rules.pro` — the keep covers `com.chat.base.search.channel.dto.**` (correct — matches the PR #83 DTOs and fixes the CCE). But siblings `com.chat.base.search.channel.*` (`ChannelSearchService` Retrofit interface, `MessageHitMapper`, error parsers) aren't kept. Confirm the release CCE was truly only in `dto` and no reflectively-accessed class outside it still needs keeping.

### Praise
- **File-preview security posture is genuinely well-built (verified):** WebView is hardened — `javaScriptEnabled=false`, `allowFileAccess=false`, `allowContentAccess=false`, `loadDataWithBaseURL(null, …)` (opaque origin), and cell contents HTML-escaped (`& < > " '`). XXE/billion-laughs are closed by the `Xml.newPullParser()` (KXmlParser, `FEATURE_PROCESS_DOCDECL=false`) choice — external/DTD entities aren't resolved. Zip-entry lookup uses `ZipFile.getEntry` (in-archive, no disk extraction) so Zip-Slip doesn't apply. Preview activities are correctly non-exported.
- `WKFileProvider.truncateToByteLength` — truncates by UTF-8 *byte* length on code-point boundaries preserving the extension; genuinely fixes the `ENAMETOOLONG` failure for long CJK/emoji names, and `sanitizeFileName` (basename + strip separators) closes path traversal via `fileContent.name`.
- `StringUtils` length limit handles surrogate pairs correctly (decrements `keep` off a high surrogate so an emoji is dropped whole, never split).
- Sticker contentType 12/13 deserialization is crash-safe (`optString/optInt` throughout, `as?` casts + null/URL guards in every provider) — a malformed remote sticker message can't NPE the provider. Upload validator uses magic-bytes (not extension) with size/dimension bounds. All three sticker unit-test suites are mutation-resistant (reverting the impl fails a test).

- Sticker `content.url` from a remote message flows to Glide via `getShowUrl` (unchanged here) with no scheme allow-list visible — a hostile sticker could point `url` at an attacker host (render-time beacon) or `file://`. Pre-existing image-path exposure, not introduced by this PR, but this feature is a new untrusted-URL sink; worth confirming Glide restricts schemes.

_Note: verdict rests on the majors; even with them fixed, APPROVE requires the `build` check to complete green and `check-sprint` to re-run (issue links are present)._

yujiawei (changes requested):
# Code Review — PR #92 (octo-android)

Reviewed at head SHA `652f35a3`. Base `main`, merge-base `e183a8d8`. 50 files, +3722/−50, 5 feature groups (revoke fix, name-length throttle, release proguard keeps, in-app file preview, custom stickers). Cross-file symbols were verified against the wukongim SDK and wkbase.

This review is organized in two stages: spec compliance first, then code quality.

## 1. Spec compliance

**Spec: ✅**

All five linked issues are implemented as described, and I found no functionality built beyond the stated scope:

- **#87 revoke** — `WKChatBaseProvider` now sends `clientMsgNO` as the authority and passes an empty `message_id` when `messageID` is empty/`"0"`; `WKUIKitApplication` marks locally on HTTP 200; `WKIMUtils.markMsgRevokedLocallyByClientMsgNO` + `scheduleRevokeRetry` implement the dual path. `saveRemoteExtraMsg`, `getWithClientMsgNO`, and every `WKSyncExtraMsg` field written (`revoke`, `revoker`, `readed`, `is_mutual_deleted`, `is_pinned`, `extra_version`, `edited_at`, …) exist in the SDK — no field loss on `insertOrReplaceExtra`.
- **#88 name length** — `ThrottledRunnable` + `StringUtils.attachLengthLimit()` consolidates the length toast; call sites (group name 30→50, thread name 50→100) updated consistently in both code and XML `maxLength`.
- **#89 / release CCE** — proguard `-keep` added for `com.chat.base.search.channel.dto.**`.
- **#90 file preview** — `PdfPreviewActivity` / `XlsxPreviewActivity` / `XlsxParser` added; `WKFileProvider` gains a Preview button gated by extension; `truncateToByteLength` fixes long-filename `ENAMETOOLONG`.
- **#91 stickers** — content types 12/13 registered in `WKUIKitApplication` (prevents `unknown_msg` fallback), providers, collect/upload/delete/reorder, local order store, 38 unit tests.

Minor note (not blocking): the `chore` commit adds agent-tooling paths (`.claude/`, `AGENTS.md`, `/app/prod/`, …) to `.gitignore`. Harmless, but unrelated to the chat scope of this PR.

## 2. Code quality

**Quality: Changes-Requested**

The happy paths are solid and well-tested. The blocking items are a focused cluster in the file-preview feature: large or crafted files are read/rendered without bounding on the wrong thread. The fix pattern (async + bounded) already exists in this same PR (`XlsxPreviewActivity` parses off the main thread), so these are targeted fixes, not a redesign.

### P1 — PDF pages rendered on the main thread
`PdfPreviewActivity.kt:176-188` — `renderer.openPage()`, `Bitmap.createBitmap(pageWidthPx, h, ARGB_8888)`, and `page.render()` all run inside `onBindViewHolder`, i.e. on the main thread during scroll. For ordinary multi-page or high-DPI PDFs this drops frames and can ANR; a page with an extreme height ratio makes `createBitmap` allocate a very large bitmap and OOM on the main thread (uncaught). Suggest rendering on a background executor (as `XlsxPreviewActivity` already does) and posting the bitmap back, and clamping bitmap height.

### P1 — XlsxParser allocates before the column bound check
`XlsxParser.kt:174-177` — `fillTo(currentRow!!, col)` runs before the `if (col < MAX_COLS)` guard on the next line. `columnIndex()` parses an arbitrary `r` attribute from the file (attacker-controlled — preview operates on chat attachments); a ref like `AAAAAAAAAA1` yields a column index of ~26^10, so `fillTo` loops `while (list.size <= index) list.add("")` and OOMs. It is wrapped in `runCatching` on a background thread, but OOM recovery is unreliable and the runaway allocation is itself a freeze/DoS. Fix: guard `fillTo` with the same `col < MAX_COLS` (also handle `columnIndex` returning `-1` for a null ref, which would otherwise index `[-1]`).

### P1 — Text preview reads the whole file before truncating, on the main thread
`TextPreviewActivity.kt:130-137` (`readFromFilePath`) and `WKFileProvider.kt:335-...` (`openFileDirectlyStatic`) call `File.readText(UTF-8)` and *then* substring to 50k. Text sends are allowed up to 100MB, so previewing a large `.log`/`.csv` loads the entire file into a `String` in `onCreate` on the main thread → OOM/ANR. This mirrors pre-existing behavior on the "Open" path, but the new Preview button makes it reachable for the same extensions, so the surface is expanded. Fix: bounded read (read only up to `MAX_CHARS`), ideally off the main thread.

### P2 — advisory
- `PdfPreviewActivity.kt:190-192` — `onViewRecycled` calls `bitmap?.recycle()`; recycling a bitmap that may still be in a display list can trigger `Canvas: trying to use a recycled bitmap` on hardware-accelerated devices. Prefer just clearing the drawable and letting GC reclaim it.
- `WKIMUtils.java:693-705` — `scheduleRevokeRetry` posts delayed DB work up to 1.5s out with no lifecycle/cancellation guard; if the app is backgrounded or the user logs out, the runnable still fires. Low impact (it null-checks the local msg), but consider tying it to a lifecycle or cancel token.
- `ChatPanelManager.kt:2595` — `WKStickerManager.stickersLiveData.observe(activity)` is registered inside `buildStickerContentView()`; if `getEmojiLayout()` is rebuilt within one activity lifecycle, observers accumulate (LiveData does not dedup). Bounded today (called once during toolbar init), but worth guarding if toolbar re-init is ever introduced.

## Coverage
- **Verified:** revoke cross-file symbols + `WKSyncExtraMsg` field completeness against the SDK; sticker network callbacks run on the main thread (`WKBaseModel.request` → `observeOn(AndroidSchedulers.mainThread())`), so all `stickersLiveData.value=` writes are main-thread-safe; content-type constants 12/13 (`WKContentType extends WKMsgContentType`); `getUploadCredentials`/`UploadFileUrl` contract; local-order pure-function logic (covered by the new tests).
- **Not verified (out of diff / runtime):** actual `PdfRenderer`/`ZipFile` behavior on device; server contract for sticker `collect`/`upload` (incl. the disclosed `handle`-field uncertainty in `WKStickerUploader`); real R8/release-build behavior of the two proguard keeps; the EditMode touch-dispatch (`CollectStickerAdapter.EditModeTouchListener`) preview-vs-drag disambiguation, which is timing-dependent and needs on-device QA.
- **Advisory legs:** this review absorbed findings from one cross-family advisor plus my own line-level pass. A second automated advisor was unavailable this run due to a gateway error on its side and did not contribute — its absence is not treated as agreement.

mochashanyao (changes requested):
[Octo-Q · automated review]

---

# Code Review — PR #92 (octo-android)

Reviewed at head `652f35a351c04178babac0bf246552c74f279a02`. Scope: sticker subsystem (send/receive/collect/reorder/upload), in-app file preview (PDF/XLSX/text), filename length truncation, name-length throttling, revoke-timing fix.

## Summary

Large feature PR adding custom sticker support, in-app file preview for PDF/XLSX/text, filename length truncation, group/thread name length throttling, and a significant revoke-timing fix. The sticker subsystem is well-architected with clean LiveData wiring, proper thread safety (volatile + main-thread callbacks), and thoughtful iOS parity. The revoke fix (HTTP-200 local mark + syncExtraMsg retry with idempotency guard) is a solid engineering solution to a known server-side cache race. However, the new XLSX parser has two OOM vectors that would let a small crafted file crash the app for any user who previews it.

## Verification

- ✅ **Sticker data flow** — `WKStickerManager.load()` → server → `StickerLocalOrderStore.applyOrder()` → `stickersLiveData` → adapter. All writes go through immutable-swap on main thread; volatile reads safe (`WKStickerManager.kt:46,79`).
- ✅ **Revoke fix idempotency** — `markMsgRevokedLocallyByClientMsgNO` sets `revoke=1` on DB thread; `scheduleRevokeRetry` checks `alreadyRevoked` before re-calling `syncExtraMsg`. Race-free (`WKIMUtils.java:693-776`).
- ✅ **Proguard rules** — Both sticker DTOs and search DTOs correctly kept for FastJson reflection (`app/proguard-rules.pro:163-165,202-206`).
- ✅ **XLSX HTML escaping** — All five HTML-sensitive chars escaped in `escape()`; sheet names and cell data both pass through it; `loadDataWithBaseURL(null, …)` prevents relative URL resolution (`XlsxPreviewActivity.kt:125,222-234`).
- ✅ **WebView security** — JavaScript disabled, file/content access off, base URL null (`XlsxPreviewActivity.kt:77-79`).
- ✅ **File name truncation** — `truncateToByteLength` correctly walks code points and respects UTF-8 byte boundaries; 200-byte cap leaves safe margin under Linux `NAME_MAX=255` (`WKFileProvider.kt:162-185`).
- ✅ **Content-type registration** — Types 12/13 registered with `registerContentMsg`, provider, `MsgConfig(true)`, and thread-creation exclusion. All four extension points covered (`WKUIKitApplication.java:336-337,355-357,375-376,427-429`).
- ✅ **Revoke dual-path parity (C1)** — `WKChatBaseProvider` passes empty `messageID` + valid `clientMsgNO`; `WKUIKitApplication` calls `markMsgRevokedLocallyByClientMsgNO` with same `clientMsgNO`. Both paths align (`WKChatBaseProvider.kt:1287-1308`, `WKUIKitApplication.java:619-629`).
- ✅ **Thread name limit consistency** — `CreateThreadActivity` uses `THREAD_NAME_MAX_LEN=100`; `ThreadDetailActivity` rename uses 100; auto-name truncation uses the same constant (`CreateThreadActivity.java:42,131`, `ThreadDetailActivity.java:212`).

Static analysis only at head `652f35a3`; build and tests not executed in this environment.

## Findings

Two P1 blockers in the new XLSX parser; twelve P2 items across stickers, preview, and utilities.

### P1 — Unbounded ArrayList allocation before MAX_COLS guard (`wkuikit/src/main/java/com/chat/uikit/chat/provider/XlsxParser.kt:176-177`)

`fillTo(currentRow!!, col)` is called **before** the `if (col < MAX_COLS)` bounds check. `columnIndex()` parses the alphabetic prefix of a cell reference (e.g., `"XFD1"` → 16383) and returns it unchecked. A malicious xlsx with cells at extreme column indices causes `fillTo` to pad `currentRow` with up to 16,384 empty strings per row. With 500 rows this produces ~190 MB of allocations from a file that could be a few kilobytes on disk, crashing the app with OOM.

```kotlin
// Current (unsafe):
fillTo(currentRow!!, col)
if (col < MAX_COLS) currentRow!![col] = text else truncated = true

// Fix:
if (col < MAX_COLS) {
    fillTo(currentRow!!, col)
    currentRow!![col] = text
} else {
    truncated = true
}
```

**Diff-scope:** New in this PR. The XlsxParser is entirely new code.

### P1 — No shared-strings size cap or zip-bomb protection (`wkuikit/src/main/java/com/chat/uikit/chat/provider/XlsxParser.kt:57-65`)

`readSharedStrings()` reads `xl/sharedStrings.xml` into an unbounded `ArrayList<String>`. XLSX files are zip-compressed; a crafted archive can decompress a few KB on disk into hundreds of MB of string data, OOMing before any row-level `MAX_ROWS`/`MAX_COLS` truncation applies. The shared string table is also looked up by index in `resolveCellText` (`:197`), so an attacker-controlled shared strings list is a direct path to memory exhaustion.

Fix: cap the shared string count (e.g., 100k entries max) and wrap the `InputStream` in a bounded reader that refuses entries exceeding a decompressed-size budget (e.g., 50 MB).

**Diff-scope:** New in this PR.

### P2 — LiveData observer accumulation on panel rebuild (`wkuikit/src/main/java/com/chat/uikit/chat/ChatPanelManager.kt:2595`)

`stickersLiveData.observe(activity)` registers a new Observer each call — LiveData does not deduplicate by `LifecycleOwner`. If `ChatPanelManager` is re-created for the same Activity (conversation switch in single-Activity architecture), stale observers accumulate and call `submitList`/toggle `visibility` on detached views. Guard with `removeObserver` on teardown.

**Diff-scope:** New in this PR.

### P2 — No upload-in-progress guard; double-upload possible (`wkuikit/src/main/java/com/chat/uikit/chat/ChatPanelManager.kt:2607-2617`)

`pickAndUploadSticker` runs a 4-step async chain (validate → GET credentials → PUT CDN → POST register) with no in-flight flag. Rapid taps on "+" trigger N parallel uploads to different CDN paths → N stickers registered. Add a concurrency guard and ideally a progress indicator.

**Diff-scope:** New in this PR.

### P2 — No DiffUtil; full `notifyDataSetChanged` causes image flicker (`wkuikit/src/main/java/com/chat/uikit/chat/sticker/CollectStickerAdapter.kt:76-79`)

Every `submitList` call triggers `notifyDataSetChanged()`, reloading all Glide images and restarting shake animators. For a ~100-item sticker grid this is tolerable but causes visible flicker on collect/delete/reorder. `ListAdapter` with `DiffUtil.ItemCallback` comparing `sticker_id` would fix this.

**Diff-scope:** New in this PR.

### P2 — `bindingAdapterPosition` can be `NO_POSITION` in delayed preview callback (`wkuikit/src/main/java/com/chat/uikit/chat/sticker/CollectStickerAdapter.kt:192`)

`EditModeTouchListener` posts a 500ms delayed Runnable that reads `holder.bindingAdapterPosition`. If `notifyDataSetChanged` fires in that window, position becomes `-1` and `onPreviewSticker` receives an invalid index. Guard with `bindingAdapterPosition != RecyclerView.NO_POSITION` before firing.

**Diff-scope:** New in this PR.

### P2 — `InputStream.read(buf)` return value unchecked (`wkuikit/src/main/java/com/chat/uikit/chat/sticker/StickerUploadValidator.kt:83-85`)

`FileInputStream(file).use { it.read(buf) }` does not guarantee filling the 12-byte buffer. Partial reads (< 8 bytes) cause valid PNG/WebP magic-byte checks to fail — false "unsupported format" rejections. Use `readNBytes(12)` or loop-read and pass the actual byte count to `detectFormatFromBytes`.

**Diff-scope:** New in this PR.

### P2 — Dimension check bypassed when `BitmapFactory` returns <= 0 (`wkuikit/src/main/java/com/chat/uikit/chat/sticker/StickerUploadValidator.kt:49-52`)

Files that pass magic-byte validation but have corrupt bodies return `width <= 0 || height <= 0`, treated as `Result.success(Meta(0, 0, format))` — bypassing `MAX_DIMENSION_PX` entirely. A 4096×4096 image with a corrupted dimension header slips through to the server. Treat `<= 0` as `Failure.UnsupportedFormat` or `Failure.CorruptImage`.

**Diff-scope:** New in this PR.

### P2 — No `clear()`/`reset()` on logout; sticker cache leaks across sessions (`wkuikit/src/main/java/com/chat/uikit/chat/sticker/WKStickerManager.kt`)

`WKStickerManager` is a singleton `object` that retains `cache` and `stickersLiveData` across the app lifetime. No `clear()` method exists. After logout/login in the same process, the new user sees the previous user's stickers until `load()` succeeds — and if `load()` fails silently, stale data persists indefinitely.

**Diff-scope:** New in this PR (singleton was not previously a concern for per-user data).

### P2 — `load()` failure completely silent (`wkuikit/src/main/java/com/chat/uikit/chat/sticker/WKStickerManager.kt:82-85`)

The `onFail` callback does nothing — no logging, no error state, no user feedback. Combined with the missing `clear()`, a network failure after login leaves the previous user's sticker collection visible. At minimum add `WKLogUtils.e()`.

**Diff-scope:** New in this PR.

### P2 — Hardcoded Chinese strings in preview activities and file provider

User-facing strings like `"无法预览此 Excel 文件"` (`XlsxPreviewActivity.kt:111`), `"空表格"` (`:117`), `"文件过大，仅显示前50000字符"` (`WKFileProvider.kt:350`, `TextPreviewActivity.kt:133`) are hardcoded in Chinese. The app supports multiple locales (`values-en/strings.xml` exists). Move to string resources.

**Diff-scope:** New in this PR.

### P2 — `WKVectorStickerContent` may double-write `url`/`localPath` to Parcel (`wkbase/src/main/java/com/chat/base/msg/model/WKVectorStickerContent.java:88-99`)

`writeToParcel` calls `super.writeToParcel(dest, flags)` then explicitly writes `url` and `localPath`. The Parcel constructor mirrors this. If `WKMediaMessageContent.writeToParcel` already writes these fields (which it likely does as they are its own fields), the Parcel stream contains duplicate entries. The read side is symmetric so it "works" for now — but any parent change would corrupt deserialization. Verify the parent's Parcel contract and remove redundant writes if the parent already handles them.

**Diff-scope:** New in this PR.

### P2 — WebView has no `WebViewClient` set (`wkuikit/src/main/java/com/chat/uikit/chat/provider/XlsxPreviewActivity.kt:76-84`)

JavaScript is correctly disabled, but no `WebViewClient` overrides `shouldOverrideUrlLoading`. Today the HTML only contains `#s{i}` fragment anchors which are safe. But a defense-in-depth gap: any future change that introduces a non-fragment URL would allow the system to handle it (potentially launching external apps for `http://`, `intent://`, `market://` schemes). Set a `WebViewClient` that blocks non-fragment navigation.

**Diff-scope:** New in this PR.

## Data-flow tracing

| Consumed data | Origin | Path verified |
|---|---|---|
| `stickersLiveData` value | `WKStickerManager.load()` → server → `StickerLocalOrderStore.applyOrder()` → `stickersLiveData.value` | Observer at `ChatPanelManager.kt:2595` receives ordered list, passes to `adapter.submitList()` ✅ |
| `sticker.path` in `sendCollectedSticker` | Server `ListStickerResp` → `WKSticker.path` → cache → LiveData → adapter → `onStickerClick` | Flows correctly but no null guard on `sticker.path` ⚠️ |
| `content.url` in `WKVectorStickerProvider.setData` | `WKMsg.baseContentMsgModel` cast to `WKVectorStickerContent` → `url` | Null-checked with `isNullOrEmpty()`; passes to Glide via `WKApiConfig.getShowUrl()` ✅ |
| `clientMsgNo` in revoke path | `WKChatBaseProvider` withdraw click → `WithdrawMsgMenu.clientMsgNo` → `WKUIKitApplication` revoke callback → `markMsgRevokedLocallyByClientMsgNO` | Full chain verified; `clientMsgNo` preserved through all hops ✅ |
| `messageId` in `scheduleRevokeRetry` | `WKIMUtils.revokeMsg` → `Handler.postDelayed(500/1500ms)` → `WKDbScheduler.scheduleDirect` → `getWithMessageID` | Reads `remoteExtra.revoke` for idempotency check ✅ |
| `fillTo(currentRow!!, col)` in XlsxParser | `columnIndex(cellRef)` from cell `r` attribute | **NOT bounds-checked before call** — see P1 ❌ |
| `sharedStrings` ArrayList | `ZipFile.readEntry("xl/sharedStrings.xml")` → `XmlPullParser` | **No size cap** — see P1 ❌ |

## Blind-spot checklist

- **C1 (dual-path parity)** — Revoke: `WKChatBaseProvider` (withdraw click) and `WKUIKitApplication` (revoke callback) both use `clientMsgNO`. Sticker: collect (`/sticker/user/collect`) and delete (`DELETE /sticker/user/{id}`) both scoped to authenticated user. Clear.
- **C2 (control-flow ordering)** — `markMsgRevokedLocallyByClientMsgNO` dispatched to `WKDbScheduler.scheduleDirect`; `scheduleRevokeRetry` also dispatched to same scheduler. `alreadyRevoked` check ensures idempotency regardless of ordering. Clear.
- **C3 (authorization boundary)** — Sticker send goes through `iConversationContext.sendMessage()` (conversation auth). Upload uses `WKConfig.getInstance().uid` for CDN path scoping; server validates ownership. Delete uses `/sticker/user/` path (server-scoped). No bypass path found. Clear.
- **C4 (authorization lifecycle)** — N/A: no auth/permission container-member cascade changes.
- **C5 (build ≠ runtime correctness)** — Proguard rules added for both sticker and search DTOs. All runtime paths are exercised through standard Android SDK APIs. Clear.
- **C6 (governance/policy)** — N/A: no governance or policy documents changed.

## Cross-round blocker re-check

N/A — automated review review.

- **Formula injection mitigated** — `XlsxParser.resolveCellText` returns cached `<v>` value for formula cells, never the formula expression. `escape()` further neutralizes prefix chars (`XlsxParser.kt:198`, `XlsxPreviewActivity.kt:222-234`).
- **No sensitive data logging** — No Logcat output of sticker URLs, CDN paths, tokens, or user IDs in any changed file.
- **Resource cleanup** — Both preview activities clean up in `onDestroy`: `XlsxPreviewActivity` calls `executor.shutdownNow()` + `webView.destroy()`; `PdfPreviewActivity` closes `pdfRenderer` + `fileDescriptor`.
- **Async loading flags cleared on failure** — `progressBar.visibility = View.GONE` runs in both success and failure callbacks regardless of outcome (`XlsxPreviewActivity.kt:108`).
- **Sticker thread safety** — `@Volatile` cache with immutable-swap on main thread. `StickerLocalOrderStore` uses `SharedPreferences` (thread-safe). `WKStickerUploader` is stateless.
- **Drag reorder persistence** — `clearView` in `ItemTouchHelper.Callback` passes `currentOrderIds()` to `WKStickerManager.reorder()`, which writes to `LocalOrderStore` and updates LiveData using the parameter order. Correctly avoids the async-apply race.
- **PDF uses `PdfRenderer` (not WebView)** — No WebView security concerns for PDF rendering.

[Octo-Q] verdict: REQUEST_CHANGES + Two P1 OOM vectors in XlsxParser (unbounded `fillTo` before `MAX_COLS` guard at line 176; unbounded sharedStrings decompression with no zip-bomb protection) allow a small crafted xlsx shared in any chat to crash the app for every user who previews it. Both fixes are straightforward — move bounds check before `fillTo`, cap shared strings count + decompressed size. Twelve P2 items should be addressed in follow-up but do not block after P1s are fixed. Revoke architecture and sticker subsystem are well-engineered.

yujiawei (commented):
# Supplemental note — PR #92 (octo-android)

Adding one blocking item to my earlier review after a second static pass. Verdict is unchanged: **CHANGES_REQUESTED**.

### P1 (additional) — XlsxParser reads sharedStrings with no size cap (`XlsxParser.kt:57-65`, parse at `:55-56`)
`readSharedStrings()` loads `xl/sharedStrings.xml` into an unbounded `ArrayList<String>` before any row/column truncation applies. XLSX is zip-compressed, so a few-KB archive can inflate into hundreds of MB of strings and OOM the app on any user who previews it. This is the same class of vector as the `fillTo`-before-`MAX_COLS` issue and equally reachable from a chat attachment. Cap the entry count (e.g. 100k) and bound decompressed size while reading.

### Note on a potential false-positive fix
A prior automated pass suggested `WKVectorStickerContent.writeToParcel` double-writes `url`/`localPath`. That is **not** the case: the parent `WKMediaMessageContent` declares those fields but does not serialize them (no `writeToParcel` override; its Parcel constructor only calls `super(in)`). The explicit writes in `WKVectorStickerContent` are required — removing them would drop `url`/`localPath` from the parcel. Please do not "simplify" that.

Jerry-Xin (approved):
Re-review at head `2b3ea962` — all three prior blockers in `XlsxParser.kt` are fixed. Approving.

## 🔴 Prior blockers — resolved

**B1 — negative column index / IOOBE on missing `r` (FIXED)**
`columnIndex()` still returns `-1` for empty/letterless refs, but `readSheet` now branches explicitly on the `</c>` handler:
```
if (col in 0 until MAX_COLS) { ... row[col] = text }
else if (col >= MAX_COLS) { truncated = true }
// col < 0  -> falls through, cell skipped
```
`col < 0` no longer reaches `currentRow[-1]`; OOXML files that omit the optional `r` attribute now parse without crashing (the affected cell is skipped rather than crashing the preview).

**B2 — `fillTo` allocating before the column cap (FIXED)**
`fillTo(row, col)` + `row[col] = text` now execute only inside the `col in 0 until MAX_COLS` branch, so `fillTo` allocates at most `MAX_COLS (50)` entries. A malicious wide coordinate (e.g. `XFD1`) hits the `col >= MAX_COLS` branch, which just sets `truncated` and allocates nothing. Memory-DoS vector closed.

**B3 — unbounded shared-strings / zip-bomb (FIXED)**
Two independent bounds:
- `MAX_SHARED_STRINGS = 100_000` — `if (out.size < MAX_SHARED_STRINGS) out.add(...)`; parsing continues past the cap but excess entries are dropped, and out-of-range `shared[i]` reads resolve to `""` via `if (it in shared.indices)`.
- `boundedStream()` wraps every zip entry's `InputStream` in `BoundedInputStream(MAX_ENTRY_BYTES = 50 MB)`, throwing `IOException` on overflow — applied uniformly to sharedStrings, sheets, rels and workbook. Zip-bomb / OOM vector closed.

## 🟡 Prior non-blocking notes — also addressed
- Whole-file read in text preview replaced by new `TextPreviewLoader.readBounded()` (chunked read to `maxChars`, probe for truncation), invoked from a background executor in `TextPreviewActivity` and reused by `WKFileProvider`.
- `PdfPreviewActivity` gained a single-thread `renderExecutor` serializing all `PdfRenderer` ops, per-page bitmap clamp to `MAX_BITMAP_PIXELS`, and OOM-per-page skip.
- `WKStickerManager.reorderCache()` extracted as a pure, unit-testable seam with a 3-case contract test locking the race-free invariant.

## ✅ Verification
- Byte-verified `XlsxParser.kt` at `ref=2b3ea962`; the three guards land as described, no off-by-one on the cap (`fillTo` ≤ 50 entries), `resolveCellText` shared-index access is bounds-checked, and legit files (missing `r`, inlineStr, shared, boolean/number) still parse.
- No new bugs introduced by the fix commits.

CI: `check-sprint` is the known governance/sprint gate; all code checks (osv-scanner, dependency-review, secret-scan, history, label, pr-title-lint) are green. Nice work turning these around.

OctoBoooot (approved):
## Review: feat(chat): 批量改进 — 贴纸/预览/名称长度/撤回/release 修复 (#92)

**Verdict:** Approve.
All four substantive findings from my prior review of this PR are fixed at this head (2b3ea962), each verified against source. Code CI is green (build ✅, osv-scanner ✅, secret-scan ✅, dependency-review ✅, pr-title-lint ✅). The only red check is `check-sprint` — a governance/board-hygiene gate (the body already carries `Closes #87..#91`, so it should re-run green); it does not block on code grounds, but note `mergeStateStatus` stays BLOCKED until it passes, so this is "approved, not yet mergeable."

I re-reviewed the delta from my prior head (652f35a) to 2b3ea962 — 8 files, all in the areas I flagged.

### Fixed (verified)
- **XlsxParser negative-index crash** → `columnIndex` now returns -1 when no column letters are present, and the `<c>` END_TAG handler has explicit `col in 0 until MAX_COLS` / `col >= MAX_COLS` (mark truncated) / `col < 0` (skip cell) branches. The `currentRow!![-1]` write is gone. A cell missing the optional `r` attribute is now skipped, not a crash.
- **Zip-bomb / decompression OOM** → every entry is now read through `BoundedInputStream` with a 50 MB cap (throws `IOException` → whole parse fails gracefully to "preview failed"), and `readSharedStrings` is capped at `MAX_SHARED_STRINGS = 100_000`.
- **PDF page OOM** → `renderPage` scales down to `MAX_BITMAP_PIXELS` before `createBitmap`, wraps the alloc+render in `catch (OutOfMemoryError)` / `catch (Exception)` returning null (single page skipped, not a crash), and rendering now runs on a background executor with results posted to the main thread — which also resolves the main-thread-render concern.
- **Sticker reorder had no test** → `reorderCache` extracted as a pure function with `WKStickerManagerReorderTest` asserting the result order is determined *only* by the passed-in ids, not a persisted SP read. Mutation-real: reverting `reorder()` to `applyOrder(read(), cache)` fails the test.
- **TextPreview full-file read on UI thread** (prior minor) → extracted to a background `TextPreviewLoader`.

### Praise
- Clean, precisely-scoped fixes — each targets the exact failure mode (bounded stream for the bomb, pixel clamp + OOM catch for the bitmap, explicit tri-state column handling for the index) rather than a blanket try/catch, and the reorder fix extracts a testable pure function rather than just asserting behavior indirectly. The XlsxParser comments now explicitly frame the input as an untrusted attachment and document each bound's rationale. Strong response to a four-reviewer round.

### Out of scope (informational, non-blocking — carried from prior review)
- Sticker `content.url` from a remote message → Glide via `getShowUrl` (unchanged) with no visible scheme allow-list — pre-existing image-path exposure, worth confirming Glide restricts schemes, but not introduced here.
- The `messageID='0'` recall path: confirm `MsgModel.revokeMsg` does the empty-string server reverse-lookup the body describes (out of this diff).
- `.gitignore` additions bundled into a feature PR (a peer noted this) — a scope nit, author's call.

_Approved on code grounds; merge still gated on `check-sprint` (issue links present, should clear on re-run) — a maintainer action, not a code change._

yujiawei (approved):
# Code Review — PR #92 (octo-android)

Re-review at head `2b3ea9621b8612dc6f5becc54b8b6f6740fdec4c`. Base `main`, merge-base `e183a8d8`. 52 files, +4022/−64, 5 feature groups (revoke fix, name-length throttle, release proguard keeps, in-app file preview, custom stickers). This round adds two fix commits (`a267bd57` preview hardening, `2b3ea962` reorder contract test) on top of the previously-reviewed `652f35a3`. Prior formal reviews were all submitted against `652f35a3`; per re-review policy their approvals/blocks do not auto-carry, so this is a fresh pass against the current head plus targeted verification of the four blockers raised last round.

## 1. Spec compliance

**Spec: ✅**

All five linked issues (#87–#91) are implemented as described, and no functionality was built beyond the stated scope:

- **#87 revoke** — `WKChatBaseProvider` passes `clientMsgNO` as the authority and an empty `message_id` when `messageID` is empty/`"0"`; `WKUIKitApplication` marks locally only on HTTP 200 (`code == success` branch); `WKIMUtils.markMsgRevokedLocallyByClientMsgNO` + `scheduleRevokeRetry` implement the dual path with an `alreadyRevoked` idempotency guard. Every `WKSyncExtraMsg` field is preserved on `saveRemoteExtraMsg` so `insertOrReplace` does not drop `readed`/`is_pinned`/`extra_version`/etc.
- **#88 name length** — `ThrottledRunnable` + `StringUtils.attachLengthLimit()` consolidate the length toast; call sites updated consistently in both code and XML `maxLength` (group 30→50, thread 50→100).
- **#89 / release CCE** — proguard `-keep` for `com.chat.base.search.channel.dto.**`.
- **#90 file preview** — `PdfPreviewActivity` / `XlsxPreviewActivity` / `XlsxParser` added; `WKFileProvider` gains an extension-gated Preview button; `truncateToByteLength` fixes long-filename `ENAMETOOLONG`.
- **#91 stickers** — content types 12/13 registered, providers, collect/upload/delete/reorder, local order store, unit tests.

## 2. Code quality

**Quality: Approved**

The four blockers from the previous round are all genuinely resolved at this head (each re-verified against source, not inherited):

- ✅ **XlsxParser OOM/DoS (was 2×P1)** — `fillTo()` now runs strictly inside the `col in 0 until MAX_COLS` guard (`XlsxParser.kt:190-196`), so a crafted column ref can no longer force an unbounded `ArrayList`; `col < 0` (missing `r` attribute) is skipped instead of indexing `[-1]`. Shared strings are capped at `MAX_SHARED_STRINGS=100_000` (`:97`) and every zip entry is read through `BoundedInputStream` with a 50 MB decompressed cap (`:263-289`), closing the zip-bomb vector. Integer overflow in `columnIndex` is neutralized by the same range guard.
- ✅ **PDF main-thread render / bitmap OOM (was P1)** — all `PdfRenderer` access is serialized on a single-thread `renderExecutor` (satisfies the renderer's thread-affinity contract), results posted back with a stale-holder check; bitmaps are clamped to `MAX_BITMAP_PIXELS` and `createBitmap`/`render` catch `OutOfMemoryError` so one bad page is skipped (`PdfPreviewActivity.kt:215-253`).
- ✅ **Text preview whole-file read (was P1)** — `TextPreviewLoader.readBounded` streams at most `MAX_CHARS`; used off the main thread in `TextPreviewActivity` and at the `WKFileProvider.openFileDirectlyStatic` call site.
- ✅ **Reorder race-free contract untested (was P1)** — `reorderCache` is now an `internal` seam that production `reorder()` delegates to, and `WKStickerManagerReorderTest` is mutation-resistant: reverting to `applyOrder(read(), cache)` would make `read()` return empty in the JVM test and fail the assertion.

None of the residual items rise to a merge blocker. The security posture of the preview feature (WebView hardened, HTML-escaped cells, non-exported activities, magic-byte upload validation) remains solid.

### P2 / non-blocking (follow-up welcome, not gating)

- `StickerUploadValidator.kt:43` — `FileInputStream.use { it.read(buf) }` ignores the return value; a short read can spuriously reject a valid PNG/WebP. Use `readNBytes`/loop-read.
- `StickerUploadValidator.kt:20-22` — when `BitmapFactory` reports `width/height <= 0` (corrupt header) the file passes as `Meta(0,0,…)`, bypassing `MAX_DIMENSION_PX`. Treat `<= 0` as a failure.
- `CollectStickerAdapter` — `submitList` triggers full `notifyDataSetChanged` (Glide reload/flicker on collect/delete/reorder); a `DiffUtil.ItemCallback` on `sticker_id` would fix it.
- `ChatPanelManager.kt:2595` — `stickersLiveData.observe(activity)` registered in `buildStickerContentView()`; guard with `removeObserver` if the toolbar is ever re-initialized within one activity lifecycle (bounded today).
- Hardcoded user-facing strings in the preview activities / `WKFileProvider` should move to resources for locale parity.
- `WKIMUtils.markMsgRevokedLocallyByClientMsgNO` empty-`messageID` branch is now documented as a known, bounded edge (only the just-sent-not-yet-acked window; the extra table is `messageID`-indexed so it genuinely cannot be written otherwise). Acceptable; a `clientMsgNO`-indexed pending-revoke table is the noted future direction.
- `PdfPreviewActivity.onDestroy` uses `renderExecutor.shutdown()` (not `shutdownNow`). Since the executor is single-threaded/FIFO and both bind and teardown run on the main thread, the close task is always enqueued last and runs — no use-after-close in practice — but adding `awaitTermination` fallback would make the teardown intent explicit.

## Coverage / blind spots

- **Verified this round:** all four prior blockers against current source; `BoundedInputStream` wraps every entry path and closes via `use{}` on the throw path; revoke local-mark fires only on HTTP success; reorder test mutation-resistance; content-type 12/13 registration.
- **Adjudicated and dismissed:** an automated pass suggested new P1s (PDF executor use-after-close, `pageCount` visibility race, BoundedInputStream leak on IOException, TextPreview double-load). Each was re-checked against the diff and found to be a non-issue — single-thread FIFO executor + main-thread-only lifecycle, `Handler.post` happens-before publication, Kotlin `use{}` closing on throw, and `content` extra being set so the async branch is skipped, respectively.
- **Not verified (out of diff / device-only):** real `PdfRenderer`/`ZipFile` behavior on device, server contract for sticker collect/upload, actual R8/release behavior of the proguard keeps, and the EditMode touch-dispatch preview-vs-drag disambiguation (timing-dependent, needs on-device QA per the PR's own test plan).

## 3. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved. All four previously-blocking P1s are fixed at the current head; remaining items are P2/nits that do not gate merge. The `build` check passes; `check-sprint` is a governance check outside code scope (issue links are present in the body).

mochashanyao (approved):

PR #95, opened 2026-07-17
feat(search): 全局搜索接入服务端聚合 L1 + L2 深度浏览

mzanalyticst:
## Summary

- **L1 广度聚合**：会话列表页顶部搜索的"聊天记录"段从本地聚合切换到服务端 `POST /v1/messages/_search_global_groups`，覆盖跨群/跨 DM/跨子区的服务端可见消息；失败/离线/裸 404 自动 fallback 到本地 `WKIM.msgManager.search`
- **L2 深度浏览**：点某群/子区/DM 桶跳新的 `GroupDrillActivity` 走 `POST /v1/messages/_search_global_messages`，服务端 "传群 id 自动展开群+子区" 生效；message + file 单视图混排，cursor 分页
- **子区头像 fallback**：子区（channel_type=5）的 channel_id 是复合结构 AvatarView 无法直接加载，沿用父群头像；子区名展示 "父群名 · 子区名" 便于识别

对齐服务端契约 `04-aggregation-api-spec.md` §2/§3。

## 改动清单

**新增（wkbase/search/global/）**
- `SearchGlobalGroupsService/BodyBuilder/Model` + DTO（GroupBucket / GlobalSearchFilters / SearchGlobalGroupsReq/Resp）
- `SearchGlobalMessagesService/BodyBuilder/Model` + Req DTO（响应壳复用现有 CursorList<CombinedHit>）

**新增（wkuikit/search/remote/）**
- `GlobalSearchViewModel`：300ms debounce + 单调 sequence + 3 层乱序防护
- `drill/GroupDrillActivity/Adapter/ViewModel` + `act_group_drill_layout.xml`

**改现有**
- `wkbase/search/channel/dto/MessageHit.kt` / `FileHit.kt`：追加可选 `channel_type` / `channel_id`（服务端全局搜索场景回填 preview），向后兼容频道内搜索
- `wkuikit/search/remote/GlobalActivity.kt`：三段缓存 rebuild，聊天记录段接 L1；点 LOCAL_MSG 跳 GroupDrill；子区名 "父群 · 子区"
- `wkuikit/search/remote/GlobalAdapter.kt`：LOCAL_MSG 分支子区头像用父群加载
- `wkuikit/AndroidManifest.xml`：注册 GroupDrillActivity
- `wkuikit/build.gradle`：加 kotlinx-coroutines-test 测试依赖

## 策略确认

| 分段 | 策略 |
|---|---|
| 联系人段 | 老代码"本地 + `/search/global` 双源合并"保留 |
| 群段 | 同上 |
| 查找用户快捷 | 保留 |
| 聊天记录段 (L1) | **服务端优先 + 本地兜底**（`FALLBACK_TO_LOCAL` / 裸 404 → 老 `WKIM.msgManager.search`） |
| L2 深度页 | **只服务端**（本地无等价数据不做兜底，失败仅 toast） |

Space 隔离：`X-Space-Id` 自动注入，只搜当前 space 内可见消息（对齐飞书/钉钉惯例）。

## 单测

- wkbase 351 + wkuikit 219 = 100% pass
- 新增 30 用例：BodyBuilder 序列化 / Resp 反序列化 / Filters 触发门 / ViewModel debounce+sequence+乱序防护 / loadMore 追加 / 失败保留已加载 / rate_limited 场景

## Test plan

- [ ] 冷启动 → 会话列表页点顶部搜索 → 输入关键词 → 联系人/群/查找用户/聊天记录段全部展示
- [ ] 聊天记录段：单条命中直接展 snippet 紫色高亮；多条展"约 N 条相关聊天记录"
- [ ] 子区桶：显示"父群名 · 子区名"，头像沿用父群
- [ ] 快速输入多字符：只发一次 L1 请求（debounce 生效）
- [ ] 断网/prod 未部署接口场景：聊天记录段回退本地聚合，不空白
- [ ] 点聊天记录段某桶 → 跳 GroupDrillActivity（不是 MessageRecordActivity）
- [ ] L2 页：预填 keyword，混合 message + file 结果，`<mark>` 高亮
- [ ] L2 页：上滑触底 loadMore（若 hasMore=true）
- [ ] L2 页：修改 keyword 重搜（debounce + reset cursor）
- [ ] L2 页：点某条命中跳 ChatActivity（本地有消息则定位；无则空白 - 老问题）
- [ ] 频道内 "查找聊天记录"（MessageRecordActivity）无回归
- [ ] SearchUserActivity 联系人搜索无回归
- [ ] 会话列表/消息收发/上下文/我 tab 等其他模块无回归

## 已知残留（老代码历史行为，本 PR 不修，后续单独跟进）

- **服务端搜出 UI 隐藏的默认分组会话**：`is_default=true` category 三端 UI 一致故意不展示；服务端 messages_search 未同步该规则 → 建议服务端 filterVisible 增加一维过滤
- **点单条命中跳 ChatActivity 定位失败空白**：`chatView(orderSeq)` 定位机制依赖本地 IMSDK 消息记录，服务端搜出的历史消息本地不一定有；建议后续 PR 用 `_search_around` 端点补齐锚点前后消息

### DIFF
--- a/wkbase/src/main/java/com/chat/base/search/global/SearchGlobalGroupsBodyBuilder.kt
+++ b/wkbase/src/main/java/com/chat/base/search/global/SearchGlobalGroupsBodyBuilder.kt
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2026-present OctoIM contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.chat.base.search.global
+
+import com.alibaba.fastjson.JSONArray
+import com.alibaba.fastjson.JSONObject
+import com.chat.base.search.global.dto.GlobalChannelRef
+import com.chat.base.search.global.dto.GlobalSearchFilters
+import com.chat.base.search.global.dto.SearchGlobalGroupsReq
+
+/**
+ * 序列化 L1 请求体为 FastJSON [JSONObject]，对齐 octo-server
+ * `modules/messages_search/search_global_groups.go` 的 `SearchGlobalGroupsReq` 与
+ * `04-aggregation-api-spec.md` §2.1。
+ *
+ * 关键规则：
+ *  - keyword 为空**不写入** body（触发门交给 [GlobalSearchFilters.canTriggerL1] 判断，此处不做二次校验）。
+ *  - sequence 恒写入（服务端原样回带，用于 UI 层丢弃过期响应）。
+ *  - filters 各字段为空/空集合时**省略**该字段，避免出现空数组导致的语义歧义。
+ *  - filters 完全为空时整个 filters 字段省略。
+ *  - 客户端不做长度/枚举校验，交由服务端 400 反馈以避免规则漂移。
+ */
+internal object SearchGlobalGroupsBodyBuilder {
+
+    fun build(req: SearchGlobalGroupsReq): JSONObject {
+        val body = JSONObject()
+        if (!req.keyword.isNullOrEmpty()) {
+            body["keyword"] = req.keyword
+        }
+        body["sequence"] = req.sequence
+        req.filters?.takeUnless { it.isEmpty() }?.let { body["filters"] = filtersToJson(it) }
+        return body
+    }
+
+    private fun filtersToJson(f: GlobalSearchFilters): JSONObject {
+        val out = JSONObject()
+        f.senderIds?.takeIf { it.isNotEmpty() }?.let { out["sender_ids"] = it }
+        f.memberUids?.takeIf { it.isNotEmpty() }?.let { out["member_uids"] = it }
+        f.memberUid?.takeIf { it.isNotEmpty() }?.let { out["member_uid"] = it }
+        f.channelIds?.takeIf { it.isNotEmpty() }?.let { out["channel_ids"] = channelRefsToJson(it) }
+        f.channelTypes?.takeIf { it.isNotEmpty() }?.let { list ->
+            out["channel_types"] = list.map { it.toInt() }
+        }
+        f.contentTypes?.takeIf { it.isNotEmpty() }?.let { out["content_types"] = it }
+        f.sentAtFrom?.takeIf { it.isNotEmpty() }?.let { out["sent_at_from"] = it }
+        f.sentAtTo?.takeIf { it.isNotEmpty() }?.let { out["sent_at_to"] = it }
+        return out
+    }
+
+    private fun channelRefsToJson(refs: List<GlobalChannelRef>): JSONArray {
+        val arr = JSONArray()
+        for (r in refs) {
+            val o = JSONObject()
+            o["channel_id"] = r.channelId
+            o["channel_type"] = r.channelType.toInt()
+            arr.add(o)
+        }
+        return arr
+    }
+}
--- a/wkbase/src/main/java/com/chat/base/search/global/SearchGlobalGroupsService.kt
+++ b/wkbase/src/main/java/com/chat/base/search/global/SearchGlobalGroupsService.kt
@@ -1,0 +1,35 @@
+/*
+ * Copyright 2026-present OctoIM contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.chat.base.search.global
+
+import com.alibaba.fastjson.JSONObject
+import com.chat.base.search.global.dto.SearchGlobalGroupsResp
+import io.reactivex.rxjava3.core.Observable
+import retrofit2.http.Body
+import retrofit2.http.POST
+
+/**
+ * 全局聚合搜索 L1 端点（`POST /v1/messages/_search_global_groups`）。
+ *
+ * `X-Space-Id` / `token` 由 [com.chat.base.net.CommonRequestParamInterceptor] 自动注入。
+ * 与 [com.chat.base.search.channel.ChannelSearchService] 共享错误信封 [com.chat.base.search.channel.dto.SearchErrorCode]
+ * 与 401 auth 中间件行为。
+ */
+interface SearchGlobalGroupsService {
+    @POST("messages/_search_global_groups")
+    fun searchGlobalGroups(@Body body: JSONObject): Observable<SearchGlobalGroupsResp>
+}
--- a/wkbase/src/main/java/com/chat/base/search/global/SearchGlobalMessagesService.kt
+++ b/wkbase/src/main/java/com/chat/base/search/global/SearchGlobalMessagesService.kt
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2026-present OctoIM contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.chat.base.search.global
+
+import com.alibaba.fastjson.JSONObject
+import com.chat.base.search.channel.dto.CombinedHit
+import com.chat.base.search.channel.dto.CursorList
+import io.reactivex.rxjava3.core.Observable
+import retrofit2.http.Body
+import retrofit2.http.POST
+
+/**
+ * 全局聚合搜索 L2 端点（`POST /v1/messages/_search_global_messages`）。
+ *
+ * 契约与频道内 `_search_all` 完全对齐——响应体是 `{data: List<SearchAllHit>, pagination}`，
+ * 其中 SearchAllHit 就是 [CombinedHit]（message 与 file 混排，按 result_type 判断）。
+ *
+ * L2 与 L1 是两条独立分页轴：L2 的 cursor 与 L1 无关，互不复用。
+ * `X-Space-Id` / `token` 由 [com.chat.base.net.CommonRequestParamInterceptor] 自动注入。
+ */
+interface SearchGlobalMessagesService {
+    @POST("messages/_search_global_messages")
+    fun searchGlobalMessages(@Body body: JSONObject): Observable<CursorList<CombinedHit>>
+}
--- a/wkbase/src/main/java/com/chat/base/search/global/dto/GroupBucket.kt
+++ b/wkbase/src/main/java/com/chat/base/search/global/dto/GroupBucket.kt
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2026-present OctoIM contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.chat.base.search.global.dto
+
+import com.chat.base.search.channel.dto.MessageHit
+
+/**
+ * L1 聚合桶。一个桶 = 一个群 / 一个子区 / 一个 DM。
+ *
+ * 语义（见 `04-aggregation-api-spec.md` §2.4）：
+ *  - presence 精确（K2=500 内 100% 精确）：列出的桶点进 L2 至少 1 条可见，不会"点进去空"
+ *  - [match_count] / [match_count_approx]=true：pre-visibility 近似，UI 显示 "约 N 条"
+ *  - [preview] 每条已做可见性过滤，可直接渲染，[MessageHit.snippet] 已带 `<mark>` 高亮
+ *  - DM 桶：channel_type=1，group_name 为对端用户名，无 parent_group_no / thread_*
+ *  - 子区桶：channel_type=5，parent_group_no + thread_id + thread_name 均非空
+ */
+class GroupBucket {
+    lateinit var channel_id: String
+    var channel_type: Byte = 0
+    /** 群/子区共父群号；DM 桶为空。 */
+    var parent_group_no: String? = null
+    var group_name: String = ""
+    /** 仅子区桶（channel_type=5）非空。 */
+    var thread_id: String? = null
+    var thread_name: String? = null
+    var match_count: Long = 0L
+    var match_count_approx: Boolean = true
+    /** RFC3339。基于可见 hit 重算，不会泄露隐藏消息时间。 */
+    var latest_at: String = ""
+    var preview: List<MessageHit> = emptyList()
+}
--- a/wkbase/src/main/java/com/chat/base/search/global/dto/GroupBucket.kt
+++ b/wkbase/src/main/java/com/chat/base/search/global/dto/GroupBucket.kt
@@ -29,7 +29,10 @@ import com.chat.base.search.channel.dto.MessageHit
  *  - 子区桶：channel_type=5，parent_group_no + thread_id + thread_name 均非空
  */
 class GroupBucket {
-    lateinit var channel_id: String
+    // 服务端契约保证非空，但客户端不用 lateinit：漏字段场景 lateinit 会抛
+    // UninitializedPropertyAccessException 崩 UI 层；给默认空串让下游 `bucketToDataVO`
+    // 走 fallback 分支（`channel_id.ifEmpty { ... }` / adapter 校验），可控降级。
+    var channel_id: String = ""
     var channel_type: Byte = 0
     /** 群/子区共父群号；DM 桶为空。 */
     var parent_group_no: String? = null
--- a/wkuikit/src/main/java/com/chat/uikit/search/remote/GlobalSearchViewModel.kt
+++ b/wkuikit/src/main/java/com/chat/uikit/search/remote/GlobalSearchViewModel.kt
@@ -147,17 +147,23 @@ class GlobalSearchViewModel(
             sequence = mySeq,
         )
         val req = SearchGlobalGroupsReq(keyword = keyword, sequence = mySeq)
-        searchCaller.call(req) { outcome -> handleResponse(mySeq, outcome) }
+        searchCaller.call(req) { outcome -> handleResponse(mySeq, keyword, outcome) }
     }
 
     private fun handleResponse(
         reqSeq: Long,
+        requestKeyword: String,
         outcome: ChannelSearchOutcome<SearchGlobalGroupsResp>,
     ) {
         // 乱序防护 1：本地已发出更新的请求 → 丢弃此响应
         if (reqSeq != nextSequence) return
         // 乱序防护 2：keyword 已被用户清空 → 丢弃
         if (_state.value.keyword.isEmpty()) return
+        // 乱序防护 3：keyword 已变（debounce 窗口内响应到达但用户已改词）→ 丢弃
+        // 场景：setKeyword("foo") → seq=1 在飞；用户改 "bar" → state.keyword 立即更新但
+        // ++nextSequence 要等 debounce 300ms 后 triggerSearch。foo 响应在此窗口回来时，
+        // reqSeq(1)==nextSequence(1) 通过，此处 requestKeyword("foo")!=state.keyword("bar") → 丢弃。
+        if (_state.value.keyword != requestKeyword) return
 
         if (outcome.ok) {
             val body = outcome.data ?: return
--- a/wkuikit/src/main/java/com/chat/uikit/search/remote/GlobalSearchViewModel.kt
+++ b/wkuikit/src/main/java/com/chat/uikit/search/remote/GlobalSearchViewModel.kt
@@ -1,0 +1,202 @@
+/*
+ * Copyright 2026-present OctoIM contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.chat.uikit.search.remote
+
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.viewModelScope
+import com.chat.base.search.channel.ChannelSearchOutcome
+import com.chat.base.search.channel.uiAction
+import com.chat.base.search.global.SearchGlobalGroupsModel
+import com.chat.base.search.global.dto.GlobalSearchFilters
+import com.chat.base.search.global.dto.GroupBucket
+import com.chat.base.search.global.dto.SearchGlobalGroupsReq
+import com.chat.base.search.global.dto.SearchGlobalGroupsResp
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.launch
+
+/**
+ * 全局搜索"聊天记录聚合"段 (L1 `_search_global_groups`) 的 ViewModel。
+ *
+ * 职责：
+ *  - 输入 keyword 做 300ms debounce，与服务端 5 QPS / 20 burst 限流配合
+ *  - 递增 [sequence] 回带服务端，只渲染最新序号的响应（§4 竞态防护）
+ *  - state 分发 loading / groups / error 三态，UI 层订阅
+ *
+ * 不做本地兜底：`FALLBACK_TO_LOCAL` 语义由 UI 层根据 [State.uiAction] 决定是否
+ * 调用 `WKIM.msgManager.search()` 显示离线结果。ViewModel 保持不依赖 IMSDK，便于纯 JVM 单测。
+ *
+ * 联系人/群段的搜索走独立本地路径，不在此 ViewModel。
+ */
+class GlobalSearchViewModel(
+    /** 注入服务端调用，测试时替换为 fake。 */
+    private val searchCaller: SearchCaller = DefaultSearchCaller,
+    private val debounceMillis: Long = DEBOUNCE_MS,
+) : ViewModel() {
+
+    /** UI 侧观察的完整状态。所有派生数据（分组数量等）都可从此对象取得。 */
+    data class State(
+        val keyword: String = "",
+        val isLoading: Boolean = false,
+        val groups: List<GroupBucket> = emptyList(),
+        val totalGroups: Long = 0L,
+        val totalGroupsApprox: Boolean = true,
+        val hasMore: Boolean = false,
+        /** 服务端结构化错误码；无错误时为 null。 */
+        val errorCode: String? = null,
+        val errorMessage: String? = null,
+        val retryAfterSec: Int = 0,
+        /** 当前渲染中的响应对应的 sequence。0 表示还未发起过请求。 */
+        val sequence: Long = 0L,
+    ) {
+        /** null 时 UI 无需响应；非 null 时按枚举决定 toast / banner / 本地兜底。 */
+        fun uiAction(): com.chat.base.search.channel.ChannelSearchUiAction? =
+            errorCode?.let {
+                ChannelSearchOutcome.failure<Unit>(
+                    httpStatus = 0,
+                    errorCode = it,
+                ).uiAction()
+            }
+    }
+
+    /** ViewModel 内部对服务端的调用抽象，仅为可测试性存在。 */
+    fun interface SearchCaller {
+        fun call(
+            req: SearchGlobalGroupsReq,
+            callback: (ChannelSearchOutcome<SearchGlobalGroupsResp>) -> Unit,
+        )
+    }
+
+    private object DefaultSearchCaller : SearchCaller {
+        override fun call(
+            req: SearchGlobalGroupsReq,
+            callback: (ChannelSearchOutcome<SearchGlobalGroupsResp>) -> Unit,
+        ) {
+            SearchGlobalGroupsModel.searchGroups(req, callback)
+        }
+    }
+
+    private val _state = MutableStateFlow(State())
+    val state: StateFlow<State> = _state.asStateFlow()
+
+    private var debounceJob: Job? = null
+
+    /** 单调递增的请求序号。用于识别乱序响应。 */
+    private var nextSequence: Long = 0L
+
+    /**
+     * 关键词变更入口。空串立即清空结果并取消所有在途 debounce；
+     * 非空串等 [debounceMillis] 之后触发一次搜索。
+     */
+    fun setKeyword(value: String) {
+        if (_state.value.keyword == value) return
+        _state.value = _state.value.copy(keyword = value)
+        debounceJob?.cancel()
+        if (value.isEmpty()) {
+            // 清空回到初始态，但保留 nextSequence 单调性
+            _state.value = State()
+            return
+        }
+        debounceJob = viewModelScope.launch {
+            delay(debounceMillis)
+            triggerSearch(value)
+        }
+    }
+
+    /**
+     * 显式强制触发（用户点击键盘搜索键、外部调用等）。跳过 debounce，
+     * 但仍会取消上一个 debounce job 保持串行。
+     */
+    fun triggerNow() {
+        val keyword = _state.value.keyword
+        if (keyword.isEmpty()) return
+        debounceJob?.cancel()
+        triggerSearch(keyword)
+    }
+
+    /** 释放 debounce job，供 Activity onDestroy 之外的显式清理。 */
+    fun reset() {
+        debounceJob?.cancel()
+        _state.value = State()
+    }
+
+    private fun triggerSearch(keyword: String) {
+        val mySeq = ++nextSequence
+        _state.value = _state.value.copy(
+            isLoading = true,
+            errorCode = null,
+            errorMessage = null,
+            retryAfterSec = 0,
+            sequence = mySeq,
+        )
+        val req = SearchGlobalGroupsReq(keyword = keyword, sequence = mySeq)
+        searchCaller.call(req) { outcome -> handleResponse(mySeq, outcome) }
+    }
+
+    private fun handleResponse(
+        reqSeq: Long,
+        outcome: ChannelSearchOutcome<SearchGlobalGroupsResp>,
+    ) {
+        // 乱序防护 1：本地已发出更新的请求 → 丢弃此响应
+        if (reqSeq != nextSequence) return
+        // 乱序防护 2：keyword 已被用户清空 → 丢弃
+        if (_state.value.keyword.isEmpty()) return
+
+        if (outcome.ok) {
+            val body = outcome.data ?: return
+            val data = body.data
+            // 乱序防护 3：服务端 echo 回的 sequence 与本次请求不匹配（理论上不应发生，防御性检查）
+            if (data.sequence != reqSeq) return
+            _state.value = _state.value.copy(
+                isLoading = false,
+                groups = data.groups,
+                totalGroups = data.total_groups,
+                totalGroupsApprox = data.total_groups_approx,
+                hasMore = body.pagination.has_more,
+                errorCode = null,
+                errorMessage = null,
+                retryAfterSec = 0,
+            )
+        } else {
+            _state.value = _state.value.copy(
+                isLoading = false,
+                groups = emptyList(),
+                totalGroups = 0L,
+                hasMore = false,
+                errorCode = outcome.errorCode,
+                errorMessage = outcome.errorMessage,
+                retryAfterSec = outcome.retryAfterSec,
+            )
+        }
+    }
+
+    override fun onCleared() {
+        debounceJob?.cancel()
+        super.onCleared()
+    }
+
+    companion object {
+        const val DEBOUNCE_MS = 300L
+    }
+}
+
+/** filters 触发门辅助：目前 [GlobalSearchViewModel] 只用 keyword 场景，此处保留兼容占位。 */
+internal fun GlobalSearchFilters?.orEmptyForL1(): GlobalSearchFilters =
+    this ?: GlobalSearchFilters()
--- a/wkuikit/src/main/java/com/chat/uikit/search/remote/drill/GroupDrillViewModel.kt
+++ b/wkuikit/src/main/java/com/chat/uikit/search/remote/drill/GroupDrillViewModel.kt
@@ -1,0 +1,222 @@
+/*
+ * Copyright 2026-present OctoIM contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.chat.uikit.search.remote.drill
+
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.viewModelScope
+import com.chat.base.search.channel.ChannelSearchOutcome
+import com.chat.base.search.channel.ChannelSearchUiAction
+import com.chat.base.search.channel.dto.CombinedHit
+import com.chat.base.search.channel.dto.CursorList
+import com.chat.base.search.channel.uiAction
+import com.chat.base.search.global.SearchGlobalMessagesModel
+import com.chat.base.search.global.dto.GlobalChannelRef
+import com.chat.base.search.global.dto.GlobalSearchFilters
+import com.chat.base.search.global.dto.SearchGlobalMessagesReq
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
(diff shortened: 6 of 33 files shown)


Jerry-Xin (dismissed):
Summary: The PR is in-scope for `octo-android`, but the new search ViewModels have a race that can render and navigate with results from an older keyword.

🔴 Blocking

🔴 Critical — Stale responses are accepted after the keyword changes but before the debounced replacement request is issued.
Files: `wkuikit/src/main/java/com/chat/uikit/search/remote/GlobalSearchViewModel.kt:108`, `wkuikit/src/main/java/com/chat/uikit/search/remote/GlobalSearchViewModel.kt:140`, `wkuikit/src/main/java/com/chat/uikit/search/remote/GlobalSearchViewModel.kt:153`; `wkuikit/src/main/java/com/chat/uikit/search/remote/drill/GroupDrillViewModel.kt:127`, `wkuikit/src/main/java/com/chat/uikit/search/remote/drill/GroupDrillViewModel.kt:154`, `wkuikit/src/main/java/com/chat/uikit/search/remote/drill/GroupDrillViewModel.kt:176`.

Scenario: request `seq=1` for keyword `foo` is in flight; the user types `bar`; `setKeyword()` updates `state.keyword` immediately, but `nextSequence` is not incremented until the 300ms debounce fires. If the `foo` response returns during that debounce window, `handleResponse()` sees `reqSeq == nextSequence` and `keyword` non-empty, so it accepts `foo` results under the current `bar` state. In `GlobalActivity`, those stale buckets are mapped with `state.keyword`, so the user can tap a bucket from `foo` and open L2 with keyword `bar`. The same race exists in `GroupDrillViewModel`, including stale load-more responses during a pending keyword reset.

Fix by tying every request to the keyword/generation that created it: either increment a generation on every keyword mutation and reject older generations, or capture `requestKeyword` and require `state.keyword == requestKeyword` in `handleResponse()`. Also consider clearing old results when a new keyword is entered.

💬 Non-blocking

🟡 Warning — `GroupDrillAdapter` uses `java.time.OffsetDateTime` even though the project has `minSdkVersion = 23` and no core library desugaring configured.
`wkuikit/src/main/java/com/chat/uikit/search/remote/drill/GroupDrillAdapter.kt:84`.
There is already a project utility, `wkbase/src/main/java/com/chat/base/search/channel/Rfc3339.kt`, explicitly added to avoid `java.time` under this minSdk/desugaring setup. Use that parser here to keep API 23-25 behavior consistent.

🟡 Warning — L1 truncation is tracked but not surfaced.
Files: `wkuikit/src/main/java/com/chat/uikit/search/remote/GlobalActivity.kt:76`, `wkuikit/src/main/java/com/chat/uikit/search/remote/GlobalActivity.kt:390`.
The comment says `state.hasMore` will append a “narrow your search” hint, but `rebuildList()` ignores it. If the server returns only the top `maxGroups`, users get no indication that results are incomplete.

✅ Highlights

The new request body builders omit empty filter fields cleanly and have focused serialization tests. The L2 adapter correctly avoids rendering server `<mark>` as raw HTML, and the service layer follows the existing Retrofit/search endpoint style.

Tests were not run locally because this checkout has no Java runtime available (`Unable to locate a Java Runtime`).

wkbase/src/main/java/com/chat/base/search/global/SearchGlobalGroupsService.kt
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * 全局聚合搜索 L1 端点（`POST /v1/messages/_search_global_groups`）。
+ *
+ * `X-Space-Id` / `token` 由 [com.chat.base.net.CommonRequestParamInterceptor] 自动注入。
+ * 与 [com.chat.base.search.channel.ChannelSearchService] 共享错误信封 [com.chat.base.search.channel.dto.SearchErrorCode]
+ * 与 401 auth 中间件行为。
+ */
+interface SearchGlobalGroupsService {
+    @POST("messages/_search_global_groups")
+    fun searchGlobalGroups(@Body body: JSONObject): Observable<SearchGlobalGroupsResp>
```
OctoBoooot, line 34: **[blocker]** This returns `Observable<SearchGlobalGroupsResp>` deserialized by FastJson field-name reflection, but `app/proguard-rules.pro` has no `-keep` for `com.chat.base.search.global.dto.**` (it keeps `channel.dto.**`, `emoji`, `SearchUserEntity` — not this new package), and the proguard file isn't in this PR. In a release/R8 build the DTO fields (`GroupBucket.channel_id`, `group_name`, …) get renamed → L1 parses to empty/wrong types or throws ClassCastException. Same class as #83/#92/#94 (the proguard file even documents '第 3 次踩坑'). L2 survives because it reuses the kept `channel.dto`, so this only breaks L1 and is easy to miss in a debug smoke test. Fix: add `-keep class com.chat.base.search.global.dto.** { *; }`.

wkbase/src/main/java/com/chat/base/search/global/dto/GroupBucket.kt
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * 语义（见 `04-aggregation-api-spec.md` §2.4）：
+ *  - presence 精确（K2=500 内 100% 精确）：列出的桶点进 L2 至少 1 条可见，不会"点进去空"
+ *  - [match_count] / [match_count_approx]=true：pre-visibility 近似，UI 显示 "约 N 条"
+ *  - [preview] 每条已做可见性过滤，可直接渲染，[MessageHit.snippet] 已带 `<mark>` 高亮
+ *  - DM 桶：channel_type=1，group_name 为对端用户名，无 parent_group_no / thread_*
+ *  - 子区桶：channel_type=5，parent_group_no + thread_id + thread_name 均非空
+ */
+class GroupBucket {
+    lateinit var channel_id: String
```
OctoBoooot: **[major]** `lateinit var channel_id` (and similarly `CombinedHit.result_type`/`sorted_at`, `MessageHit.message_id`/`sender_id`) on an untrusted server DTO crashes on a missing field: a response omitting `channel_id` (e.g. `{"channel_type":2,"group_name":"x"}`) throws `UninitializedPropertyAccessException` on first read. Consistent with the pre-existing `channel.dto` convention so not a regression, but a real crash path on hostile/misconfigured server output. Prefer `var channel_id: String = ""` + validation, or guard at read sites. Parse tests are happy-path only and don't cover this.

wkbase/src/main/java/com/chat/base/search/global/SearchGlobalMessagesService.kt
```diff
@@ -0,0 +1,38 @@
+/*
```
OctoBoooot, line 1: **[major]** No response-body size cap on the two new untrusted aggregation endpoints. `_search_global_groups` / `_search_global_messages` go through raw Retrofit + the shared FastJson converter (`value.string()` buffers the whole body) with no maxBody guard. #94's `EmojiManifestBodyLimitInterceptor` is path-scoped to `common/emojis` and does NOT cover these, so a hostile/misconfigured server returning a multi-hundred-MB `groups[].preview[]` (chunked, no Content-Length) OOMs before any field cap. Add a path-scoped body limit for the two `_search_global_*` paths, mirroring the emoji interceptor.

wkuikit/src/main/java/com/chat/uikit/search/remote/drill/GroupDrillViewModel.kt
```diff
@@ -0,0 +1,222 @@
+/*
```
OctoBoooot, line 1: **[minor]** load-more appends `_state.value.hits + body.data` with no cursor-advance or de-dup check. A server returning `has_more=true` with an unchanged `next_cursor` (or a retry) re-fetches and duplicates a page. Consider de-dup by message id on append, or assert `next_cursor != currentCursor` before issuing the next page. Not currently tested.

wkuikit/src/main/java/com/chat/uikit/search/remote/GlobalSearchViewModel.kt
```diff
@@ -0,0 +1,202 @@
+/*
```
OctoBoooot, line 1: **[minor]** The out-of-order guards and the load-more re-entry flag rest entirely on the main-thread single-owner invariant (callbacks land on main via RxJava `observeOn`) with a plain `var nextSequence` and `_state.value = copy(...)`. Correct today, but fragile — if a callback ever moves off-main this becomes a check-then-act race with no memory barrier. Make the invariant explicit (a comment, or `AtomicLong`). Also: `triggerNow()` re-issues a request for the current keyword even when an identical debounced one is already in flight (duplicate request; defeats the stated rate-limit intent though the later result wins). The `isLoadingMore` re-entry guard is untested.

wkbase/src/main/java/com/chat/base/search/global/SearchGlobalGroupsBodyBuilder.kt
```diff
@@ -0,0 +1,74 @@
+/*
```
OctoBoooot, line 1: **[praise]** No query-injection surface — keyword/cursor/filters/sender_ids are all set as typed JSON values on `JSONObject`/`JSONArray`, never string-concatenated, and the body-builder tests are genuinely rigorous (omit-when-empty logic, snake_case key mapping, channel_types serialized as Int not Byte, byte-exact cursor passthrough). Combined with the verified additive DTO changes, the backward-compat and injection posture here is solid.

OctoBoooot (changes requested):
## Review: feat(search): 全局搜索接入服务端聚合 L1 + L2 深度浏览 (#95)

**Verdict:** Request changes.
One release-build blocker + two untrusted-input majors, all verified against source at c6580a1a. The feature is well-built — backward-compat holds, no query injection, strong body-builder tests, correct fallback — but the proguard gap will silently break L1 in release, which is exactly the class this repo has hit three times.
CI: `check-sprint` red is governance; `build` was pending at review time. Verdict rests on the code findings regardless.

Reviewed by cluster (concurrency / data-layer / UI-compat), each blocker+major adversarially verified.

## Parity checked
The PR claims "向后兼容频道内搜索." Diffed `MessageHit.kt` / `FileHit.kt` base→head: the additions are strictly additive — every pre-existing field is unchanged in name/type/nullability, and the new fields carry non-null defaults (`channel_type: Byte = 0`, `channel_id: String = ""`), so FastJson leaves them at default when absent and existing channel-scoped parsing (`_search`/`_search_files`/`_search_all`, PR #83) is unaffected. Existing consumers (`MessageHitMapper`, `ChannelCombinedHitAdapter`) never read the new fields. Backward-compat claim holds.

### Blockers
- `app/proguard-rules.pro` — **no `-keep` for `com.chat.base.search.global.dto.**`, and the file isn't in this PR.** `SearchGlobalGroupsService` deserializes `Observable<SearchGlobalGroupsResp>` by FastJson field-name reflection (`GroupBucket` etc. are field-mapped, incl. `lateinit channel_id`). In a release/R8 build those fields get renamed → L1 (`_search_global_groups`) parses to empty/wrong types or throws CCE. `channel.dto.**`, `emoji`, and `SearchUserEntity` are all kept; this new package is not. Same bug class as #83/#92/#94 — the proguard file even comments "第 3 次踩坑." Fix: add `-keep class com.chat.base.search.global.dto.** { *; }`. (L2 reuses the kept `channel.dto`, so only L1 breaks — easy to miss in a debug-only smoke test.)

### Major
- No response-body size cap on the two new untrusted aggregation endpoints. `_search_global_groups` / `_search_global_messages` go through raw Retrofit + the shared FastJson converter (`value.string()` buffers the whole body) with no maxBody guard. #94's `EmojiManifestBodyLimitInterceptor` is **path-scoped to `common/emojis`** and does not cover these. A hostile/misconfigured server returning a multi-hundred-MB `groups[].preview[]` (chunked, no Content-Length) OOMs before any field cap. Same gap #94 was filed to fix; add a path-scoped body limit for the two `_search_global_*` paths.
- `lateinit` / non-null-String ID fields on untrusted DTOs crash on missing server fields. `GroupBucket.channel_id` (lateinit), `CombinedHit.result_type`/`sorted_at` (lateinit), `MessageHit.message_id`/`sender_id`/etc. — a response omitting any (e.g. `{"data":[{"message":{"snippet":"x"}}]}`) throws `UninitializedPropertyAccessException` on first read. Consistent with the pre-existing `channel.dto` convention (not a regression), but a real crash path on hostile input; prefer `var x: String = ""` + validation, or guard at read sites. Parse tests are happy-path only and don't cover this.

### Minor
- `GroupDrillViewModel` load-more append (`_state.value.hits + body.data`) has no cursor-advance / de-dup check — a server returning `has_more=true` with an unchanged `next_cursor` (or a retry) duplicates a page. Consider de-dup by message id or asserting the cursor advanced.
- `GlobalSearchViewModel.triggerNow()` re-issues a request for the current keyword even when an identical debounced request is already in flight — a duplicate request that defeats the file's stated rate-limit intent (the later result wins, so no correctness bug). Add an "already searching this keyword" guard.
- The double-fire load-more protection (`isLoadingMore`) and the out-of-order guards rest entirely on the main-thread single-owner invariant with a plain `var nextSequence` / `_state.value = copy(...)` — correct today (callbacks land on main via RxJava `observeOn`), but fragile if a callback ever moves off-main. Make the invariant explicit (comment or `AtomicLong`). The `isLoadingMore` re-entry guard is also untested.
- `GlobalActivity` L79 comment references a `state.hasMore` "narrow your search" hint that `rebuildList()` never emits — stale/aspirational comment or a missing affordance on truncated results.

### Praise (verified)
- **No query injection** — both body builders set keyword/cursor/filters/sender_ids as typed JSON values on `JSONObject`/`JSONArray`, no string concatenation; FastJson escapes on serialize.
- **Concurrency guards are real** — the monotonic-sequence + keyword-cleared + echo-mismatch triple guard genuinely drops stale/out-of-order responses (verified the classic "type a → seq1 slow, type ab → seq2 fast" interleaving is handled), debounce jobs are cancelled on every new input / reset / onCleared, and the load-more-failure branch correctly preserves already-loaded hits. Each has a mutation-real test.
- **Backward-compat + fallback done right** — additive DTO changes (parity section above); L1 server-first correctly falls back to local `WKIM.msgManager.search` on 5xx/offline/bare-404 via `FALLBACK_TO_LOCAL`, with no double-render (old `/search/global` messages segment no longer consumed) and a clean 3-segment cache rebuild. Subarea (channel_type=5) avatar degrades safely to the parent group with no NPE when the parent is uncached; `GroupDrillActivity` is correctly non-exported. Body-builder tests are rigorous (omit-logic, snake_case mapping, Int-not-Byte channel_types, byte-exact cursor).

- The two known-residuals the author documented (server returns UI-hidden default-group conversations; single-hit jump to ChatActivity can land on a locally-absent message) are correctly deferred with a server-side / `_search_around` follow-up noted.

_Verdict rests on the proguard blocker + two majors; even fixed, APPROVE also needs `build` green (check-sprint is governance)._

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #95 (octo-android)

## Summary

This PR wires the global search screen to two new server-side aggregation endpoints: L1 (`_search_global_groups`) for conversation-level bucket summaries, and L2 (`_search_global_messages`) for per-conversation drill-down with cursor pagination. It adds two ViewModels, a new `GroupDrillActivity` with adapter and layout, DTOs/body builders/services for both endpoints, and modifies `GlobalActivity` and `GlobalAdapter` to consume L1 results. The architecture is clean — ViewModels are decoupled from IMSDK for testability, sequence-based race protection is correctly implemented, and error fallback to local IMSDK search is preserved. Comprehensive unit tests cover body builders, response parsing, and ViewModel state machines.

## Verification

- ✅ **Sequence-based race protection** — `GlobalSearchViewModel.nextSequence` monotonically increments on each `triggerSearch`; stale responses are discarded via `reqSeq != nextSequence` check. Same pattern in `GroupDrillViewModel` with reset vs loadMore correctly sharing/incrementing sequence.
- ✅ **Cursor pagination integrity** — `GroupDrillViewModel.loadMore()` guards on `hasMore && !isLoading && !isLoadingMore && nextCursor != null`. On reset, hits are replaced; on loadMore, hits are appended. Failure on loadMore preserves existing hits.
- ✅ **Data flow: L1 → UI** — `SearchGlobalGroupsModel.searchGroups()` → callback → `GlobalSearchViewModel.handleResponse()` → `State.groups` → `GlobalActivity.observeChatRecordsState()` → `bucketToDataVO()` → `chatRecords` → `rebuildList()` → adapter. Chain verified end-to-end.
- ✅ **Data flow: L2 → UI** — `SearchGlobalMessagesModel.searchMessages()` → callback → `GroupDrillViewModel.handleResponse()` → `State.hits` → `GroupDrillActivity.observeState()` → `adapter.setList()`. Chain verified end-to-end.
- ✅ **Error fallback** — `GlobalActivity` maps `FALLBACK_TO_LOCAL` to IMSDK `localMessagesAsDataVO()`; other error codes show toast and clear chat records. `GroupDrillActivity` has `lastToastedErrorCode` dedup.
- ✅ **Backward compatibility** — `FileHit.channel_id`/`channel_type` and `MessageHit.channel_type` additions default to empty/0, not breaking existing channel-search deserialization.
- ✅ **Thread safety** — `StateFlow` updates are thread-safe; `nextSequence` only mutated from `viewModelScope` (Main dispatcher); RxJava callbacks from background threads only touch `StateFlow`.

Static analysis only at head `c6580a1a278edfc3fe9298ed6041adf8f27629bb`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. Four P2 items and two nits below.

### P2 — Duplicate error toast on state re-collection (`GlobalActivity.kt:197`)

`observeChatRecordsState()` calls `showErrorToast(state)` on every `StateFlow` emission where `uiAction()` is non-null. Unlike `GroupDrillActivity` which deduplicates with `lastToastedErrorCode`, `GlobalActivity` has no such guard. When the lifecycle drops below STARTED and resumes (e.g., returning from GroupDrillActivity, config change, or background/foreground), `repeatOnLifecycle` re-collects the last state and fires the toast again — same error message, same user, no new request.

```kotlin
// Fix sketch: add dedup mirroring GroupDrillActivity
private var lastToastedErrorCode: String? = null
// in collector:
if (state.errorCode != lastToastedErrorCode) {
    lastToastedErrorCode = state.errorCode
    showErrorToast(state)
}
```

### P2 — `bucketToDataVO` indexes `preview[0]` gated by `match_count` not `preview.isNotEmpty()` (`GlobalActivity.kt:231`)

The condition `bucket.match_count <= 1L && bucket.preview.isNotEmpty()` correctly avoids crash when preview is empty with count ≤ 1. However, the real concern is the opposite edge: if the server returns `match_count >= 1` with an empty `preview` list (contract violation), the `match_count <= 1` branch would attempt `preview[0]` and throw `IndexOutOfBoundsException`. Consider restructuring to check `preview.isNotEmpty()` as the primary gate.

```kotlin
val text = if (bucket.preview.isNotEmpty() && bucket.match_count <= 1L) {
    stripMarkTags(bucket.preview[0].snippet)
} else { ... }
```

### P2 — `GroupDrillAdapter.convert()` does not reset views for unknown `result_type` (`GroupDrillAdapter.kt:43`)

The `convert()` method has two branches: `item.isFile()` and `item.isMessage()`. If a `CombinedHit` arrives with an unrecognized `result_type` (or both `file` and `message` null), neither branch executes. Because `BaseQuickAdapter` reuses `ViewHolder` instances, all child views (`nameTv`, `timeTv`, `contentTv`, `fileIcon`, `avatar`) retain content from the previously bound item. The user sees a phantom duplicate row.

Add an `else` branch that clears all views or sets them to a safe placeholder.

### P2 — `GroupDrillActivity` missing `android:resizeableActivity="true"` (`AndroidManifest.xml:115`)

The parent `GlobalActivity` declares `android:resizeableActivity="true"` plus full `configChanges`, but the new `GroupDrillActivity` omits both. On devices in split-screen or freeform mode, navigating from `GlobalActivity` to `GroupDrillActivity` may cause the system to destroy and recreate the activity or refuse the resize, resulting in a jarring transition. Align with the parent activity's manifest declaration.

### Nit — `loadMore()` silently no-ops when `has_more=true` but cursor is empty (`GroupDrillViewModel.kt:150`)

If the server returns `has_more=true` with `next_cursor=""`, `nextCursor` becomes null (via `takeIf { it.isNotEmpty() }`), and `loadMore()` returns without firing a request. The SmartRefreshLayout footer stays in loading state until the next `finishLoadMore()` call, which never comes since no request is dispatched. Extremely unlikely under the documented server contract, but a defensive `finishLoadMore()` in the early-return path would close the gap.

### Nit — `GlobalAdapter` itemType 4 does not use DiffUtil (`GlobalAdapter.kt:85`)

`GlobalAdapter.setList()` replaces the entire list without DiffUtil. For the chat records segment which can contain many L1 buckets, this causes full rebind and visual flicker on each keystroke debounce cycle. Not a regression from existing behavior, but worth noting for future optimization.

## Data Flow Trace

**L1 (Groups)**: `GlobalActivity.afterTextChanged` → `viewModel.setKeyword()` → debounce 300ms → `triggerSearch()` → `SearchGlobalGroupsModel.searchGroups()` → Retrofit POST → `handleResponse()` checks `reqSeq == nextSequence` → updates `State.groups` → `observeChatRecordsState()` collector → `bucketToDataVO()` maps each `GroupBucket` to `DataVO(LOCAL_MSG)` → `rebuildList()` merges with friends/groups/searchShortcut → adapter renders.

**L2 (Messages)**: `GroupDrillActivity.initView()` → `viewModel.init(channelId, channelType, keyword)` → `triggerSearch(keyword, cursor=null, isReset=true)` → `SearchGlobalMessagesModel.searchMessages()` → Retrofit POST → `handleResponse()` checks `reqSeq == nextSequence` → updates `State.hits` (replace on reset, append on loadMore) → `observeState()` collector → `adapter.setList(state.hits)` → `GroupDrillAdapter.convert()` renders MESSAGE or FILE hits with mark-to-spannable highlighting.

**Click routing**: `GlobalActivity` LOCAL_MSG click → `messageCount==1 && orderSeq>0` ? direct chat positioning : `GroupDrillActivity` (L2 drill-down). Since L1 server results always have `orderSeq=0`, single-hit buckets go to L2 — correct behavior.

## Blind-spot Checklist

- **C1 (dual-path parity)**: N/A — no paired add/remove or create/delete paths.
- **C2 (control-flow ordering)**: Clear — `setKeyword` cancels pending debounce before launching new one; `triggerNow` also cancels debounce. No nested reuse of `triggerSearch` from multiple entry points that could double-fire.
- **C3 (authorization boundary)**: Clear — `X-Space-Id` and `token` injected by `CommonRequestParamInterceptor` for both L1 and L2 endpoints. No new endpoints bypass the interceptor.
- **C4 (authorization lifecycle)**: N/A — no auth state changes within the search flow.
- **C5 (build/note)**: N/A — no build system changes affecting runtime behavior (only test dependency added).
- **C6 (governance/docs)**: N/A — no governance or policy changes.

## Cross-round Blocker Review

N/A — first review pass.

## Verdict: APPROVED

Well-structured implementation with solid race-condition protection, clean separation of concerns, and good test coverage. The P2 items are all defensive-coding improvements (toast dedup, adapter view reset, preview indexing safety, manifest alignment) — none block the feature's correctness under normal operating conditions.

[Octo-Q] verdict: APPROVE — No P0/P1 blockers. Four P2 findings are non-blocking quality improvements (toast dedup, preview index safety, adapter stale-view reset, manifest resizeableActivity). Feature is functionally correct under the documented server contract.

yujiawei (dismissed):
# Code Review — PR #95 (octo-android)

Reviewed at head SHA `c6580a1a278edfc3fe9298ed6041adf8f27629bb` against merge-base `86cc27e2`. This adds global search: server-side L1 aggregation (`_search_global_groups`) in the conversation-list search, plus an L2 deep-browse page (`_search_global_messages`) with cursor pagination.

Overall this is a solid, well-structured change. The ViewModels are cleanly testable (JVM-only, no IMSDK), the Rx callbacks are confined to the main thread (`observeOn(AndroidSchedulers.mainThread())`), so the `sequence`/`_state` access is race-free at the threading level. Unit coverage (570 tests, +30 new) is genuinely thorough. No correctness, security, data-loss, or build-breaking issues found — nothing here must be fixed before merge. The items below are all non-blocking (P2 / nits); a few are worth a fast follow-up.

## Spec compliance

The diff implements exactly what the linked issue describes — L1 broad aggregation with local fallback, L2 deep browse (server-only), sub-thread avatar fallback to the parent group, and the `"parent · thread"` display name. No unrequested features, no scope drift, no contradicted approach. The three-segment cache rebuild correctly drops the old server `messages` segment so a message is not shown twice.

One comment/behavior mismatch (see P2-1 below) is the only spec-adjacent gap.

### P2-1 — Promised "narrow your search" hint for truncated L1 results is not implemented *(consensus: two independent review passes + manual)*
`GlobalActivity.kt:80` comment states: *"命中群 > maxGroups 时通过 state.hasMore 由 [rebuildList] 追加'缩小范围'提示项"*, but `GlobalSearchViewModel.State.hasMore` is **never read** anywhere in `GlobalActivity` — `rebuildList()` (`GlobalActivity.kt:346`) only appends the chat-records section. When the server truncates buckets at `maxGroups` (200), the user gets silently incomplete results with no cue to refine. Either implement the hint or drop the comment so it does not read as done.

### P2-2 — Debounce-window stale-response race in both ViewModels
`GlobalSearchViewModel.setKeyword()` (`GlobalSearchViewModel.kt:108`) updates `State.keyword` immediately but does **not** bump `nextSequence` until the debounced `triggerSearch()` fires 300 ms later. If a response for the *previous* keyword lands during that window, `handleResponse()` (`:153`) still sees `reqSeq == nextSequence` and a non-empty keyword, so it accepts and publishes stale buckets under the new query. `GroupDrillViewModel` (`:127`, `:176`) has the same shape, and there the stale response is a `loadMore` that gets **appended** (`hits + body.data`) onto the list before the new reset request replaces it. It is self-correcting (the next request overwrites within ~300 ms), so it is a transient flash rather than a persistent wrong state — hence P2, not a blocker — but it does undercut the "3-layer race protection" the PR claims. Cheap fix: increment the generation counter inside `setKeyword` on every keyword change, or also compare the response's request-keyword in `handleResponse`.

### P2-3 — Reset search does not clear prior results while the new request is in flight
`GlobalSearchViewModel.triggerSearch()` (`:140`) and `GroupDrillViewModel.triggerSearch(isReset=true)` (`:154`) set `isLoading=true` but leave the previous `groups`/`hits`/`hasMore`/`nextCursor` populated until the response arrives. Because the Activity re-renders on every emitted state, the user can briefly see (and tap) previous-query rows under the new keyword. Consider clearing the result fields when entering the loading state for a reset.

### P2-4 — `<mark>` highlighting stripped, losing server-side synonym/fuzzy matches
`bucketToDataVO()` (`GlobalActivity.kt:295`) calls `stripMarkTags()` on the server snippet and relies on client-side `highlightKeyword(text, keyword)` in `GlobalAdapter`. When the server highlights a token the raw keyword does not literally match (stemming/synonym/fuzzy — e.g. keyword `run`, server marks `running`), the client re-highlight finds nothing and the match is shown unhighlighted. The L2 adapter (`GroupDrillAdapter.markToSpannable`) correctly preserves the server marks; L1 diverges. Prefer reusing the server `<mark>` spans in L1 too for consistency.

### P2-5 — `java.time.OffsetDateTime` on minSdk 23 without core-library desugaring
`GroupDrillAdapter.formatTime()` (`:88`) uses `java.time.OffsetDateTime.parse(...)`. `minSdk=23`, `coreLibraryDesugaring` is not enabled, and the project already ships `Rfc3339.kt` whose header explicitly says it avoids `java.time` for this exact reason. On API 23–25 this throws `NoClassDefFoundError` — but it is inside `catch (t: Throwable)`, so it degrades to showing the raw RFC3339 string rather than crashing. Still, on those devices every L2 timestamp renders as an ISO string. Reuse `Rfc3339.toEpochSeconds()` (already the pattern in `ChannelCombinedHitAdapter`) instead.

### P2-6 — RecyclerView binding robustness on malformed rows
- `GroupDrillAdapter.convert()` (`:45`) has no `else` branch: a `CombinedHit` with an unexpected `result_type` (or a null nested `message`/`file`) leaves the recycled row showing stale avatar/name/time from a prior item. Add an explicit clear/skip.
- `sent_at` is `lateinit`/non-null in `MessageHit`/`FileHit`; if the server ever omits it for a global hit, `formatTime(m.sent_at)` (`:65`) throws `UninitializedPropertyAccessException` during bind before the try/catch can help. This matches existing channel-search adapters (same assumption), so it is a latent codebase-wide pattern, not a regression — noting for hardening.

### P2-7 (nit) — Dead code
`GlobalSearchFilters.canTriggerL1()` (`GlobalSearchFilters.kt:52`) and `GlobalSearchFilters?.orEmptyForL1()` (`GlobalSearchViewModel.kt:201`) are never called from production code. Either wire `canTriggerL1` into the trigger gate (it would also close part of P2-2/P2-3) or remove them.

## Coverage / blind spots (could not verify from the diff alone)
- **Server contract fidelity**: whether the backend ever emits `"groups": null` / `"preview": null` (vs `[]`). FastJSON would overwrite the Kotlin `emptyList()` defaults with `null`, NPE-ing `state.groups.map{}` / `bucket.preview.isNotEmpty()`. The new DTOs follow the existing `CursorList`/`AroundResult` convention, so this is a pre-existing whole-codebase assumption, not introduced here — but it depends on the server always sending arrays.
- **DM bucket avatar/name**: for `channel_type=1` buckets the display name and avatar key come straight from `channel_id`/`group_name`; correct rendering depends on runtime IMSDK/avatar behavior not visible in the diff.
- **Sub-thread `____` composite channel_id** parsing via `SpaceFilter.extractParentGroupId` — relies on the server always using that separator for `channel_type=5`.
- Actual on-device fallback (`FALLBACK_TO_LOCAL`), pagination on a real >maxGroups dataset, and soft-keyboard/scroll behavior were not exercised (static review only).

OctoBoooot (approved):
## Review: feat(search): 全局搜索接入服务端聚合 L1 + L2 深度浏览 (#95) — re-review @ 63e403ac

**Verdict:** Approve.
Every blocking and major finding from the prior round (both my review and the peer round) is fixed at this head, each verified against source. Code CI is green — build ✅, osv-scanner ✅, secret-scan ✅. Fully mergeable on code grounds.
CI: only `check-sprint` red — governance (Sprint link), not code CI; `mergeStateStatus` stays BLOCKED until a maintainer clears it.

Re-reviewed the delta from my prior head c6580a1a → 63e403ac (11 files).

## Parity checked
Prior round confirmed the `MessageHit`/`FileHit` additions are strictly additive; this delta only touches `GroupBucket.channel_id` (lateinit → `var = ""`), which is a safe widening (nullable-safe default), no existing consumer affected.

### Prior findings → resolved (verified)
- **Proguard blocker** (release-build L1 break) — `-keep class com.chat.base.search.global.dto.** { *; }` added, with a comment citing this review. R8 will no longer rename the FastJson-mapped L1 DTO fields.
- **`java.time.OffsetDateTime` on minSdk 23** (peer blocker) — `GroupDrillAdapter.formatTime` now uses the repo's existing `Rfc3339.toEpochSeconds` (SimpleDateFormat) instead of `OffsetDateTime`, with a comment noting minSdk/desugaring. No more `NoClassDefFoundError` on API 23–25.
- **Response-body OOM** (my major) — new `SearchGlobalBodyLimitInterceptor` (2 MB cap, Content-Length fast path + `peekBody(MAX+1)` for lying/absent length), scoped to both `_search_global_groups` (L1) and `_search_global_messages` (L2), and **wired innermost after `LogInterceptor`** — correctly applying the exact ordering lesson from PR #94 (BodyLimit must run before LogInterceptor's `body().string()`). Backed by a 174-line test.
- **`lateinit` DTO crash** (my major) — `GroupBucket.channel_id` is now `var = ""` with a documented downstream fallback; a response omitting the field degrades instead of throwing `UninitializedPropertyAccessException`.
- **Debounce-window stale response** (peer 🔴 / my minor) — `handleResponse` now captures `requestKeyword` and drops the response when `state.keyword != requestKeyword`, closing the window where an old-keyword response arrived before `nextSequence` incremented. Applied to both `GlobalSearchViewModel` and `GroupDrillViewModel`, each with a new test.

### Praise
- Clean, complete response to a two-reviewer round: each fix targets the root cause (path-scoped body limit reusing #94's proven pattern *and* its ordering lesson; the generation/keyword binding is the exact fix prescribed; `java.time`→`Rfc3339` reuses the in-repo helper rather than adding desugaring), and every fix ships with a test. The interceptor even carries the "must be innermost" comment so the ordering invariant survives future refactors.

### Out of scope (informational, non-blocking)
- Prior minors that remain optional follow-ups: load-more cursor-advance/de-dup guard, `triggerNow` duplicate-request suppression, and the L1-truncation "narrow your search" UX hint. None blocks.

*Refreshes my prior REQUEST_CHANGES at c6580a1a (auto-dismissed on the head change). All code findings across both rounds resolved; code CI green; merge gated only on the `check-sprint` governance link.*

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #95 (octo-android)

Reviewed at head `63e403ac`. Scope: global search server-side aggregation L1 (`_search_global_groups`) + L2 drill-down (`_search_global_messages`), ViewModels, DTOs, body-size interceptor, adapter updates, and proguard rules.

## Summary

This PR wires the Android global search to two new server-side aggregation endpoints: L1 returns per-group buckets with preview snippets, and L2 returns cursor-paginated mixed hits (message + file) for a selected group/thread/DM. The architecture is clean — ViewModels own debounce + sequence-based race protection, a shared OkHttp interceptor caps response body size, and the UI gracefully falls back to local IMSDK search on network failure. The three-segment list composition (contacts / groups / chat records) is well-isolated with per-segment stale-response guards. Two P2 items found in the L2 drill page.

## Verification

- ✅ **Race protection** — Both `GlobalSearchViewModel` and `GroupDrillViewModel` use monotonically-incrementing `nextSequence` with three-layer guards (seq match, keyword non-empty, keyword match). LoadMore reuses the current sequence so a reset properly invalidates in-flight pagination responses.
- ✅ **Body cap interceptor** — `SearchGlobalBodyLimitInterceptor` correctly matches only the two target endpoints via path-segment comparison, uses Content-Length fast path + peekBody slow path, and is wired after `LogInterceptor` in `OkHttpUtils.java:100-104`.
- ✅ **Proguard keep rules** — `app/proguard-rules.pro:166-168` adds `-keep class com.chat.base.search.global.dto.** { *; }` for FastJson reflection, matching the existing pattern for channel search DTOs.
- ✅ **Error mapping parity** — `SearchGlobalGroupsModel.mapFailure` and `SearchGlobalMessagesModel.mapFailure` are structurally identical, both routing 429→RATE_LIMITED, 5xx→UPSTREAM_UNAVAILABLE, 404→LOCAL_ERROR_NETWORK (triggers local fallback), 400→VALIDATION_FAILED.
- ✅ **Sub-thread avatar handling** — Both `GlobalAdapter` (`:85-91`) and `GroupDrillAdapter` (`showChannelAvatar`) extract parent group ID for COMMUNITY_TOPIC (type=5) channels, preventing avatar load failures on composite channel IDs.
- ✅ **Manifest + layout** — `GroupDrillActivity` registered with proper `configChanges` and `windowSoftInputMode`; layout reuses existing SmartRefreshLayout pattern.
- ✅ **C1 dual-path parity** — L1/L2 body builders share identical `filtersToJson` and `channelRefsToJson` logic. Error mapping duplicated but consistent.
- ✅ **C2 control-flow ordering** — Interceptor chain ordering correct (emoji→search, both after log). Debounce cancel-before-relaunch pattern holds.
- ✅ **C3 authorization boundary** — Both new endpoints inherit `CommonRequestParamInterceptor` auth (X-Space-Id + token). No new auth bypass introduced.

## Findings

No P0/P1 issues; two P2 items below.

### P2 — loadMore resets scroll position (`wkuikit/src/main/java/com/chat/uikit/search/remote/drill/GroupDrillActivity.kt:157`)

`observeState()` calls `adapter.setList(state.hits)` on every state emission, including loadMore completions. BRVAH's `BaseQuickAdapter.setList()` internally clears and re-adds all items, which triggers `notifyDataSetChanged()` and resets the RecyclerView scroll position. After the user scrolls to the bottom and loadMore fetches the next page, the list jumps back to the top — the user must scroll down again to see the newly loaded items.

Fix: distinguish reset vs append in the observer. When the state change is from a loadMore (hits grew, keyword unchanged), use `adapter.addData(newItems)` with only the newly appended items; for reset (keyword change / first load), use `setList`. Alternatively, add a `isReset` flag to `State` so the observer can branch.

### P2 — IME search key does not trigger immediate search (`wkuikit/src/main/java/com/chat/uikit/search/remote/drill/GroupDrillActivity.kt:104`)

`GlobalActivity` (`:97-100`) calls `viewModel.triggerNow()` when the user presses the IME search action key, skipping the 300ms debounce. `GroupDrillActivity` only hides the keyboard, so the user must wait up to 300ms for the debounce to fire. Minor UX inconsistency — pressing the search key on the L2 page should feel at least as responsive as on L1.

Fix: add `viewModel.triggerNow()` (or an equivalent skip-debounce path) before `return@setOnEditorActionListener true`.

- Cursor pagination: `GroupDrillViewModel` passes `nextCursor` from `body.pagination.next_cursor` on loadMore; first page uses `null` cursor. `SearchGlobalMessagesBodyBuilder` omits empty cursor from body.
- `ChannelSearchUiAction` enum mapping covers all server error codes: RATE_LIMITED shows retry countdown, FEATURE_DISABLED shows "搜索功能未启用", FALLBACK_TO_LOCAL triggers local IMSDK search (L1) or toast (L2), VALIDATION_ERROR/BLOCK_NOT_FOUND are silent.
- Three-segment list composition in `GlobalActivity.rebuildList()` correctly orders contacts → groups → search shortcut → chat records, with stale-response guards (`requestKeyword != keyword`) on the old `/search/global` callback.
- Toast deduplication via `lastToastedErrorCode` prevents repeat toasts on `repeatOnLifecycle` state replay.
- `GroupBucket` defaults all fields (no `lateinit`) — FastJson missing-field scenarios degrade gracefully instead of `UninitializedPropertyAccessException`.
- `stripMarkTags` in `GlobalActivity` and `markToSpannable` in `GroupDrillAdapter` handle server `<mark>` tags with different strategies (strip vs span) appropriate to their rendering context.

## Verdict: APPROVED

No P0/P1 blockers. The architecture is solid with proper race protection, body-size guards, error fallback chains, and proguard coverage. The two P2 items (scroll position reset on loadMore, IME key inconsistency) are non-blocking UX improvements.

[Octo-Q] verdict: APPROVE — no P0/P1 found; 2 P2 (loadMore scroll reset, IME key missing triggerNow) are non-blocking UX items.

yujiawei (approved):
# Code Review — PR #95 (octo-android)

Reviewed at head SHA `63e403ac79621ef842aeb2b2d8ce4e3d5a03731e` against merge-base `86cc27e2`. This head is the fix commit that responds to the earlier round of reviews (proguard keep, debounce race, body cap, and three other items), so this is a fresh verdict on the current head — earlier decisions do not carry over.

This PR wires global search to two new server-side aggregation endpoints: **L1** (`_search_global_groups`) for conversation-level bucket summaries in the top search screen, and **L2** (`_search_global_messages`) for a per-conversation deep-browse page (`GroupDrillActivity`) with cursor pagination. It adds two testable JVM-only ViewModels, a response body-size interceptor, DTOs, and a thread-avatar fallback.

Overall this is a well-structured, well-tested change. The ViewModels are cleanly isolated from the IMSDK, the body-builder / response-parse tests are rigorous, there is no query-injection surface (all params are typed JSON values, never concatenated), and the earlier blocking issues have been addressed and verified on this head.

## Verification of the prior-round fixes (all confirmed on head)

- **proguard**: `-keep class com.chat.base.search.global.dto.** { *; }` is present (`app/proguard-rules.pro:166`), alongside the existing `channel.dto.**` keep. This closes the R8-rename-breaks-FastJson-reflection gap.
- **Debounce-window stale race**: `handleResponse` now takes a `requestKeyword` and discards responses whose keyword no longer matches current state (`GlobalSearchViewModel.kt:150`, `GroupDrillViewModel.kt:186`). Regression tests cover both the L1 and L2 in-flight-keyword-change and stale-`loadMore` scenarios.
- **Body cap**: `SearchGlobalBodyLimitInterceptor` mirrors the merged emoji-manifest interceptor, is registered last (innermost, after `LogInterceptor`) so the log interceptor can't buffer first, and uses strict path matching for the two endpoints only. Covered by tests.

## Findings (all advisory — none block merge)

**[P2] Untrusted-DTO deserialization hardening.** The L2 path consumes `CombinedHit` (`result_type`/`sorted_at` are `lateinit`) plus `MessageHit`/`FileHit` (several `lateinit` fields), and L1 `GroupsResult.groups` / `GroupBucket.preview` default to `emptyList()`. FastJson field-reflection can leave a `lateinit` unset (→ `UninitializedPropertyAccessException`) or inject an explicit JSON `null` over a non-null default (→ NPE downstream) if a malformed/partial body slips past the size cap. This is a real robustness gap, **but it is the exact trust model already shipping** in `ChannelCombinedHitAdapter` for `_search_all` against the same DTOs and same server contract — this PR reuses it rather than introducing it. Worth a follow-up sweep (nullable + safe-call, or a sanitize/drop step before adapter binding) across the whole search DTO surface, not a blocker for this PR.

**[P2] Brief staleness during the debounce window.** On each keystroke only `keyword` is updated; the previous `groups`/`hits` (and the legacy `friendsList`/`groupsList` in `GlobalActivity`) remain rendered until the new response arrives (~300ms). This is standard debounce UX and the click target is safe — `DataVO.keyword` is captured at bucket-build time, so tapping a still-visible bucket opens L2 with the keyword that produced it, not a mismatched newer one. If you want to avoid transient "torn" segments, reset the cached segment lists synchronously in `afterTextChanged` before firing the new request.

**[P2] `reset()` / clear does not advance the sequence token.** `GlobalSearchViewModel.setKeyword("")` / `reset()` clears state but leaves `nextSequence` unchanged. A narrow race exists: request for `foo` in flight → user clears → retypes `foo` before debounce fires → the old response passes both `reqSeq == nextSequence` and the keyword check. Very low likelihood; consider bumping an invalidation token on clear.

**[P2] Stale comment + silent L1 truncation.** `GlobalActivity.kt:85` says `state.hasMore` drives a "narrow your search" hint appended by `rebuildList()`, but `rebuildList()` never reads `hasMore`. When hits exceed the server `maxGroups`, the truncation is silent. Either implement the hint or drop the comment.

**[note] L2 has no local fallback — by design.** `SearchGlobalMessagesModel` maps a bare 404 to `FALLBACK_TO_LOCAL`, and `GroupDrillActivity` handles that with a toast only (no local search). This matches the stated strategy (L2 = server-only; local has no equivalent cross-conversation data). Flagging only because the action name `FALLBACK_TO_LOCAL` reads as if a local fallback should occur — a comment at the handler noting the intentional divergence would prevent future confusion.

## Coverage / what I did not verify

- **Runtime behavior**: not run on device; the ViewModel/interceptor logic is verified by the included JVM unit tests only.
- **Server contract**: the DTO field names, the "pass group id → auto-expand group + threads" semantics, cursor opacity, and `maxGroups`/preview budgets are assumed to match the aggregation API spec; not verifiable from this repo.
- **R8 output**: the proguard keep is present and correct by inspection; I did not build a release artifact to confirm the minified DTOs deserialize.
- **`hasMore` truncation UX** and the transient-staleness window are reasoned from code, not observed.

## Verdict

No P0/P1 issues. The prior blocking items are fixed and verified on the current head; all remaining findings are P2 hardening / UX polish, and the most severe-looking DTO-safety and interceptor concerns are pre-existing, already-reviewed patterns this PR reuses rather than regressions it introduces. The change is spec-compliant (no missing, extra, or diverging behavior relative to the described L1/L2 strategy) and appropriately tested.

Approving. Recommend a follow-up to harden the shared search-DTO deserialization surface and to reconcile the stale `hasMore` comment.

Jerry-Xin (approved):
## Re-review @ 63e403ac — APPROVE (code side)

All findings from my prior CR are byte-verified fixed against source at this head. The delta is a clean single fix commit (ahead_by=1, behind_by=0), each file mapping exactly to a flagged finding.

### 🔴 Blockers — resolved
- **`java.time` on minSdk 23 (mine)** — `GroupDrillAdapter` no longer calls `OffsetDateTime`; `formatTime` now uses the repo's existing `Rfc3339.toEpochSeconds` (`SimpleDateFormat`-based), with a comment citing the minSdk/desugaring constraint. Verified `OffsetDateTime` is gone and no core-library desugaring was added (correct — reuse of the established in-repo pattern from #83). No more `NoClassDefFoundError` on API 23–25.
- **Missing proguard `-keep` for L1 DTOs** — `-keep class com.chat.base.search.global.dto.** { *; }` added to `proguard-rules.pro:168`. Closes the release-only R8 field-rename break on the FastJson-reflected L1 DTOs.

### Majors — resolved
- **Debounce/out-of-order race (mine, graded 🔴)** — `handleResponse` now takes `requestKeyword`; guard #3 (`state.keyword != requestKeyword → return`, line 166) fires **before** any `groups =` assignment, so a response for an old keyword arriving inside the 300ms debounce window is dropped instead of rendered. Layered with a sequence guard, empty-keyword guard, and a server-echo `sequence` check. The identical guard set is present in `GroupDrillViewModel`. Both covered by tests.
- **No response-body size cap (OOM)** — new `SearchGlobalBodyLimitInterceptor` (2 MB, Content-Length + bounded peek), wired for both L1 and L2 endpoints, with a test.
- **`lateinit` DTO field crash** — `GroupBucket.channel_id` is now `var = ""` with a documented fallback, avoiding the missing-field crash.

### Notes
- Non-blocking UX polish raised by peers (load-more scroll position, soft-keyboard search key, L1-truncation hint) can be follow-ups.

Each fix reuses an established in-repo pattern and ships with a test. Clean iteration — approving on the code.

PR #123, opened 2026-08-14
fix(db): Bugly ANR/OOM 批次 —— WAL 连接池、缺失索引、同步内存峰值

### LINKED ISSUE #122: [Android] Bugly ANR / OOM 批次修复（SQLCipher 连接池、缺失索引、同步内存峰值）
## 问题

Bugly 上两类线上事故：

- **ANR**：主线程 park 在 `SQLiteConnectionPool` 等连接。单台 Xiaomi 2308CPXD0C 133 次；19 份快照里 7 份卡在 `getMember`（跨 vivo / 小米 / 三星 6 台设备）。
- **OOM**（256MB 顶格）：崩溃栈停在会话同步链路（`ResponseBody.string()` / `insertOrReplaceExtra → queryWithMsgIds`）。

## 根因

1. SQLCipher 连接池被钉死为 1 条 —— `openOrCreateDatabase` 的 11 个重载全部写死 `CREATE_IF_NECESSARY`，拿不到 `ENABLE_WRITE_AHEAD_LOGGING`，事后 `PRAGMA journal_mode=WAL` 改不了池大小
2. 阅后即焚清理定时器空转：`if (isEmpty) return` 写在 `timer.cancel()` 之前，查询恒为空 → 每秒一次全表扫且永不停止。实测单次执行 **89.8 秒**并独占唯一连接
3. `message` 表四条 `max()` 查询无覆盖索引；`@提醒` 查询 `LIKE 'groupNo____%'` 通配符不在末尾 → 索引优化失效，在双层循环里放大成 N 次全表扫
4. RecyclerView bind 路径逐行查库：`channel_members` 完全无缓存、子区会话逐个查
5. conv/sync 响应整包物化（okio 缓冲 + 整包 String + 对象树同时三份）；同步一次性构建全量列表再落盘，峰值与账号会话总数成正比
6. ANR 上报的异常堆栈恒为 `ANRWatchdog.run`，Bugly 把根因完全不同的 ANR 折叠进少数 issue

## 实测收益（真机 A/B）

| 项 | 改前 → 改后 |
|---|---|
| 主线程等连接（后台持事务 4s 期间） | 3102ms → 1ms |
| 阅后即焚轮询 | 60 次/分钟 → 4 次后静默 |
| 成员缓存命中率 | 61% → 96% |
| OOM 峰值 | 132MB → 78MB |
| 收包 + 解析 | +100MB → +15.2MB |
| ANR 分组 | 全折叠 → 按主线程真实堆栈分桶 |

## 影响面

`wkbase` / `wkim` / `wkuikit` 三模块 + 3 个 DB 迁移文件（全部是 `CREATE INDEX IF NOT EXISTS`，不改表结构、不改数据，迁移失败的后果是性能回到改前而非数据损坏）。

## 未定性 / 不在本批

- 「翻历史翻着翻着加载不出来」修了 4 处各自独立成立的缺陷，但**哪一处对应实机那次卡死尚未坐实**（加埋点后未能复现）
- P2-4（sync rebuild 前批量预载 channel/members）要碰 `SpaceFilter`，串 Space 风险大于收益，挂起
- 连接池预热：实测能把冷启动 slow-db 从 10 条降到 0，但属**卡顿级非崩溃级**（最坏 418ms，ANR 门槛 5s）且未在低端机验证，不搭车
- `WKUIKitApplication` 里 `setDebug(true)` 写死的根因（改动会让全 SDK 数百处历史日志在 release 集体闭嘴，需单独立项）

mzanalyticst:
Closes #122

## 修了什么

| 根因 | 修法 |
|---|---|
| SQLCipher 连接池被钉死为 1 条 | `openDatabase` 显式传 `ENABLE_WRITE_AHEAD_LOGGING`，池取 8 |
| 阅后即焚定时器空转（单次执行 89.8s 独占连接） | 轮询改自排程，空结果直接停；查询收窄到 9 列 + 部分索引 |
| `message` 四条 `max()` 无覆盖索引；`@提醒` LIKE 全表扫 × N | 两个复合索引 + synckey 过滤下推；`@提醒` 改一次批量取 + 内存前缀匹配 |
| bind 路径逐行查库 | `channel_members` 共享缓存（含负结果）；子区会话批量化 |
| conv/sync 整包物化 + 一次性构建全量列表 | >256K 字符走 `JSONReader` 流式；每 20 会话一批落盘 + 消费完释放 `recents` |
| ANR 上报堆栈恒为 `ANRWatchdog.run` | `setStackTrace(mainStack)`，Bugly 按业务帧分桶；补冻结守卫 |
| 排序 comparator 里查库 + TimSort 契约违规 | 比较键排序前快照 |
| 落进 Rx 错误管道的 OOM 静默消失 | `BaseObserver` 补一条 `postCatchedException`（每进程 ≤3 条） |

17 笔，`wkbase` / `wkim` / `wkuikit` + 3 个迁移文件（全是 `CREATE INDEX IF NOT EXISTS`，不改表结构与数据）。

## 实测（真机 vivo V2464A）

| 项 | 改前 → 改后 |
|---|---|
| 主线程等连接（后台持事务 4s 期间） | 3102ms → 1ms |
| 阅后即焚轮询 | 60 次/分钟 → 4 次后静默 |
| 成员缓存命中率 | 61% → 96% |
| OOM 峰值 | 132MB → 78MB（3× 载荷改前推算 336MB > 256MB 上限，必崩） |
| 收包 + 解析 | +100MB → +15MB |

## 测试重点

- **老用户覆盖安装**（回退 SP 的 `_db_upgrade_index` 后启动）：3 个迁移全部执行、版本号正确推进
- **全新安装**：表与索引齐备
- 大账号会话同步：列表条数 / 未读 / 置顶 / @红点 与 main 一致
- 翻历史到底、断网恢复、多频道交替翻页
- 成员缓存失效：群内改备注、对方改昵称头像、**切账号**
- 子区未读与父群汇总

## 已知 / 不在本 PR

- 「翻历史加载不出来」修了 4 处各自独立成立的缺陷，**哪一处对应实机那次卡死未坐实**（加埋点后未能复现）
- P2-4（sync rebuild 前批量预载）要碰 `SpaceFilter`，挂起
- 连接池预热：实测可把冷启动 slow-db 从 10 条降到 0，但属卡顿级非崩溃级（最坏 418ms）且未在低端机验证
- `setDebug(true)` 写死的根因：改动会让全 SDK 数百处历史日志在 release 集体闭嘴，需单独立项
- `WKHeapProbe` 与各 `PROBE_*` / `DEBUG_*` 开关是诊断用，全部 debug 门控且默认关，问题闭环后删

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,9 @@ build/
 # Release output
 app/release/
 
+# Bugly 符号表上传工具的产物（buglySymbolAndroid.jar 的工作目录，不是模块）
+/buglybin/
+
 # Local configuration file (sdk path, etc)
 local.properties
 /local.properties
--- a/wkbase/src/main/java/com/chat/base/net/BaseObserver.java
+++ b/wkbase/src/main/java/com/chat/base/net/BaseObserver.java
@@ -23,9 +23,12 @@ import com.chat.base.WKBaseApplication;
 import com.chat.base.config.WKConfig;
 import com.chat.base.endpoint.EndpointManager;
 import com.chat.base.utils.ActManagerUtils;
+import com.tencent.bugly.crashreport.CrashReport;
 import com.xinbida.wukongim.WKIM;
 
 import org.jetbrains.annotations.NotNull;
+
+import java.util.concurrent.atomic.AtomicInteger;
 
 import io.reactivex.rxjava3.core.Observer;
 import io.reactivex.rxjava3.disposables.Disposable;
@@ -36,6 +39,14 @@ import io.reactivex.rxjava3.disposables.Disposable;
  * 服务器返回的状态是在http的状态码上
  */
 public abstract class BaseObserver<T> implements Observer<T> {
+
+    /**
+     * 本进程内最多上报几条被 Rx 吞掉的 {@link Error}。堆已经耗尽时上报本身也要分配内存，
+     * 不能因为要报告 OOM 反而把进程推下去；而且同一次事故通常连着抛好几条，报 3 条足够定性。
+     */
+    private static final int MAX_SWALLOWED_ERROR_REPORTS = 3;
+    private static final AtomicInteger swallowedErrorReports = new AtomicInteger(0);
+
     @Override
     public void onComplete() {
     }
@@ -48,6 +59,7 @@ public abstract class BaseObserver<T> implements Observer<T> {
 
     @Override
     public void onError(@NotNull Throwable e) {
+        reportIfSwallowedError(e);
         ResponseThrowable throwable = ResponseExceptionHandle.getInstance().handleException(e);
         if (throwable != null) {
             String msg = throwable.getMessage();
@@ -70,6 +82,48 @@ public abstract class BaseObserver<T> implements Observer<T> {
                 onFail(throwable.getCode(), msg, errJson);
             }
         }
+    }
+
+    /**
+     * 把落到 Rx 错误管道里的 {@link Error}（主要是 {@link OutOfMemoryError}）补一条 Bugly 上报。
+     *
+     * <p><b>为什么需要</b>：{@link ResponseExceptionHandle#handleException} 的兜底分支
+     * 把任何非 {@code HttpException} 一律转成 {@code code=-1} 的普通请求失败，业务层只看到
+     * 「这次请求没成功」。OOM 走到这里就彻底消失了——**不崩溃、不上报**，线上表现是
+     * 「会话列表突然空了」而 Bugly 里一条记录都没有，等于把排查线索掐断。
+     *
+     * <p>2026-08-14 真机实测：压舱到只剩 3.84MB 可用堆时，会话同步的 OOM 正是这样被吞掉的
+     * （logcat 只有一行 {@code Throwing OutOfMemoryError}，无 FATAL EXCEPTION、无崩溃记录）。
+     * 而同一个 OOM 若抛在 {@code Retrofit.parseResponse} 里则会被原样重抛 → 崩溃 → Bugly 收到
+     * ——同一类事故，能不能被发现全看它恰好抛在哪一帧，不可接受。
+     *
+     * <p><b>刻意不改崩溃语义</b>：这里只补一条 {@code postCatchedException}，然后让流程照常往下走。
+     * 该崩的路径（Retrofit / OkHttp 对 VirtualMachineError 的重抛）不受影响，行为与 main 一致。
+     *
+     * <p>release 也执行 —— 这是线上诊断通道，不是开发期日志。
+     */
+    private static void reportIfSwallowedError(Throwable e) {
+        if (!containsError(e)) return;
+        if (swallowedErrorReports.incrementAndGet() > MAX_SWALLOWED_ERROR_REPORTS) return;
+        try {
+            CrashReport.postCatchedException(e);
+        } catch (Throwable ignored) {
+            // 上报失败不能反过来影响业务的错误处理
+        }
+    }
+
+    /** 顺着 cause 链和第一层 suppressed 找 {@link Error}；OkHttp 会把原始 Throwable 挂在 suppressed 上。 */
+    private static boolean containsError(Throwable e) {
+        Throwable t = e;
+        for (int depth = 0; t != null && depth < 5; depth++) {
+            if (t instanceof Error) return true;
+            for (Throwable s : t.getSuppressed()) {
+                if (s instanceof Error) return true;
+            }
+            if (t.getCause() == t) break;
+            t = t.getCause();
+        }
+        return false;
     }
 
     @Override
--- a/wkbase/src/main/java/com/chat/base/net/FastJsonResponseBodyConverter.kt
+++ b/wkbase/src/main/java/com/chat/base/net/FastJsonResponseBodyConverter.kt
@@ -49,15 +49,21 @@ class FastJsonResponseBodyConverter<T>(
     override fun convert(value: ResponseBody): T? {
         return value.use { body ->
             val reader = body.charStream()
-            val head = CharArray(STREAM_THRESHOLD_CHARS)
+            // 探测缓冲按需翻倍，不要一上来就按阈值分配：那样每个响应（99% 是几 KB 的小接口）
+            // 都要白拿 512KB，超过 TLAB、每次走慢路径分配，在一个「降内存峰值」的改动里
+            // 反而制造出固定的 per-request 垃圾。
+            var head = CharArray(INITIAL_PROBE_CHARS)
             var headLen = 0
-            while (headLen < head.size) {
+            while (headLen < STREAM_THRESHOLD_CHARS) {
+                if (headLen == head.size) {
+                    head = head.copyOf(minOf(head.size * 2, STREAM_THRESHOLD_CHARS))
+                }
                 val read = reader.read(head, headLen, head.size - headLen)
                 if (read == -1) break
                 headLen += read
             }
 
-            if (headLen < head.size) {
+            if (headLen < STREAM_THRESHOLD_CHARS) {
                 // 小响应：整包已在手上，保持原路径
                 if (headLen == 0) null else JSON.parseObject(String(head, 0, headLen), type)
             } else {
@@ -67,7 +73,7 @@ class FastJsonResponseBodyConverter<T>(
                     JSONReader(pushback).use { it.readObject<T>(type) }
                 }
                 WKHeapProbe.span(
-                    "http JSONReader(stream)", ">${head.size}chars", before, WKHeapProbe.usedNow()
+                    "http JSONReader(stream)", ">${headLen}chars", before, WKHeapProbe.usedNow()
                 )
                 result
             }
@@ -77,5 +83,12 @@ class FastJsonResponseBodyConverter<T>(
     private companion object {
         /** 超过这个长度才值得流式。256K 字符 ≈ 512KB，远小于会话同步那种 350 万字符的响应。 */
         const val STREAM_THRESHOLD_CHARS = 256 * 1024
+
+        /**
+         * 探测缓冲的起始容量，按需翻倍到 [STREAM_THRESHOLD_CHARS] 为止。
+         * 16K 字符（32KB）覆盖绝大多数接口，一次扩容都不用；越过阈值的大包最多 4 次
+         * arraycopy（总拷贝量 < 2× 阈值），相对后面的解析可以忽略。
+         */
+        const val INITIAL_PROBE_CHARS = 16 * 1024
     }
 }
--- a/wkbase/src/main/java/com/chat/base/net/FastJsonResponseBodyConverter.kt
+++ b/wkbase/src/main/java/com/chat/base/net/FastJsonResponseBodyConverter.kt
@@ -15,18 +15,67 @@
  */
 
 package com.chat.base.net
+
 import com.alibaba.fastjson.JSON
+import com.alibaba.fastjson.JSONReader
+import com.xinbida.wukongim.utils.WKHeapProbe
 import okhttp3.ResponseBody
 import retrofit2.Converter
+import java.io.PushbackReader
 import java.lang.reflect.Type
 
+/**
+ * 大响应体走流式解析，避免整包物化。
+ *
+ * 改动前是 `JSON.parseObject(value.string(), type)`：`ResponseBody.string()` 内部先
+ * `Buffer.writeAll(source)` 把整个响应解压进 okio Segment 链，再整体转成一个 UTF-16 String，
+ * 然后才建对象树——同一时刻内存里有三份。线上 OOM 有一条就死在第一步
+ * （`okio.Segment.<init>`，栈退时释放 231MB），FastJson 还没跑到。
+ *
+ * 现在按大小分流：
+ * - 响应 ≤ [STREAM_THRESHOLD_CHARS]：一次读完，走原来的 `JSON.parseObject(String, Type)`，
+ *   行为与改动前完全一致（绝大多数接口走这条，不引入回归面）。
+ * - 超过阈值：把已读的头部推回去，交给 [JSONReader] 增量解析。okio 不再缓冲整包，
+ *   也不再产生整包 String，峰值只剩对象树本身。
+ *
+ * 注意 fastjson 1.2.83 的 `JSON.parseObject(InputStream, ...)` **不是流式**
+ * （`JSON.java:569` 起先 `allocateBytes(64KB)` 再按 1.5 倍扩容 + arraycopy 读完整包），
+ * 用它没有意义，必须走 `JSONReader(Reader)`。
+ */
 class FastJsonResponseBodyConverter<T>(
     private val type: Type
 ) : Converter<ResponseBody, T> {
 
     override fun convert(value: ResponseBody): T? {
-        return value.use {
-            JSON.parseObject(value.string(), type)
+        return value.use { body ->
+            val reader = body.charStream()
+            val head = CharArray(STREAM_THRESHOLD_CHARS)
+            var headLen = 0
+            while (headLen < head.size) {
+                val read = reader.read(head, headLen, head.size - headLen)
+                if (read == -1) break
+                headLen += read
+            }
+
+            if (headLen < head.size) {
+                // 小响应：整包已在手上，保持原路径
+                if (headLen == 0) null else JSON.parseObject(String(head, 0, headLen), type)
+            } else {
+                val before = WKHeapProbe.usedNow()
+                val result = PushbackReader(reader, headLen).use { pushback ->
+                    pushback.unread(head, 0, headLen)
+                    JSONReader(pushback).use { it.readObject<T>(type) }
+                }
+                WKHeapProbe.span(
+                    "http JSONReader(stream)", ">${head.size}chars", before, WKHeapProbe.usedNow()
+                )
+                result
+            }
         }
+    }
+
+    private companion object {
+        /** 超过这个长度才值得流式。256K 字符 ≈ 512KB，远小于会话同步那种 350 万字符的响应。 */
+        const val STREAM_THRESHOLD_CHARS = 256 * 1024
     }
 }
--- a/wkbase/src/main/java/com/chat/base/net/FastJsonResponseBodyConverter.kt
+++ b/wkbase/src/main/java/com/chat/base/net/FastJsonResponseBodyConverter.kt
@@ -70,6 +70,9 @@ class FastJsonResponseBodyConverter<T>(
                 val before = WKHeapProbe.usedNow()
                 val result = PushbackReader(reader, headLen).use { pushback ->
                     pushback.unread(head, 0, headLen)
+                    // unread 已把探测缓冲拷进 PushbackReader 自己的 buf，这里必须放掉本地引用：
+                    // 否则两份 512KB 在整个 readObject 期间同时强可达，正好抵掉这条路径省下的量。
+                    head = CharArray(0)
                     JSONReader(pushback).use { it.readObject<T>(type) }
                 }
                 WKHeapProbe.span(
--- a/wkbase/src/main/java/com/chat/base/net/LogInterceptor.java
+++ b/wkbase/src/main/java/com/chat/base/net/LogInterceptor.java
@@ -95,9 +95,12 @@ public class LogInterceptor implements Interceptor {
 
             if (requestBody != null && !requestBody.isOneShot()) {
                 long reqLen = requestBody.contentLength();
-                if (reqLen > MAX_LOG_BODY_BYTES) {
-                    // 文件上传这类：writeTo 会把整包灌进内存，不能为了打一条日志付这个代价
-                    requestParams = "body 过大，已跳过打印（" + reqLen + " bytes）";
+                if (reqLen < 0 || reqLen > MAX_LOG_BODY_BYTES) {
+                    // 文件上传这类：writeTo 会把整包灌进内存，不能为了打一条日志付这个代价。
+                    // 长度未知（-1，流式 body）同样跳过：写完才知道多大，等于没有上界。
+                    requestParams = reqLen < 0
+                            ? "body 长度未知，已跳过打印"
+                            : "body 过大，已跳过打印（" + reqLen + " bytes）";
                 } else if (freeHeapBytes() < LOW_HEAP_SKIP_BYTES) {
                     requestParams = "堆吃紧，跳过 body 打印";
                 } else {
--- a/wkbase/src/main/java/com/chat/base/net/LogInterceptor.java
+++ b/wkbase/src/main/java/com/chat/base/net/LogInterceptor.java
@@ -37,6 +37,7 @@ import okhttp3.MediaType;
 import okhttp3.Request;
 import okhttp3.RequestBody;
 import okhttp3.Response;
+import okhttp3.ResponseBody;
 import okio.Buffer;
 import okio.BufferedSource;
 import okio.ByteString;
@@ -47,6 +48,40 @@ import okio.ByteString;
  */
 public class LogInterceptor implements Interceptor {
 
+    /**
+     * 只打印小于这个字节数的 body，超过的只打一行大小提示。
+     *
+     * <p>为什么要有上界：会话同步这类响应有 350 万字符，原实现 {@code body().string()} 整包物化 →
+     * {@code formatJson} 再建一遍 JSON DOM 并缩进美化 → StringBuilder 再拼一份 → 最后用整包
+     * String 重建 body 回写，同一份数据五六份，debug 包凭空多吃 80MB+，堆水位测量完全失真。
+     *
+     * <p>而且 {@link WKLogUtils} 按 3900 字符分片输出，350 万字符 = 900+ 条独立 logcat 记录：
+     * 人根本拼不回来，还会把 logcat 环形缓冲区冲爆，ANR 堆栈 / 堆探针那些真正要看的日志全被挤掉。
+     * 也就是说大 body 这样打等于没打，代价却是实打实的——所以直接不打，只留大小和 URL。
+     *
+     * <p>小 body（99% 的接口）行为与改动前一致，该看的都还在。
+     */
+    private static final long MAX_LOG_BODY_BYTES = 256L * 1024L;
+
+    /**
+     * 可用堆低于这个数就整段跳过 body 打印。
+     *
+     * <p>日志工具绝不能是压死进程的最后一根稻草。2026-08-14 压舱实测：堆只剩 9MB 时，
+     * 下面那句 {@code peekBody(256KB+1)} 自己 OOM 了，崩溃栈停在
+     * {@code LogInterceptor.intercept → PeekSource.read → InflaterSource → okio.Segment.<init>}
+     * ——被测代码没问题，是量具把进程压垮了。
+     *
+     * <p>注意 peek 的开销不止 256KB：gzip 响应要边解压边填 okio Segment，
+     * 真实占用是解压后的字节数 + 段链开销。
+     */
+    private static final long LOW_HEAP_SKIP_BYTES = 32L * 1024L * 1024L;
+
+    /** 当前可用堆（max - used）。 */
+    private static long freeHeapBytes() {
+        Runtime rt = Runtime.getRuntime();
+        return rt.maxMemory() - (rt.totalMemory() - rt.freeMemory());
+    }
+
     @NotNull
     @Override
     public Response intercept(@NotNull Chain chain) throws IOException {
@@ -54,27 +89,31 @@ public class LogInterceptor implements Interceptor {
         if (WKBinder.isDebug) {
             Request request = chain.request();
             //获取request内容
-            RequestBody requestBody = request.body();
+            // 已格式化、可直接打印的请求体；body 过大时是一行提示
             String requestParams = "";
+            RequestBody requestBody = request.body();
 
             if (requestBody != null && !requestBody.isOneShot()) {
-                boolean oneShort = requestBody.isOneShot();
-                WKLogUtils.e("判断执行次数"+oneShort);
-                MediaType type = requestBody.contentType();
-                Buffer source = new Buffer();
-                requestBody.writeTo(source);
-                try {
-                    Charset charset = type == null ? Charset.defaultCharset() : type.charset(Charset.defaultCharset());
-                    readBomAsCharset(source, charset);
-//                    Util.readBomAsCharset(source, charset);
-//                    Util.readBomAsCharset()
-                    requestParams = source.readString(charset);
-                } catch (Exception e) {
-                    e.printStackTrace();
-                } finally {
-                    closeQuietly(source);
+                long reqLen = requestBody.contentLength();
+                if (reqLen > MAX_LOG_BODY_BYTES) {
+                    // 文件上传这类：writeTo 会把整包灌进内存，不能为了打一条日志付这个代价
+                    requestParams = "body 过大，已跳过打印（" + reqLen + " bytes）";
+                } else if (freeHeapBytes() < LOW_HEAP_SKIP_BYTES) {
+                    requestParams = "堆吃紧，跳过 body 打印";
+                } else {
+                    MediaType type = requestBody.contentType();
+                    Buffer source = new Buffer();
+                    requestBody.writeTo(source);
+                    try {
+                        Charset charset = type == null ? Charset.defaultCharset() : type.charset(Charset.defaultCharset());
+                        readBomAsCharset(source, charset);
+                        requestParams = formatJson(source.readString(charset));
+                    } catch (Exception e) {
+                        e.printStackTrace();
+                    } finally {
+                        closeQuietly(source);
+                    }
                 }
-                // request = request.newBuilder().post(requestBody).build();
             }
             // 请求日志
             StringBuilder reqSb = new StringBuilder();
@@ -86,7 +125,7 @@ public class LogInterceptor implements Interceptor {
                     .append(request.headers());
             if (!TextUtils.isEmpty(requestParams)) {
                 reqSb.append("==============Body==============\n")
-                        .append(formatJson(requestParams)).append("\n");
+                        .append(requestParams).append("\n");
             }
             reqSb.append("**************Request***************").append("\n");
             WKLogUtils.d("wkHttpLog", reqSb.toString());
@@ -99,26 +138,38 @@ public class LogInterceptor implements Interceptor {
 
             if (!TextUtils.isEmpty(requestParams)) {
                 sb.append(" \n").append("============Request Body============\n")
-                        .append(formatJson(requestParams)).append("\n");
+                        .append(requestParams).append("\n");
             }
             sb.append("*************Response**************\n")
                     .append(response.request().url()).append("\n")
                     .append(String.format(Locale.getDefault(), "本次请求响应时间: %.1f ms", (t2 - t1) / 1e6d)).append("\n")
                     .append("=============Headers=============\n")
                     .append(response.headers()).append("\n");
-            String content = response.body().string();
             sb.append("==============Body==============\n");
-            if (response.isSuccessful()) {
-                sb.append(formatJson(content)).append("\n");
-            } else {
+            if (!response.isSuccessful()) {
                 sb.append("http请求失败，").append(response.networkResponse()).append("\n");
+            } else if (freeHeapBytes() < LOW_HEAP_SKIP_BYTES) {
+                // 堆已经吃紧，peek 本身就可能 OOM（实测过）。这条日志的价值远不如进程活着。
+                sb.append("堆吃紧（剩 ").append(freeHeapBytes() / (1024 * 1024))
+                        .append("MB），跳过 body 打印\n");
+            } else {
+                // peekBody 只缓冲最多 MAX+1 字节，且不消费原 body：小 body 拿到的就是全文，
+                // 大 body 拿到 MAX+1 字节即可判定超限。原实现的 body().string() +
+                // ResponseBody.create(content) 回写整段删掉了——那一步会把流式解析
+                // (FastJsonResponseBodyConverter) 在 debug 包里彻底抵消，因为转换器拿到的
+                // body 已经是内存里的整包 String，白流式了。
+                ResponseBody peeked = response.peekBody(MAX_LOG_BODY_BYTES + 1);
+                if (peeked.contentLength() > MAX_LOG_BODY_BYTES) {
+                    sb.append("body 过大，已跳过打印（>")
+                            .append(MAX_LOG_BODY_BYTES / 1024).append("KB）\n");
+                } else {
+                    sb.append(formatJson(peeked.string())).append("\n");
+                }
             }
             sb.append("*************Response**************");
             WKLogUtils.d("wkHttpLog", "   " + sb);
-            MediaType mediaType = response.body().contentType();
-            return response.newBuilder()
-                    .body(okhttp3.ResponseBody.create(content, mediaType))
-                    .build();
+            // 原 body 未被消费，原样返回，不再重建
+            return response;
         } else {
             return chain.proceed(chain.request());
         }
--- a/wkbase/src/main/java/com/chat/base/net/OkHttpUtils.java
+++ b/wkbase/src/main/java/com/chat/base/net/OkHttpUtils.java
@@ -86,6 +86,10 @@ public class OkHttpUtils {
                             .addInterceptor(mRewriteCacheControlInterceptor)
                             .addInterceptor(new CommonRequestParamInterceptor())
                             .addNetworkInterceptor(mRewriteCacheControlInterceptor)
+                            // LogInterceptor 只在 debug 生效（内部 WKBinder.isDebug 判定，release 走
+                            // chain.proceed 直通）。它对 body 有 256KB 上界：小 body 照旧全打，大 body
+                            // 只打大小——原实现会把整个响应体物化五六份，debug 包凭空多吃 80MB+，
+                            // 既带偏堆测量，又把 logcat 缓冲区冲爆。详见 LogInterceptor.MAX_LOG_BODY_BYTES。
                             .addInterceptor(new LogInterceptor())
                             // Emoji manifest 端点公开无鉴权，parse 前先做 raw body 字节上限
                             // 检查，防 hostile / MITM 推巨大 body → FastJson buffer 后 OOM。
@@ -93,9 +97,9 @@ public class OkHttpUtils {
                             //
                             // ⚠️ 必须放在 LogInterceptor 之后（即 list 末尾，最内层）：
                             // application interceptor 按 list order 由外到内包装，list 末尾的
-                            // interceptor 最先看到 response。LogInterceptor 会 body().string()
-                            // 全量读 body 后再打日志，若 BodyLimit 在 Log 之前（更外层），
-                            // Log 已经 OOM 了 BodyLimit 才拿到 response——检查太晚。
+                            // interceptor 最先看到 response，能在任何人读 body 之前先做大小检查。
+                            // （LogInterceptor 现在只 peek 最多 256KB，不再整包读，但这个顺序仍是
+                            // 正确的：让上限检查处在最内层，谁都别想先碰到超大 body。）
                             // 参考 PR #94 Jerry-Xin round-3 review B2。
                             .addInterceptor(new EmojiManifestBodyLimitInterceptor())
                             // 全局搜索 L1/L2 端点也做 body 大小上限（同 Emoji 拦截器语义），
--- a/wkbase/src/main/java/com/chat/base/utils/ANRWatchdog.java
+++ b/wkbase/src/main/java/com/chat/base/utils/ANRWatchdog.java
@@ -4,12 +4,15 @@ import android.content.Context;
 import android.os.Build;
 import android.os.Handler;
 import android.os.Looper;
+import android.os.SystemClock;
 import android.util.Log;
 
 import com.tencent.bugly.crashreport.CrashReport;
 
 import java.text.SimpleDateFormat;
+import java.util.ArrayList;
 import java.util.Date;
+import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 
@@ -21,10 +24,20 @@ public final class ANRWatchdog extends Thread {
 
     private static final String TAG = "ANRWatchdog";
     private static final int CHECK_INTERVAL_MS = 5000;
+    /**
+     * 实际睡眠超过这个时长说明看门狗线程自己被冻结/挂起（cached-app freezer、Doze 深睡），
+     * 此时 responded=false 不代表主线程卡住，本轮判定作废。
+     */
+    private static final long FREEZE_THRESHOLD_MS = CHECK_INTERVAL_MS * 2L;
+    /** 从日志文本还原堆栈时最多取多少帧，够 Bugly 分组即可。 */
+    private static final int MAX_PARSED_FRAMES = 32;
 
     private final Handler mainHandler = new Handler(Looper.getMainLooper());
     private volatile boolean responded = true;
     private volatile boolean stopped = false;
+    /** 因冻结跳过的轮数，随下一次真实 ANR 一起上报，不静默丢弃。 */
+    private volatile int freezeSkipCount = 0;
+    private volatile String lastFreezeSkipDetail = null;
     private static volatile ANRWatchdog instance;
     private static Context appContext;
 
@@ -65,21 +78,40 @@ public final class ANRWatchdog extends Thread {
             responded = false;
             mainHandler.post(() -> responded = true);
 
+            long startRealtime = SystemClock.elapsedRealtime();
+            long startUptime = SystemClock.uptimeMillis();
             try {
                 Thread.sleep(CHECK_INTERVAL_MS);
             } catch (InterruptedException e) {
                 break;
             }
+            long elapsedRealtime = SystemClock.elapsedRealtime() - startRealtime;
+            long elapsedUptime = SystemClock.uptimeMillis() - startUptime;
 
             if (!responded && !stopped) {
-                String fullReport = buildReport();
-                String briefReport = buildBriefReport();
+                if (elapsedRealtime > FREEZE_THRESHOLD_MS) {
+                    freezeSkipCount++;
+                    lastFreezeSkipDetail = "realtime=" + elapsedRealtime + "ms uptime=" + elapsedUptime
+                            + "ms (expected " + CHECK_INTERVAL_MS + "ms)";
+                    Log.w(TAG, "Skip round #" + freezeSkipCount + ", watchdog itself was frozen: "
+                            + lastFreezeSkipDetail);
+                    continue;
+                }
+
+                StackTraceElement[] mainStack = Looper.getMainLooper().getThread().getStackTrace();
+                String fullReport = buildReport(mainStack, freezeSkipCount, lastFreezeSkipDetail);
+                String briefReport = buildBriefReport(mainStack, freezeSkipCount, lastFreezeSkipDetail);
                 Log.e(TAG, briefReport);
 
                 DiagnosticLogFile.append(appContext, fullReport);
 
                 try {
-                    CrashReport.postCatchedException(new ANRError(briefReport));
+                    ANRError error = new ANRError(briefReport);
+                    // 让 Bugly 按主线程真实堆栈聚合，否则所有 ANR 都折叠到 ANRWatchdog.run 一个 issue 里
+                    if (mainStack.length > 0) {
+                        error.setStackTrace(mainStack);
+                    }
+                    CrashReport.postCatchedException(error);
                     Log.w(TAG, "ANR report posted to Bugly successfully");
                 } catch (Throwable e) {
                     Log.e(TAG, "Failed to post to Bugly: " + e.getMessage());
@@ -94,7 +126,8 @@ public final class ANRWatchdog extends Thread {
         }
     }
 
-    private static String buildBriefReport() {
+    private static String buildBriefReport(StackTraceElement[] mainStack,
+                                           int freezeSkipCount, String lastFreezeSkipDetail) {
         SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss.SSS", Locale.US);
         StringBuilder sb = new StringBuilder(2048);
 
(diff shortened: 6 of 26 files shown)


Jerry-Xin (dismissed):
Summary: In scope for `Mininglamp-OSS/octo-android`. This ANR/OOM batch addresses real Android DB/sync hot paths (SQLCipher WAL connection pool, missing composite/partial indexes, reminder query de-amplification, member negative-cache, streaming JSON) without any substantiated blocking correctness issue. Verified independently against head `cf5571e`.

💬 Non-blocking

🟡 `LogInterceptor.java:97` still buffers request bodies when `contentLength()` returns `-1`. In that case the `> MAX_LOG_BODY_BYTES` guard does not fire, and `requestBody.writeTo(source)` (~line 106) can still materialize a large unknown-length body in debug builds. Consider skipping unknown-length bodies or using a bounded sink. Debug-only, so low severity.

🔵 `ChannelMembersDbManager.java` — `invalidateChannel()`/`deleteWithChannel()` remove entries from `memberCache` but leave the same keys in `keysByMemberUid`. Since `deleteWithChannel()` can run repeatedly, the reverse index can retain dead keys and make later per-uid invalidations do (harmless) extra removes. Bounded by `MAX_CACHE_SIZE` full-clear, but cleaning matching reverse-set keys on channel delete would keep metadata tidy.

🔵 Verification caveat: no JVM in review env, so `FastJsonStreamConverterTest` was not executed locally — relying on code inspection.

✅ Highlights

- WAL open-flag fix (`WKDBHelper.openDatabase` with `ENABLE_WRITE_AHEAD_LOGGING` + `setWALConnectionPoolSize(8)`) correctly targets the pool-pinned-at-1 root cause; post-`open` `PRAGMA journal_mode=WAL` cannot resize the pool, so doing it via open flags is right. Fallback to rollback-journal on ROMs that reject WAL is non-blocking. All new debug samplers/self-tests are `BuildConfig.DEBUG`-gated and default-off.
- Index migrations match the actual query shapes: `idx_message_channel_order_seq`/`idx_message_channel_msg_seq (channel_id, channel_type, order_seq|message_seq)` cover the `max()` hot queries and synckey subquery; `idx_reminders_undone(type, channel_id) WHERE done=0` matches the new `WHERE done=0 AND type=?`; `idx_message_flame(is_deleted) WHERE flame=1` is a partial index. All are `CREATE INDEX IF NOT EXISTS` (no schema/data change, no unique constraint risk).
- Reminder query refactor (`hasUndoneReminderWithChannelPrefix` → `queryUndoneChannelIds` + in-memory prefix match) preserves the original `channel_id LIKE 'groupNo____%'` semantics: `hasUndoneThreadMention` requires `length >= parentLen+4` and case-insensitive `regionMatches`, matching SQLite's default case-insensitive LIKE with 4 mandatory chars. The single caller in `ChatFragment` was updated correctly; the double-loop-per-group full scan is collapsed to one query.
- Member negative-cache is well-designed: generation-counter guards the read-then-backfill race, double-check inside the synchronized slow path, negative sentinel `NULL_MEMBER` caches misses, reverse `keysByMemberUid` index enables precise per-uid invalidation. Fail-safe direction (drop-on-generation-change).
- Sync batching + `recents` release in `ConversationManager`, streaming `JSONReader` path for >256K payloads in `FastJsonResponseBodyConverter`, comparator key snapshots (removes TimSort contract violation + in-comparator DB reads), and batched thread-conversation queries all remove credible main-thread amplification without changing callback semantics.

yujiawei (changes requested):
# Code Review — PR #123 (octo-android)

Reviewed at head `cf5571e5ab60bff8ae9e84acf2ef51c1bf32d008`, diffed against `origin/main` (merge-base `4922864`). 26 files, +2053/−338.

I verified the load-bearing library claims against the actual `net.zetetic:sqlcipher-android:4.9.0` bytecode rather than taking the comments on trust, and traced every call site for the queries whose column sets or semantics changed.

---

## 1. Specification / scope compliance

**Spec: ✅**

The PR body's root-cause table lists 8 fixes. All 8 are present, and each one's mechanism holds up:

| Claim | Verification |
|---|---|
| Pool pinned at 1 because `openOrCreateDatabase` can't pass `ENABLE_WRITE_AHEAD_LOGGING` | Confirmed in bytecode: `SQLiteConnectionPool.setMaxConnectionPoolSizeLocked` sets `mMaxConnectionPoolSize = SQLiteGlobal.getWALConnectionPoolSize()` only when `openFlags & 0x20000000` is set, else hard-codes `1`. The old `PRAGMA journal_mode=WAL` genuinely could not change pool size. |
| Dropping the explicit `PRAGMA synchronous=NORMAL` is safe | Confirmed: `SQLiteGlobal.getWALSyncMode()` returns `"normal"`, applied per connection at open. No regression. |
| Writes still serialize with pool > 1 | Confirmed: `SQLiteDatabase.getThreadDefaultConnectionFlags(false)` returns the primary-connection-affinity flag, so write transactions all land on the primary connection. No new `SQLITE_BUSY` surface. |
| 3 migrations are index-only | Confirmed: all three files contain only `CREATE INDEX IF NOT EXISTS`; no schema or data mutation. |
| `idx_reminders_undone` / `idx_message_flame` partial indexes are usable | Both query predicates (`done=0 and type=?`, `flame=1 and is_deleted=0`) imply the index `WHERE` clauses with literal constants, so the optimizer can select them. |
| `queryLastMsgSeqs` predicate pushdown is result-equivalent | Confirmed: both pushed predicates (`channel_id<>''`, `is_deleted=0`) are `conversation`'s own columns, passed through unchanged by `select *`. Same surviving row set. |
| Batch conversation query is verbatim-equivalent to the single-row version | Diffed the SQL character by character (`ConversationDbManager.java:333` vs `:358`): same two `LEFT JOIN`s, same `is_deleted=0`, same `serializeMsg`. And `getUIConversationMsg` already returns `null` for a missing row (`ConversationManager.java:395-400`), so "absent from map" == "was null" — the `threadConv != null` fallbacks behave identically. |
| `queryWithFlame` column narrowing loses nothing | The only consumer chain is `MsgManager.getWithFlame` → `MsgModel.sweepFlameMsg` → `deleteMsg`/`deleteWithClientMsgNos`, which read exactly `clientMsgNO`, `messageID`, `channelID`, `channelType`, `messageSeq`, `flame`, `viewed`, `viewedAt`, `flameSecond` — all nine are in the new projection. |
| `syncChat.conversations[i].recents = null` is safe | Confirmed: `.recents` has exactly one production reader (`ConversationManager.java:632-633`); no `onBack` implementation touches it. |
| `hasUndoneThreadMention` matches the old `LIKE 'X____%'` | Equivalent: `____%` requires ≥ 4 trailing chars and SQLite's ASCII-case-insensitive `LIKE`, matched by the `length() >= parentLen + 4` + `regionMatches(true, …)` pair. |
| Member-cache invalidation surface is complete | Confirmed independently: `ChannelMembersDbManager` is the only writer of `channel_members` repo-wide, and `ChannelDBManager` has exactly 4 writers of `channel` (`:158`, `:189`, `:207`, `:394`) — all four now invalidate for `PERSONAL`. Account switch is covered via `ChannelManager.clearAllChannelCache`. |

Nothing from the stated scope is missing.

**Undeclared changes** — none add functionality, but they are not in the PR description and should be called out so a reader isn't surprised:

- `ChatFragment.java`: `hasMentionInCategory(...)` deleted. Verified dead (zero callers), so this is pure dead-code removal.
- `UserModel.java:267`: besides moving the block onto a background pool, the diff **deletes a second pass** over `tempList` that re-applied the same `online`/`lastOffline` values and re-added the same object instances to `list`. By my read it was fully redundant (`saveOrUpdateChannels` on duplicates is idempotent), so this is behavior-preserving — but please confirm that was intentional rather than a merge artifact.
- `ChooseChatActivity.sortByTopAndTime`: beyond snapshotting the sort keys, null `uiConveursationMsg` entries now sort deterministically last instead of comparing "equal" to everything. That's a correct fix for a non-transitive comparator, but it does change output ordering for those rows.
- `.gitignore`: `/buglybin/`.
- `ChatFragment.filterAndDisplay`: doc-comment rewrite only.

---

## 2. Code quality

**Quality: Changes-Requested**

### P1 — a single exception permanently disables burn-after-reading cleanup for the rest of the process

`wkuikit/src/main/java/com/chat/uikit/message/MsgModel.java:139-150`

```java
private void scheduleFlameSweep(long delayMs) {
    if (!flameLoopRunning.get()) return;
    flameLoopTask = WKDbScheduler.get().scheduleDirect(() -> {
        if (!flameLoopRunning.get()) return;
        if (sweepFlameMsg()) {
            scheduleFlameSweep(FLAME_INTERVAL_MS);
        } else {
            flameLoopRunning.set(false);
        }
    }, delayMs, TimeUnit.MILLISECONDS);
}
```

`flameLoopRunning` is the CAS gate in `startCheckFlameMsgTimer` (`:135`), and the only starter is `ChatActivity:4014`. There is no `try`/`finally` around `sweepFlameMsg()`. If it throws:

- `WKDbScheduler` is `Schedulers.from(newSingleThreadExecutor)`, so the throwable goes to `RxJavaPlugins.onError` and the worker survives — but the recursive `scheduleFlameSweep` never runs;
- `flameLoopRunning` is left `true`, so **every subsequent `startCheckFlameMsgTimer()` short-circuits on the CAS** for the remainder of the process;
- expired flame messages are then never deleted until the user logs out (`stopTimer` at `WKUIKitApplication:927`) or restarts the app. Silently.

This is a new failure mode, not an inherited one. The old code's `TimerTask` body was just `deleteFlameMsg()`, which pushed all the throwing work *inside* a separate `WKDbScheduler.scheduleDirect` runnable — so an exception there never killed the `Timer`, and the 1 s tick kept coming. The new design puts the throwing work on the loop's own critical path.

**Failure scenario:** during a heavy conversation sync (exactly the pressure this PR targets), `sweepFlameMsg` → `deleteMsg` (builds a `JSONArray`) or `deleteWithClientMsgNos` (SQLite write) throws `OutOfMemoryError` or `SQLiteException`. From that moment the app never deletes another expired burn-after-reading message, and nothing surfaces.

**Fix:**

```java
flameLoopTask = WKDbScheduler.get().scheduleDirect(() -> {
    if (!flameLoopRunning.get()) return;
    boolean keepGoing = false;
    try {
        keepGoing = sweepFlameMsg();
    } catch (Throwable t) {
        keepGoing = true;   // transient failure — retry next tick rather than dying
    } finally {
        if (keepGoing) scheduleFlameSweep(FLAME_INTERVAL_MS);
        else flameLoopRunning.set(false);
    }
}, delayMs, TimeUnit.MILLISECONDS);
```

(If you prefer not to retry forever on a hard failure, add a consecutive-failure counter — but `flameLoopRunning` must never be left `true` with no scheduled successor.)

### P2 — `putCache`'s generation check is not atomic with the `put`, so a stale entry can be pinned indefinitely

`wkim/src/main/java/com/xinbida/wukongim/db/ChannelMembersDbManager.java:93-104`

```java
if (cacheGeneration.get() != generation) return;
...
memberCache.put(key, member == null ? NULL_MEMBER : member);
```

The Javadoc at `:63-77` says the generation counter "堵住" (plugs) the read-through race. It narrows it; it does not close it. `querySlowPath`, `insert`, `update`, `updateWithField` are mutually `synchronized`, but `insertMembers` (`:381`, `:406`) and `deleteWithChannel` (`:500`) are not — exactly as the comment acknowledges. So an `invalidate()` from one of those can land in the window between the `get()` and the `put()`, and the stale value is written after the removal.

**Failure scenario:** a group-member sync (`insertMembers`) renames a user at the instant a bind-path read is backfilling. The old nickname/avatar stays in `memberCache` with no TTL — it will not self-heal until the next write to that member or an account switch.

**Fix:** re-check after the write and roll back — cheap and closes the window:

```java
memberCache.put(key, member == null ? NULL_MEMBER : member);
if (cacheGeneration.get() != generation) { memberCache.remove(key); return; }
keysByMemberUid.computeIfAbsent(...)...
```

Please also soften the Javadoc claim to match whatever guarantee ends up being true.

### P2 — the cache hands out shared mutable instances, and two existing call sites mutate them in place

`ChannelMembersDbManager.java:296` (the fast path returns the cached instance) vs `wkuikit/.../chat/ChatActivity.java:1672-1675` and `:1707-1709`, which do:

```java
...getMemberOfFromIfCached().memberRemark = channel.channelRemark;
...getMemberOfFromIfCached().memberName   = channel.channelName;
```

`WKMsg.memberOfFrom` is populated from `ChannelMembersManager.getMember` → `ChannelMembersDbManager.query`, so these writes now land on the **process-wide shared** `WKChannelMember`, not on a per-message copy. Note `serializableChannelMember` (`:650`) fills `memberRemark` from `channel_members.member_remark` (the group-scoped remark) while `channel.channelRemark` is the personal-channel remark — so the mutation writes a semantically different value into the shared entry.

Today this happens to be harmless: every `setRefreshChannel` path that can reach the `PERSONAL` branch (`ChannelManager.saveOrUpdateChannel:168-173`, `saveOrUpdateChannels:325-331`, `updateName`/`updateMute`/`updateStatus` → `ChannelDBManager.updateWithField:386`) performs the DB write — and therefore `invalidateMemberUid` — *after* firing the listener, so the mutated instance is evicted immediately afterwards. But that ordering is incidental. The one existing in-memory-only `setRefreshChannel` (`ChatActivity:947`) is `GROUP`-typed and so misses the mutating branch by luck; the next one added won't necessarily.

The class Javadoc already states "调用方不得就地改字段", and the PR touches `ChatActivity.java` — so please either fix these two call sites to go through a proper update API, or have `query()` return a defensive copy on the fast path.

### P2 — `invalidateChannel` leaks entries in `keysByMemberUid`

`ChannelMembersDbManager.java:130-141` removes matching keys from `memberCache` but never prunes `keysByMemberUid`, unlike `invalidate(...)` at `:106-113`. Orphaned key sets accumulate and are only reclaimed by the `MAX_CACHE_SIZE` full wipe. Small, but trivially fixable in the same loop.

### P2 — the streaming path holds two 512 KB char buffers for the whole parse

`wkbase/src/main/java/com/chat/base/net/FastJsonResponseBodyConverter.kt:55-76`

In the stream branch `headLen == STREAM_THRESHOLD_CHARS` exactly, so `PushbackReader(reader, headLen)` allocates its own 256 K-char (512 KB) buffer and `unread` copies into it — while the local `head` array stays strongly reachable for the entire `readObject` call. That's ~1 MB pinned across the parse in a change whose purpose is cutting the peak. Dropping `head = CharArray(0)` right after `unread` reclaims half of it.

### P2 — the two parser paths are not the same parser

Same file. Below the threshold you get `JSON.parseObject(String, Type)`; above it, `JSONReader(Reader).readObject(type)`. These differ in more than buffering — notably `JSONReader.readObject` does not run `handleResolveTask`, so `$ref` / back-reference resolution behaves differently, and the two use different lexers (`JSONScanner` vs `JSONReaderScanner`). The split is size-keyed, which means the largest and most structurally complex payload in the product (conversation sync) is the *only* one that takes the newer path in production.

`FastJsonStreamConverterTest` is a good regression net for field-level equivalence on the conv-sync shape (including escapes, multi-byte chars, and ±3 chars around the threshold — nice). Two gaps worth closing:

- no malformed / truncated-JSON case, so nothing pins down that the stream path surfaces the same exception type Retrofit/`ResponseExceptionHandle` expects;
- no empty-body case (`headLen == 0` → `null`), which is the one place the new code returns `null` by an explicit branch rather than via fastjson.

### P2 — slow-query instrumentation misses the `while (cursor.moveToNext())` shape, and is unthrottled in release

`wkim/src/main/java/com/xinbida/wukongim/db/WKDBHelper.java:552-598`

`SlowQueryCursor` overrides `getCount`, `moveToFirst`, `moveToLast`, `moveToPosition` — but not `moveToNext`. Call sites that iterate with `while (cursor.moveToNext())` and never call `getCount()`/`moveToFirst()` (e.g. `ChannelDBManager.isExist`, `:117-140`, whose only cursor call is `moveToNext()` at `:135`) do the full `CursorWindow` fill completely unmeasured. Since the point of this wrapper is to catch the next `queryWithFlame`-class 89 s scan before it becomes an ANR, that blind spot matters.

Separately: `guardCursor` (`:518`) and `report` (`:592`) log via `WKLoggerUtils` in **release** — deliberate, per the comment — but with no rate limiting. `WKLoggerUtils` also writes to file (the PR notes `setDebug(true)` is hard-coded in this repo), so a sustained slow-DB episode now adds unbounded production disk I/O to an already-degraded device. A simple per-SQL dedupe or a per-minute cap would keep the signal without the amplification.

### P2 — batching trades transaction atomicity, and `guardLoadGateSize` can double-post a preview

- `ConversationManager.saveConversationBatch:738` — one transaction per 20 conversations instead of one for the whole sync. Partial success is documented and defensible for self-healing list data; just noting it's a real semantic change for anyone reasoning about sync atomicity.
- `MsgDbManager.guardLoadGateSize` clears **both** `requestCounts` and `previewPosted` when either exceeds 256, including entries for in-flight loads on other channels — which can re-open a preview gate and post a duplicate preview. The comment says `ChatActivity.applyDataToAdapter` tolerates that; worth a one-line assertion of that in a test, since it's the kind of thing that quietly stops being true.

### P2 — nits

- `ANRWatchdog.java:39,93`: `freezeSkipCount` / `lastFreezeSkipDetail` are never reset after being reported, so every later ANR report carries the cumulative count. Fine if intended — say so, otherwise reset after `appendFreezeSkips`.
- `LogInterceptor.java:97`: `requestBody.contentLength()` returns `-1` for unknown-length bodies, and `-1 > MAX_LOG_BODY_BYTES` is false, so those still fall through to `writeTo(source)` and materialize fully. Debug-only, but it's the exact case (streamed uploads) the guard is meant to catch — treat `< 0` as "too large".
- `ConversationManager.java:664`: nulling `syncChat.conversations[i].recents` mutates an object that outlives the method (it's handed to `iSyncConversationChatBack.onBack(syncChat)`). Correct today — I verified the single reader — but it's a landmine for the next person who wants `recents` in that callback. Consider taking a local reference and clearing the list instead, or at minimum documenting the invariant at the `onBack` boundary.
- `ConversationDbManager.queryLastMsgSeqs:279`: the pushdown changes the query plan, and SQLite's `GROUP_CONCAT` ordering is unspecified — so `synckey` element order may differ from before. Almost certainly irrelevant (servers parse by splitting on `|`), but confirm the backend is order-insensitive.

### What I liked

The measurement discipline here is unusually good, and several of these are real bugs found the hard way rather than by pattern-matching:

- `MsgManager.setSyncChannelMsgListener` (`:987-1004`) — the missing `onBack(null)` on the unregistered-listener branch is a genuine "paging never recovers" bug, and the fix is right: a callback API must answer on every branch. Verified `MsgDbManager:282` is the sole caller and handles `null` as terminal.
- `ChatActivity` `subscribe(onNext, onError)` — the single-arg `subscribe` leaving `isRefreshLoading`/`isMoreLoading` stuck at `true` on any `buildUiMsgList` throw is the same class of bug, and the error branch correctly releases the flags *before* the `isFinishing()` guard while keeping view work after it.
- `requestCount` → `requestCounts` keyed by `channelId_channelType_oldestOrderSeq` — correct, and I verified the recursion at `:283` reuses the same `oldestOrderSeq` so the key is stable across rounds.
- Snapshotting `top` into an `IdentityHashMap` before `sort` fixes a real `TimSort` contract violation rather than papering over it.
- `ANRWatchdog.setStackTrace(mainStack)` plus the freeze guard turns a single useless Bugly bucket into per-frame buckets — and capturing the stack *after* the sleep is the right moment.
- `BaseObserver.reportIfSwallowedError` deliberately reports without changing crash semantics, and caps at 3 per process so reporting an OOM can't finish the process off.

---

## 3. Overall verdict

**REQUEST_CHANGES**

Spec compliance passes cleanly. The architecture and the library-level reasoning are correct — I checked the SQLCipher claims against the AAR bytecode and they hold. What blocks is the one P1: `scheduleFlameSweep` has no exception guard, so a single throw from `sweepFlameMsg` leaves `flameLoopRunning` stuck at `true` and permanently disables burn-after-reading deletion for the process, silently. That's a ~5-line fix.

Everything else is P2 and can land in the same round or a follow-up, but the two cache items (`putCache` atomicity, shared-instance mutation from `ChatActivity`) are worth doing now while the context is fresh — they're the parts most likely to produce "wrong name on a message" reports that are impossible to reproduce later.

## 4. Suggested next steps

1. Wrap the `scheduleFlameSweep` runnable in `try`/`finally` so the loop can never be left armed-but-dead. **(blocking)**
2. Re-check `cacheGeneration` after `memberCache.put` and remove on mismatch; adjust the Javadoc to claim only what's true.
3. Either fix `ChatActivity:1672-1675` / `:1707-1709` to stop mutating cached members in place, or return a copy from `ChannelMembersDbManager.query`'s fast path.
4. Prune `keysByMemberUid` in `invalidateChannel`.
5. Release `head` after `PushbackReader.unread`, and add malformed-JSON + empty-body cases to `FastJsonStreamConverterTest`.
6. Override `moveToNext` in `SlowQueryCursor`, and rate-limit the release-mode `[slow-db]` log.
7. Confirm the `UserModel` second-pass deletion was intentional.

## 5. Beyond the diff — things to watch

- **Snapshot isolation is now observable.** With pool > 1, a main-thread read no longer blocks behind a background write transaction — it gets the pre-transaction WAL snapshot instead. That is the fix, but it means UI computed mid-sync (e.g. `ChatFragment.buildFollowDisplayList` → `queryUndoneChannelIds`, unread totals, `@` badges) can now render briefly stale values where it previously stalled. Worth an explicit pass during the "large-account sync" test to confirm every such surface has a post-commit refresh; `setOnRefreshMsg` covers the conversation list, but the reminder/badge paths are computed on the main thread independently.
- **`-wal` growth is no longer capped.** Dropping `wal_autocheckpoint=100` / `journal_size_limit=1MB` is correctly justified (there is no global setter for either in `SQLiteGlobal` — I checked, only `setWALConnectionPoolSize` exists — so with pool > 1 you can only configure one arbitrary connection). But the concern the removed lines were written for is real and now unmitigated at the library defaults (1000 pages / 10000 bytes). Please watch `-wal` file size and first-checkpoint latency on a large account after a heavy `saveSyncChat`.
- **Reminder matching is now O(items × undone-rows) on the main thread.** Strictly better than N full-table scans, and bounded in practice — but it scales with total undone `@` reminders rather than with the visible list. Worth a sanity check on an account with a large mention backlog.
- **Still main-thread DB.** `queryUndoneChannelIds` and the thread-conversation batch queries are 1 query instead of N, but they still run on the main thread. The debug-only `warnIfMainThread` guard is a good step; the follow-up is moving these off the main thread entirely.
- **PR-stated open item stands:** the "history paging won't load" root cause is still unconfirmed against the observed device hang. Four independently valid defects were fixed; if the symptom recurs after this ships, that's a signal none of the four was the actual cause, not that the fixes were wrong.

mzanalyticst:
感谢 review，几条都核过了。P1 认，6 条 P2 已修，剩下的据实说明。

## 已修（6 笔，`badb2bd..c0858b3`）

**P1 `scheduleFlameSweep`（`badb2bd`）** — 成立，而且确认是自排程引入的新失效模式：旧 `TimerTask` 只把活儿丢进另一个 `scheduleDirect`，异常杀不掉 Timer；改自排程后 `sweepFlameMsg()` 落在了循环自己的关键路径上。不变量已按你说的建立：runnable 退出时要么已排下一轮，要么 `flameLoopRunning` 已置回 false。

一处与建议不同：异常时没有无条件重试，而是连续失败 3 次收摊。理由是无条件重试会把一个必然失败的 DB 操作变成每秒一次的热循环（OOM 场景下尤其不想这样）；停掉之后下次进聊天页 `ChatActivity:4014` 的 CAS 能重新起来，所以仍可自愈。补充一个产品事实：阅后即焚在 Android 端实际没有在用，库里没有 flame 消息时第一轮 `sweepFlameMsg()` 就 `return false` 停掉，所以这条的真实影响接近零 —— 修它是为了守不变量，不是为了救数据。

**`putCache` 代际原子性（`d8f9478`）** — 成立，按建议改成 put 后复查、不一致就 remove。两种交错都安全：失效方的 remove 与这里的 remove 至少有一个执行在 put 之后。Javadoc 里"堵住"的说法已改成实话，并写清为什么必须查两次。

**共享实例被就地改（`d8f9478`）** — 成立，包括你指出的语义错一层（`channel.channelRemark` 是个人频道备注，`member_remark` 是群内备注）。选了你给的第二个方案：在缓存边界返回副本，而不是改 `ChatActivity` 那两处。理由是那个循环本来就是逐条消息改自己的 `memberOfFrom`，换副本后行为不变（每条消息有了自己的实例），而只改调用点堵不住"下一个人加的"。新增 `copyOf()`（`extraMap` 也复制一层），`query` 快路径、`querySlowPath` 的 double-check 命中、`putCache` 存入都走副本，缓存里那份从此不被任何外部引用别名。类 Javadoc 那句"返回共享实例，调用方不得就地改字段"也一并改掉了 —— 靠注释约束调用方本来就不是个约束。

**`invalidateChannel` 漏剪反向索引（`d8f9478`）** — 成立。补充一点：不只是多做无害 remove，`putCache` 的上限只看 `memberCache.size()`，拦不住这里留下的孤儿 key，频道反复删除时反向索引只增不减。空 `Set` 留着不删是有意的：删它有"删除瞬间另一线程刚 add"的竞态，代价是漏一次失效（脏名字），比留个空 Set 贵。

**两个 512KB 缓冲同时在世（`815c17b`）** — 成立，`unread` 之后置空 `head`，1MB 降到 512KB。

**`SlowQueryCursor` 漏 `moveToNext` + release 日志无节流（`212db3c`）** — 两条都成立。`moveToNext` 已覆盖（盲区正好落在最常见的迭代写法上，`ChannelDBManager.isExist` 就是例子）。节流按 SQL 每分钟一条、key 数上限 200 超了整清，与隔壁 `warnIfMainThread` 的按 SQL 去重对称起来。

**`LogInterceptor` 的 `-1`（`98467be`）** — 成立，`reqLen < 0` 并入跳过分支。Jerry-Xin 也提了同一条。

**测试（`c0858b3`）** — 按你说的补了空 body 和截断 JSON 两例，`FastJsonStreamConverterTest` 现在 7/7。

## 说明，不改代码

**两条 parser 路径不是同一个 parser** — 差异客观存在，但对上层无影响，已核实两点：(1) 截断 JSON 时两条路径抛的是同一类异常，新加的 `malformedPayloadThrowsSameTypeAsLegacy` 一次通过，现在有测试钉着；(2) `ResponseExceptionHandle.handleException` 只对 `HttpException` 按 code 分支，其余一律落到 `else` 包成 `ResponseThrowable(e, -1)` 取 `getMessage()`，**不存在按异常类型分派的逻辑**。至于 `$ref`，那是 fastjson 序列化侧的产物，服务端返回的 JSON 不带回引。

**`saveConversationBatch` 每 20 条一事务** — 确认是有意的语义变化，PR 描述里写了。会话列表是可自愈数据，用整体原子性换掉一个长事务对连接的独占，这个交换是这批改动的目的之一。

**`UserModel:267` 删掉的第二遍循环** — 有意，不是 merge 事故，你的判断对。已 diff 确认：第二个 block 对 `tempList` 再扫一遍写入**完全相同**的 `online` / `lastOffline`，再把同一批对象引用二次 `list.add`，而 `saveOrUpdateChannels` 对重复项幂等。就是去冗余。

**`queryLastMsgSeqs` 的 `GROUP_CONCAT` 顺序** — 客户端这边只能确认是按 `|` 拼接，SQLite 不保证顺序这点你说得对。已记下找 server 侧确认解析是否顺序无关，有结论会回到这条。

## 暂不处理，理由如下

**`ANRWatchdog` 的 `freezeSkipCount` 不重置** — 你的两个选项我选"说清楚"：它是**进程累计**的"watchdog 自己被冻住（设备睡眠 / doze）所以跳过判定"的次数，作为"这个进程经历过多少次冻结"的背景信息附在每份报告上。同意累计值容易被误读成与本次 ANR 相关，但现在改成按报告区间重置会让已经在 Bugly 里的历史数据和新数据不可比。留到下一批统一处理。

**`ConversationManager:664` 的 `recents = null`** — 同意是给下一个人埋的雷，但改动落在同步回调的热路径上，而这批已经动了不少同步路径，真机验证成本大于收益。下一批一起改成局部引用。

**`guardLoadGateSize` 双清的测试断言** — 认同该有，同上，下一批补。

## 你"Beyond the diff"那几条

都记下了，尤其两条：

- **快照隔离现在可观测** — pool > 1 之后主线程读不再堵在后台写事务后面，代价是可能读到事务前的 WAL 快照。`setOnRefreshMsg` 覆盖会话列表，但提醒/角标那几条是主线程独立算的，会在大账号同步测试里专门盯一遍这些界面有没有 post-commit 刷新。
- **`-wal` 不再有上界** — 确实是"正确的取舍 + 未缓解的风险"并存。会在大账号重同步后看 `-wal` 文件大小和首次 checkpoint 延迟，有数据回来补。

"history paging 根因未确认"这条也接受：四个缺陷各自独立成立，如果症状在这批上线后复现，那说明四个都不是那个根因，而不是修错了。

Jerry-Xin (changes requested):
Re-review at head `c0858b33a167` (previously APPROVED by me at `cf5571e5ab60`). This commit is a focused response to peer feedback — I diffed only the true delta `cf5571e5ab60..c0858b33a167` (6 files, no rebase drift) rather than the full PR.

## What changed since cf5571e5ab60 — prior feedback is well addressed

yujiawei's REQUEST_CHANGES items are all handled:

- **P1 (blocking): `scheduleFlameSweep` exception guard** — added try/catch so `flameLoopRunning` can no longer be left stuck-true; transient failures retry, and a failure streak (3) stops the loop cleanly (restarts on next `startCheckFlameMsgTimer`). The stall bug is fixed. ✅
- **putCache atomicity** — added a second `cacheGeneration` re-check *after* `memberCache.put`, closing the invalidate-between-check-and-put window; Javadoc updated to match. ✅
- **Shared-instance mutation** — `query` fast path and `querySlowPath` double-check now return `copyOf(...)`. Verified `copyOf` covers all 19 public fields of `WKChannelMember` (no superclass beyond `Parcelable`) and deep-copies `extraMap`. ✅
- **`keysByMemberUid` pruning** — `invalidateChannel` now prunes the reverse index in the same loop. ✅
- **Two 512KB buffers in stream path** — `head = CharArray(0)` right after `unread`. ✅
- **Converter tests** — added empty-body (`headLen==0 -> null`) and malformed/truncated cases asserting same exception type as the legacy path, small and large payloads. ✅
- **`SlowQueryCursor.moveToNext` blind spot + unthrottled release logging** — `moveToNext` now measured; `shouldLogSlow` per-SQL 60s throttle with a 200-key cap. ✅
- **`LogInterceptor` unknown-length body** — `reqLen < 0` now treated as "skip printing". ✅

## 🔴 Blocking

- **`MsgModel.scheduleFlameSweep` catches `Throwable` and swallows `Error` silently in release.** `wkuikit/src/main/java/com/chat/uikit/message/MsgModel.java:154` — the new `catch (Throwable t)` includes `OutOfMemoryError`, and in release builds the handler does nothing but `++flameFailureStreak` (the `Log.w` is `BuildConfig.DEBUG`-gated). This directly contradicts this batch's own stated theme — `BaseObserver.reportIfSwallowedError` and the `ANRWatchdog` work exist precisely to *surface* swallowed OOMs to Bugly — yet this new path drops them with zero release visibility. It also retries an operation that just `OutOfMemoryError`'d up to two more times. Suggested minimal fix: narrow the retry `catch` to `Exception` (recoverable transients), and for `Error` report via the same mechanism the rest of the batch uses (WKLoggerUtils/Bugly) before rethrowing or stopping — do not swallow it silently. The guard against the stuck-true stall is correct and should stay; only the `Error` handling needs to preserve reporting semantics.

## 💬 Non-blocking

- 🔵 A tiny fake-scheduler test around the flame state machine (empty result → stop; recoverable exception → retry; `Error` → surfaced/not swallowed) would pin the new state machine, complementing the good converter tests.
- 🟡 Unit tests could not be executed in the review checkout (no JRE); CI should confirm the new `FastJsonStreamConverterTest` cases compile and pass.

## ✅ Highlights

- The double-generation-check `putCache` reasoning is correct: at least one of the two `remove`s always runs after the `put`, so no interleaving leaves a stale entry in a TTL-less cache.
- Returning copies from the cache read paths is the right call given `ChatActivity` mutates member fields in place.
- `moveToNext` instrumentation closes the exact `while (cursor.moveToNext())` blind spot yujiawei flagged (e.g. `ChannelDBManager.isExist`), and the per-SQL throttle stops a struggling device from adding unbounded release disk I/O.
- The malformed-JSON test asserting the *same exception type* across the size-keyed parser split is exactly the right regression net for the two-parser risk.

Everything except the single `Error`-swallow item is resolved. Fix the release-mode `Error` visibility in `scheduleFlameSweep` and this is an approve.

mzanalyticst:
这条 blocking 成立，已修（`df4cdb9`）。而且核下来比 review 里描述的更严重一点。

**掐掉的是一条原本存在的通道，不只是少了条日志。** 本仓库没有设 `RxJavaPlugins.setErrorHandler`（全仓 grep 无命中），所以 `scheduleDirect` 里抛出的 `Error` 原本会走：`RxJavaPlugins.onError` → OOM 不属于 `isBug()` 那几类 → 包成 `UndeliverableException` → `printStackTrace()` + `uncaught()`，而 `uncaught()` 直接调线程的 UncaughtExceptionHandler，Bugly 装的就是全局默认 handler。也就是说改之前它是**能到 Bugly** 的，`badb2bd` 之后一点痕迹都不留。

你指出的自相矛盾也确实存在：`BaseObserver.reportIfSwallowedError` 的 Javadoc 自己写着"同一类事故能不能被发现全看它恰好抛在哪一帧，不可接受"和"**刻意不改崩溃语义**"，而 `badb2bd` 正好违反了本批次立的规矩。

**改法比建议的更简单：`catch (Exception)` + `finally`，不新增上报通道。**

```java
boolean keepGoing = false;
try {
    keepGoing = sweepFlameMsg();
    flameFailureStreak = 0;
} catch (Exception e) {
    keepGoing = ++flameFailureStreak < FLAME_MAX_FAILURE_STREAK;
    if (BuildConfig.DEBUG) Log.w(...);
} finally {
    if (keepGoing) scheduleFlameSweep(FLAME_INTERVAL_MS);
    else flameLoopRunning.set(false);
}
```

- 只接 `Exception`（可恢复的瞬时故障，如同步高峰的 `SQLiteException`），按 3 次上限重试；
- `Error` 原样穿出去，崩溃语义与 `cf5571e` 逐字节一致 —— 不接住就等于没吞，所以不需要额外 `postCatchedException`；顺带也不再重试一个刚 `OutOfMemoryError` 的操作；
- 不变量移进 `finally`：正常返回、`Exception`、`Error` 三条路径下 `flameLoopRunning` 都不会留在 `true`，`badb2bd` 要修的"清理永久禁用"仍然守住。

`:wkuikit` 编译通过。

## 你的两条非阻塞

**🟡 CI 跑测试** — 已有答案：`FastJsonStreamConverterTest` 本地 7/7，CI 的 `build` 在 `c0858b3` 上已绿。

**🔵 flame 状态机的 fake-scheduler 测试** — 认同该有，但这轮不做，理由是成本结构：有价值的断言是"`Error` 会穿出去"和"`finally` 一定复位标志"，两者都必须真的跑 runnable，而 `wkuikit` 的 test 源集只有 junit（无 Mockito / Robolectric），`sweepFlameMsg` 又依赖 `WKIM` / `WKConstants` 一串静态单例。要么引 `mockito-inline` 做 `mockStatic`，要么把"扫一轮"从 `MsgModel` 里反转成可注入接口 —— 对一个在 Android 端实际未使用的功能（库里没有 flame 消息时第一轮就 `return false` 停掉）来说都偏重。记在下一批：如果那时要动这块代码，就连带把 runnable 体抽成不依赖 Android 的包私有方法并补三分支测试。

Jerry-Xin (approved):
🔄 Re-review at `df4cdb9414dd` (was `c0858b33a167`). **APPROVED** — the single blocking item from my previous review is now correctly resolved.

**What changed since `c0858b33a167`**
Exactly one file changed (`ahead_by=1, behind_by=0`, no rebase drift): `wkuikit/src/main/java/com/chat/uikit/message/MsgModel.java`, `scheduleFlameSweep()` (+18 / −11). This is a targeted fix of the prior blocker, nothing else touched.

**Prior blocker — RESOLVED ✅**
My REQUEST_CHANGES at `c0858b33` flagged that `scheduleFlameSweep()` caught `Throwable`, swallowing `Error`/`OutOfMemoryError` (only a debug-gated `Log.w` + a counter in release), which contradicted the batch's Bugly-visibility theme and retried an op that had just OOM'd. The new commit fixes this properly:
- Retry catch narrowed `Throwable` → `Exception`, so only recoverable transient faults (e.g. sync-peak `SQLiteException`) are retried; the OOM'd op is no longer re-run.
- `Error`/`OutOfMemoryError` now propagates out of the RxJava runnable. Verified at head: the repo sets no `RxJavaPlugins.setErrorHandler`, so it travels the original `RxJavaPlugins.onError` → uncaught-handler → Bugly path — consistent with `BaseObserver.reportIfSwallowedError` (confirmed to exist in this batch at `wkbase/src/main/java/com/chat/base/net/BaseObserver.java:105`), instead of being silently swallowed.
- A new `finally` block preserves the loop invariant on **all** exit paths (normal / `Exception` / `Error` propagating out): it either schedules the next sweep or sets `flameLoopRunning=false`. This closes the exact "cleanup permanently disabled" risk that would otherwise arise from letting `Error` unwind — `flameLoopRunning` can no longer be stranded at `true`.

Crash semantics for OOM are now identical to pre-batch behavior, the burn-after-reading sweep can no longer be silently and permanently stalled, and the just-OOM'd op is not retried. Correct fix.

💬 Non-blocking (carried over, all still non-blocking)

🟡 Gradle verification could not be run here (no Java runtime). Intended: `./gradlew :wkbase:testDebugUnitTest :wkim:compileDebugJavaWithJavac :wkuikit:compileDebugJavaWithJavac`. Compile/test status remains unverified in this environment.

🔵 The temporary heap diagnostics are well-gated but now live in main source across module boundaries (`WKHeapProbe` in `wkim/src/main/java/com/xinbida/wukongim/utils/WKHeapProbe.java:22`, parsing at `wkbase/src/main/java/com/chat/base/net/FastJsonResponseBodyConverter.kt:70`, sync at `wkim/src/main/java/com/xinbida/wukongim/manager/ConversationManager.java:497`). Track their removal after the incident closes.

🟡 Slow-write logging is not throttled unlike slow-query logging (`WKDBHelper.reportSlowWrite` at `wkim/src/main/java/com/xinbida/wukongim/db/WKDBHelper.java:756`); a device stuck in a slow-write state adds disk/log pressure during the same failure mode being diagnosed.

✅ Highlights

- WAL open-flag change (`WKDBHelper.java:138`) directly addresses the single-connection pool issue.
- Member cache returns defensive copies and handles negative caching/invalidation carefully (`ChannelMembersDbManager.java:100`, `:343`).
- Sync batching + skipped extra re-query (`ConversationManager.java:667`, `MsgDbManager.java:1074`) meaningfully reduce peak memory.
- Streaming JSON converter has focused regression coverage (`FastJsonStreamConverterTest.java`).
- This re-review's error-handling fix is minimal, correct, and matches the batch's Bugly-visibility intent.

yujiawei (approved):
# Code Review — PR #123 (octo-android)

Re-review at head `df4cdb9414ddc85dab9a582c493e7fab7cdf6ebe`, merge-base `4922864` (unchanged — no rebase drift since the last round). Full PR: 26 files, +2210/−337. Delta since the previously reviewed head `cf5571e5`: 7 commits, 6 files, +172/−14.

**Note on method.** This review environment has a working JDK and SQLite, which earlier rounds did not. So rather than reason about the load-bearing claims, I executed them: I ran the two parser paths against each other on a 3.2M-char payload, applied all 24 migrations to a real SQLite database and diffed `EXPLAIN QUERY PLAN` with and without the three new indexes, differential-tested the `LIKE` → in-memory rewrite, and built a deterministic interleaving harness for the member-cache publication protocol. Results are inline below. One of those experiments contradicts a claim I made in my own previous round; it is corrected here.

---

## 1. Specification / scope compliance

**Spec: ✅**

The delta is a pure response to review feedback — no new features, no scope growth, no new flags or public API. Every item raised in the previous round is addressed:

| Prior finding | Resolution at `df4cdb94` | Verified |
|---|---|---|
| **Blocking:** `scheduleFlameSweep` had no exception guard → `flameLoopRunning` stuck `true`, burn-after-reading cleanup permanently disabled | `try`/`catch (Exception)`/`finally`; `keepGoing` initialised `false` so an `Error` propagating out still runs the `finally` and releases the flag | ✅ `MsgModel.java:149-176` |
| `Error` swallowed silently in release (raised after my round) | `catch` narrowed to `Exception`; `Error` propagates via `RxJavaPlugins.onError` → uncaught handler → crash reporter, unchanged from `main` | ✅ `MsgModel.java:154` |
| `putCache` generation check not atomic with the `put` | Second generation re-check added after the `put` | ⚠️ closes the two `memberCache`-scanning paths but **not** `invalidateMemberUid` — see P2 below |
| Cache handed out shared mutable instances, mutated in place by `ChatActivity:1672-1675` | `query` fast path and `querySlowPath` double-check now return `copyOf(...)`; `putCache` also stores a copy | ✅ all 19 fields of `WKChannelMember` covered, `extraMap` copied one level |
| `invalidateChannel` leaked `keysByMemberUid` entries | Reverse index pruned in the same loop | ✅ `ChannelMembersDbManager.java:173-184` |
| Two 512 KB char buffers pinned across the parse | `head = CharArray(0)` right after `unread` | ✅ `FastJsonResponseBodyConverter.kt:75` |
| `SlowQueryCursor` missed the `while (moveToNext())` shape | `moveToNext` override added | ✅ `WKDBHelper.java:615-627` |
| Release-mode `[slow-db]` logging unthrottled | `shouldLogSlow` — 60 s per-SQL throttle, 200-key cap | ✅ `WKDBHelper.java:527-540` |
| No malformed-JSON / empty-body converter tests | Both added, asserting the same exception type across the size-keyed split | ✅ and executed — see below |
| Confirm the `UserModel` second-pass deletion was intentional | Still unanswered (a question, not a code change) | ⬜ please confirm |

`CrashReport` is `api`-scoped in `wkbase/build.gradle:135`, so the `BaseObserver` import is not a build break.

---

## 2. Code quality

**Quality: Approved** — no P0/P1. The P2s below are worth doing, and the first one is worth doing *now* rather than as a follow-up, because it ships an invariant comment that is not true.

### P2 — the reverse-index add sits outside the generation window, so `invalidateMemberUid` can still miss a just-published stale entry

`wkim/src/main/java/com/xinbida/wukongim/db/ChannelMembersDbManager.java:100-118`

```java
memberCache.put(key, member == null ? NULL_MEMBER : copyOf(member));
if (cacheGeneration.get() != generation) {   // check 2
    memberCache.remove(key);
    return;
}
keysByMemberUid.computeIfAbsent(memberUid, k -> ConcurrentHashMap.newKeySet()).add(key);
```

The second check closes the window for the invalidation paths that scan `memberCache` directly — `invalidate(channelId, channelType, uid)` at `:147` and `invalidateChannel` at `:171`. It does **not** close it for `invalidateMemberUid` at `:190`, which is the one path that reaches `memberCache` *through* the reverse index:

```java
cacheGeneration.incrementAndGet();
Set<String> keys = keysByMemberUid.remove(memberUid);
if (keys == null) return;
for (String key : keys) memberCache.remove(key);
```

If the rename lands after check 2 but before the `add`, then `keysByMemberUid.remove(memberUid)` runs while `key` is not yet in the set, so nothing is removed from `memberCache` — and the `add` immediately afterwards re-registers the key. The stale entry survives in a cache with no TTL.

I confirmed this is real rather than theoretical with a deterministic harness that replicates the exact code and drives that single interleaving with a latch (not a stress loop):

```
interleaving: read-through publishes, passes gen re-check, THEN a rename fires
  head df4cdb94 ordering (put -> recheck -> index add):
     STALE VALUE SURVIVES: OLD_NICKNAME   reverseIndex={uidA=[grp1|2|uidA]}
  proposed ordering (index add -> put -> recheck):
     cache empty (correct — stale value evicted)
```

**Failure scenario:** a contact sync calls `ChannelDBManager.saveOrUpdateChannels` (→ `invalidateMemberUid`) for a user who just changed their nickname, at the moment a `RecyclerView` bind is backfilling that same member. The old nickname/avatar stays pinned until that member is written again, the channel is deleted, the cache passes 2000 entries, or the account is switched. That is the "wrong name on a message, impossible to reproduce later" class.

**Fix** — move the reverse-index add *before* the `put` so one generation check covers both structures:

```java
keysByMemberUid.computeIfAbsent(memberUid, k -> ConcurrentHashMap.newKeySet()).add(key);
memberCache.put(key, member == null ? NULL_MEMBER : copyOf(member));
if (cacheGeneration.get() != generation) {
    memberCache.remove(key);
    Set<String> s = keysByMemberUid.get(memberUid);
    if (s != null) s.remove(key);
}
```

Either interleaving is then safe: if `keysByMemberUid.remove(uid)` runs before the add, the re-check sees the bumped generation and evicts; if it runs after, it sees the key and evicts. (An orphaned empty set is harmless, consistent with the reasoning already in `invalidateChannel`.)

**Independently of the code**, please correct the Javadoc. `:78-82` now asserts that the post-`put` check "才是真正关窗口" and `:110-112` that "两种交错都安全". With `invalidateMemberUid` in the picture that is false, and a comment claiming a guarantee the code does not provide is how the next person introduces a genuine bug here. I graded the previous, structurally identical race P2 and am staying consistent — but the false invariant should not ship either way.

### P2 — `flameFailureStreak` is never reset on restart, so after one bad episode every restart gets a single attempt

`wkuikit/src/main/java/com/chat/uikit/message/MsgModel.java:125-140`

`keepGoing = ++flameFailureStreak < FLAME_MAX_FAILURE_STREAK` leaves the field at `3` when the loop gives up, and neither `startCheckFlameMsgTimer` (`:136`) nor `stopTimer` (`:127`) clears it. The next `startCheckFlameMsgTimer` therefore gets `++3 = 4 < 3` → `false` on its *first* failure: one attempt instead of the documented three, until a sweep succeeds and resets to `0`. One line in `startCheckFlameMsgTimer`:

```java
if (!flameLoopRunning.compareAndSet(false, true)) return;
flameFailureStreak = 0;
```

(Note this makes the field cross-thread — `startCheckFlameMsgTimer` runs on the main thread while the sweep runs on `WKDbScheduler`. `volatile` or an `AtomicInteger` would then be needed, which also makes the `:124` comment accurate again.)

### P2 — two sweep chains can coexist across a `stopTimer` → `startCheckFlameMsgTimer` pair

Same file, `:143-176`. If `stopTimer()` (logout, `WKUIKitApplication:927`) and then `startCheckFlameMsgTimer()` (chat exit, `ChatActivity:4014`) both run while a sweep is mid-flight, the CAS succeeds and schedules chain B, and the in-flight runnable's `finally` then sees `flameLoopRunning == true` and schedules chain A′. Both chains self-reschedule, so the poll rate doubles.

Bounded and self-healing — `WKDbScheduler` is single-threaded so they cannot overlap in execution, and the first empty result sets the flag `false`, after which the other chain's reschedule early-returns at `:144`. Worth a comment at minimum; a generation/epoch int compared inside the runnable would close it properly.

### P2 — `shouldLogSlow` key cardinality is driven by SQL strings that inline values

`wkim/src/main/java/com/xinbida/wukongim/db/WKDBHelper.java:527-540`

The throttle key is `abbreviate(sql)`, which is stable for parameterised statements but not for the several call sites that interpolate — e.g. `ChannelMembersDbManager.search:217` and `queryWithPage:233` inline `limit queryPage,size`, and `MsgDbManager:1278` inlines an `IN (...)` list. Those produce a fresh key per page, so a paging burst can walk the map to `SLOW_LOG_KEY_CAP`, trigger the wholesale `clear()`, and reset throttling for everything. Normalising the key (strip digits, or hash the statement shape) would make the cap mean what it says. The unbounded-release-IO concern from last round is nonetheless substantially addressed.

Also: keys are truncated at `SQL_LOG_MAX_LEN = 300`, so statements differing only past 300 chars share a throttle slot and under-log.

### P2 — correcting my own previous round: the `LIKE` → `regionMatches` rewrite is *not* exactly equivalent

`wkuikit/src/main/java/com/chat/uikit/fragment/ChatFragment.java:2735-2745`

Last round I asserted this was equivalent to `channel_id LIKE 'parent____%'`. I differential-tested it against real SQLite this time and it diverges in three cases:

```
parent       child           SQLite LIKE  regionMatches  verdict
group_a      group_a1234            True           True  SAME
group_a      groupXa1234            True          False  >>> DIVERGES
ab%          abZZZZZZ               True          False  >>> DIVERGES
GRP1         grp1abcd               True           True  SAME
ГРУППА       группа1234            False           True  >>> DIVERGES
grp1         grp1abc               False          False  SAME
grp1         grp1abcd               True           True  SAME
```

Two mechanisms: `_` and `%` inside the *parent* id are LIKE metacharacters but literals to `regionMatches`; and `regionMatches(true, …)` is Unicode-case-aware whereas SQLite's default `LIKE` folds ASCII only.

**This does not block, and the first mechanism is arguably a fix.** Group ids are ASCII and the thread separator is the literal four-underscore string `ThreadModel.CHANNEL_ID_SEPARATOR = "____"` (`ThreadModel.java:41`), so for real data the two agree; where they differ on `_`, the old `LIKE` was over-matching (treating the separator's underscores as wildcards) and the new code is stricter, i.e. it removes false-positive `@` badges. The Unicode-case case would over-match, but non-ASCII group numbers are not a thing here. I am recording it because my previous "equivalent" claim was too strong and the docstring at `:2732-2734` repeats it.

While there: **both** the old and the new logic match `parent` + *any* ≥4 characters rather than requiring the literal `"____"` separator, so a sibling group whose id starts with the parent's id (e.g. `123` and `1234567`) still cross-triggers an `@` badge. Pre-existing, not a regression — but since this code is already open, `channelId.startsWith("____", parentLen)` would close it.

### P2 — carried over from last round, unaddressed (all fine to defer)

- `ANRWatchdog.java:39-40, 93-94` — `freezeSkipCount` / `lastFreezeSkipDetail` are still never reset after `appendFreezeSkips`, so later reports carry the cumulative count.
- `MsgDbManager.java:104-109` — `guardLoadGateSize` still clears **both** maps wholesale when either exceeds 256, including in-flight entries for other channels, which can re-open a preview gate.
- `ConversationManager.java:664` — nulling `syncChat.conversations[i].recents` on an object that is then handed to `onBack(syncChat)`. Correct today (single reader, verified) but still a landmine.
- `FastJsonResponseBodyConverter.kt:66-77` — the size-keyed split still means two different fastjson parsers (`JSONScanner` vs `JSONReaderScanner`, and `JSONReader.readObject` does not run `handleResolveTask`, so `$ref` back-references resolve differently). The new tests substantially de-risk the conv-sync shape; the structural difference remains.
- `ConversationManager.saveConversationBatch:738` — one transaction per 20 conversations. Documented and defensible for self-healing list data; noting it as a real change to sync atomicity.

---

## Verification I ran

**Parser equivalence — 15/15 checks pass.** Ported the converter's algorithm verbatim and ran it against `JSON.parseObject(String, Type)` on the conv-sync shape (escapes, multi-byte, `\` and `"` fillers):

```
big payload chars = 3247481 (threshold 262144)
  ok:   every field deep-equal across the two parsers (incl. escapes + multibyte)
  ok:   empty body -> null   (both new and legacy path)
  ok:   same exception class on both paths [3, 2, 100]       (JSONException / unclosed.str)
  ok:   same exception class on both paths [300, 5, 2000]    (JSONException)
  ok:   threshold-3 / -1 / +0 / +1 / +3 chars — content length preserved
15/15 checks passed — ALL PASS
```

Worth knowing: on the truncated large payload the exception *class* matches but the *message* differs (`unclosed.str` vs `not close json text, token : ,`). The test asserts on class, and `ResponseExceptionHandle` dispatches on type, so this is fine — just don't tighten that assertion to the message later.

**Migrations — 61/61 statements apply cleanly** across all 24 files into a fresh SQLite 3.45 database, producing all three new indexes; no schema or data mutation.

**The three new indexes do what the PR claims** — `EXPLAIN QUERY PLAN`, without → with:

| Query | Without | With |
|---|---|---|
| `max(order_seq)` (`MsgDbManager:1395`) | `SEARCH … msg_channel_index` | `SEARCH … COVERING INDEX idx_message_channel_order_seq` |
| `max(message_seq) … is_deleted=0` (`:1411`) | `SEARCH … msg_channel_index` | `SEARCH … idx_message_channel_msg_seq` |
| `max(message_seq)` (`:1427`) | `SEARCH … msg_channel_index` | `SEARCH … COVERING INDEX idx_message_channel_msg_seq` |
| `queryWithFlame` (the 89 s scan) | **`SCAN message`** | `SEARCH … idx_message_flame (is_deleted=?)` |
| `queryUndoneChannelIds` | **`SCAN reminders`** | `SEARCH … COVERING INDEX idx_reminders_undone (type=?)` |
| history paging (`:406`) | `SEARCH … msg_channel_index` + **`USE TEMP B-TREE FOR ORDER BY`** | `SEARCH … idx_message_channel_order_seq (… AND order_seq<?)` |
| `queryLastMsgSeqs` subquery | `SEARCH … msg_channel_index` | `SEARCH … COVERING INDEX idx_message_channel_msg_seq` |

Both partial indexes are selected, so the predicate/`WHERE`-clause implication holds in practice. A bonus the PR does not claim: `idx_message_channel_order_seq` also eliminates the temp B-tree sort on the history-paging query.

**`queryLastMsgSeqs` predicate pushdown is byte-identical**, closing the `GROUP_CONCAT`-ordering caveat I raised last round. On 60 conversations (mixed deleted / empty `channel_id`) with random message counts:

```
rows: old=53 new=53
byte-identical  : True
same element set: True
same order      : True
```

The outer plan stays `SCAN conversation` in rowid order in both variants, so the new index cannot perturb element order. No backend order-sensitivity question remains.

Incidental finding from building the schema: `conversation` has a UNIQUE constraint on `(channel_id, channel_type)`, so the `channel_id<>''` filter can drop at most one row per channel type — the pushdown's cost saving comes almost entirely from `is_deleted=0`, not from the empty-id predicate.

---

## What I liked

The response to review was disciplined: every item was fixed at the mechanism rather than papered over, and two of the fixes are better than what I suggested. Narrowing `catch (Throwable)` to `catch (Exception)` while initialising `keepGoing = false` so the `finally` still releases the flag is the right resolution of "don't swallow `Error`" versus "never leave the loop armed-but-dead" — both properties hold simultaneously, and the comment at `:157-163` explains *why* `Error` is deliberately not caught, which is the part that stops someone re-adding it. The `copyOf` route covers all 19 fields including `extraMap`, and the reasoning in `invalidateChannel` about preferring an orphaned empty set over a remove-during-add race is exactly the right trade to make explicit.

---

## 3. Overall verdict

**APPROVE**

Spec compliance is clean, the blocking item from each of the previous two rounds is genuinely fixed, and everything I could execute — parser equivalence, migration application, index selection, `GROUP_CONCAT` ordering — confirms the PR's claims rather than merely failing to contradict them.

Nothing here rises to P0/P1. The member-cache race is real and I have a deterministic reproduction of it, but its consequence is a stale display name that self-heals on any subsequent write to that member, a cache wipe, or an account switch — the same class and severity I graded P2 last round, and I am not going to re-grade an identical failure mode upward just because it is the one still standing. Holding a batch of production ANR/OOM fixes for it would be the wrong trade.

This is the 4th review round on this PR, and findings have converged (1 blocking + 9 advisory → 1 blocking → 0 blocking). That is a healthy trend rather than whack-a-mole, which is the other reason to land this and take the remainder as follow-ups instead of opening round 5.

## 4. Suggested next steps

1. Reorder the reverse-index add before the `put` in `putCache`, **and** correct the two Javadoc passages that claim the window is fully closed. If you'd rather not touch code again in this PR, at least fix the comments here and file the reorder — the code being racy is recoverable, the comment asserting it isn't, is not.
2. Reset `flameFailureStreak` in `startCheckFlameMsgTimer` (and make it `volatile`/atomic if you do).
3. Normalise the `shouldLogSlow` key so the 200-key cap tracks statement shapes, not paged offsets.
4. Confirm the `UserModel:267` second-pass deletion was intentional — still the one open question from last round.
5. Optional, since the code is already open: require the literal `"____"` separator in `hasUndoneThreadMention` to stop sibling-group `@` badge cross-triggering.

## 5. Beyond the diff — things to watch on rollout

- **Snapshot isolation is now observable.** With the pool above 1, a main-thread read no longer blocks behind a background write transaction — it gets the pre-transaction WAL snapshot. That is the fix, but UI computed mid-sync (reminder badges, unread totals) can now render briefly stale where it previously stalled. `setOnRefreshMsg` covers the conversation list; the badge paths are computed independently on the main thread and deserve an explicit pass during the large-account sync test.
- **`-wal` growth is no longer capped.** Dropping `wal_autocheckpoint` / `journal_size_limit` is correctly justified (no global setter exists for either), but the concern those lines were written for is now unmitigated at library defaults. Watch `-wal` size and first-checkpoint latency on a large account after a heavy `saveSyncChat`.
- **Reminder matching is now O(items × undone-rows) on the main thread.** Strictly better than N full scans — the plan above confirms the query is now a covering-index search — but it scales with total undone mentions rather than with the visible list. Sanity-check an account with a large mention backlog.
- **Still main-thread DB.** `queryUndoneChannelIds` and the batched thread-conversation queries are 1 query instead of N, but they still run on the main thread. The debug-only `warnIfMainThread` guard is a good step toward the real follow-up.
- **The PR's own open item stands:** the "history paging won't load" root cause is still unconfirmed against the observed device hang. Four independently valid defects were fixed; if the symptom recurs after this ships, that means none of the four was the cause — not that the fixes were wrong.
- **CI should run `FastJsonStreamConverterTest`.** I executed an equivalent harness outside Gradle and it passes, but I could not run the module's own test task, so the test's compilation against the real module remains unverified here.

PR #125, opened 2026-08-14
fix(im): 成员缓存发布顺序竞态 + 阅后即焚轮询状态机 + 慢查询日志 key

mzanalyticst:
#123 review 最后一轮 APPROVE 时留下的 4 条 P2，其中第 1 条附带一段断言了代码并不成立的不变量的注释。

| commit | 修什么 |
|---|---|
| `40647a4` | `putCache` 改为先发布反向索引与缓存、再复查代际。原顺序只关掉扫 `memberCache` 的两条失效路径，没关 `invalidateMemberUid`（唯一经反向索引到达缓存的路径）；review 用 latch 做了确定性复现。一并更正两处过度承诺的注释。 |
| `413d2b3` | `flameFailureStreak` 在 CAS 成功后清零（原来停在上限值 → 下次只剩 1 次重试而非 3 次），字段改 `volatile`；新增 `flameEpoch` 关掉 `stopTimer` → `startCheckFlameMsgTimer` 之间双链并存导致的轮询频率翻倍。 |
| `1c2db82` | `shouldLogSlow` 的 key 按语句形状归一（数字串 → `?`），200 上限不再被分页偏移走穿后整清。 |

净代码 33 行（去注释）。不碰消息收发 / 存储 / 同步 / UI 渲染。

## 验证

- `:wkim` `:wkuikit` 编译通过；`wkim` + `wkbase` + `wkuikit` 单测 776 个 0 失败。
- 真机：群聊头像/昵称正确、改好友备注后聊天页即时更新、反复进出聊天页 + 退登重登无异常。
- 竞态本身无法手工触发（需要落在几条指令的窗口内），依据是代码推理 + review 那份 latch 复现。

## 不在本 PR 范围

`hasUndoneThreadMention` 要求字面 `____` 分隔符（兄弟群 `@` 角标交叉触发）。既有问题，改动 `@` 角标行为，需单独真机验证。

Closes #124

PR #127, opened 2026-08-16
fix(chat): 重连 sync 完成后刷新会话列表

### LINKED ISSUE #126: [Android] 锁屏挂起恢复后会话列表不刷新
## 现象

App 锁屏挂起一段时间后再打开，左上角状态正常从「连接中」变为「连接上」，但会话列表仍停留在挂起前的内容 —— 顺序、最后一条消息预览、未读数都不更新。

来一条新消息，或者随便进一个聊天窗口再返回，列表就恢复正常。

## 根因

`onResume` 与 WebSocket 重连的时序是错位的：

1. 解锁 → `ChatFragment.onResume()` 重绘一次 —— 此时 socket 还没连上，用的是挂起前的旧数据
2. WebSocket 重连成功 → conversation sync 回来 → `allConversations` 被更新
3. 但增量刷新路径对「已存在会话的字段更新」只做逐行 `notifyItemChanged`，不重建列表

界面就停在第 1 步那一帧。数据是对的，UI 是旧的。

SDK 在 sync 落库结束时会发 `WKConnectStatus.syncCompleted`，但 `ChatFragment` 的连接状态监听器从来没接这个信号。

## 影响范围

- 仅影响会话列表的重绘时机，数据层无问题
- 既有缺陷，非近期回归（`main` 上一直如此）

## 修复方向

在连接状态监听器里接上 `syncCompleted`，收到即刷新一次列表。

mzanalyticst:
Closes #126

## 问题

锁屏挂起恢复后，连接状态正常变为「已连接」，但会话列表停在挂起前那一帧 —— 顺序、预览、未读都不更新。来新消息或进出聊天页才自愈。

## 根因

`onResume` 与重连的时序错位：

1. 解锁 → `onResume` 重绘一次，此时 socket 未连上，用的是旧数据
2. 重连成功 → conversation sync 回来 → `allConversations` 已更新
3. 但增量路径对「已存在会话的字段更新」只做逐行 `notifyItemChanged`，不重建列表

数据是对的，UI 是旧的。SDK 在 sync 落库结束时会发 `WKConnectStatus.syncCompleted`（`ConversationManager:698`），但 `ChatFragment` 的连接状态监听器从来没接这个信号。

## 改动

`ChatFragment` 的监听器新增 `syncCompleted` 分支，收到即 `filterAndDisplay()` 一次。

数据就绪性由主线程队列 FIFO 保证：`setOnRefreshMsg`（`ConversationManager:682`）先于 `setConnectionStatus`（`:698`）投递，两者都走 `BaseManager.runOnMainThread` → `Handler.post`。真机实测两者相隔 7ms，顺序稳定。

排序不需要额外处理：`filterAndDisplay` → `buildRecentDisplayList()` 每次 build 都按 `top` + `lastMsgTimestamp` 重排（`ChatFragment:3028`）；关注 tab 的顺序由 category 配置决定，与时间戳无关。这条重绘路径正是「来新消息时列表刷新」走的同一个，线上已验证。

顺带给 `WKUIKitApplication:320` 的连接日志加 `BuildConfig.DEBUG` 门控（本仓库 `setDebug(true)` 写死，裸日志会进 release 包）。

## 影响范围

- 纯新增分支，不修改任何既有判断
- 只碰 UI 重绘时机，未触碰数据层、消息收发、Space 过滤、列表构建逻辑
- `syncCompleted` 原本在该监听器中无任何处理
- 既有缺陷，非 #123 / #125 引入（`main` 上一直如此）

## 测试

- `:wkuikit:compileDebugJavaWithJavac` / `:wkuikit:testDebugUnitTest` / `:wkim:testDebugUnitTest` 通过
- 真机：挂起 16 分钟后回前台，日志确认 `syncCompleted -> filterAndDisplay` 触发，列表立即正确，单次重建约 150ms，无卡顿
- 回归：正常收发消息、列表滑动未见异常

### DIFF
--- a/wkuikit/src/main/java/com/chat/uikit/WKUIKitApplication.java
+++ b/wkuikit/src/main/java/com/chat/uikit/WKUIKitApplication.java
@@ -317,7 +317,7 @@ public class WKUIKitApplication {
                         .postDelayed(this::startChat, 100);
                 return;
             }
-            Log.e("去连接", "-->");
+            if (BuildConfig.DEBUG) Log.e("去连接", "-->");
             WKIM.getInstance().getConnectionManager().connection();
         }
     }
--- a/wkuikit/src/main/java/com/chat/uikit/fragment/ChatFragment.java
+++ b/wkuikit/src/main/java/com/chat/uikit/fragment/ChatFragment.java
@@ -1182,6 +1182,22 @@ public class ChatFragment extends WKBaseFragment<FragChatConversationLayoutBindi
                         });
                     }
                 }
+            } else if (i == WKConnectStatus.syncCompleted) {
+                // 会话 sync 落库完成 —— 重绘一次列表。
+                //
+                // 锁屏挂起恢复的时序是错位的：onResume 先跑（用挂起前的旧数据重绘一次）→
+                // WebSocket 才开始重连 → sync 回来更新 allConversations，但增量路径对
+                // 「已存在会话的字段更新」只做逐行 notifyItemChanged，不重建列表 → 界面停在
+                // onResume 那一帧。表现为「状态显示已连接、列表还是旧的」，直到来新消息或
+                // 进聊天页返回才自愈。
+                //
+                // 这里的数据是就绪的：setOnRefreshMsg 与本状态都经 BaseManager.runOnMainThread
+                // 投到主线程队列，且前者先投（ConversationManager 内 :682 早于 :698），
+                // FIFO 保证 allConversations 已更新完再走到这里。
+                if (BuildConfig.DEBUG) {
+                    android.util.Log.d("ConvSync", "[ConvSync] syncCompleted -> filterAndDisplay");
+                }
+                filterAndDisplay();
             } else if (i == WKConnectStatus.connecting) {
                 wkVBinding.textSwitcher.setText(getString(R.string.connecting));
                 wkVBinding.spaceChevronIv.setVisibility(View.GONE);


Jerry-Xin (approved):
This relevant, narrowly scoped fix correctly refreshes the conversation list after reconnection synchronization completes.

💬 Non-blocking

- 🔵 Suggestion: Add a regression test covering `syncCompleted` after existing conversations receive timestamp, preview, or unread-count updates. This would protect the required refresh and reordering behavior. (`wkuikit/src/main/java/com/chat/uikit/fragment/ChatFragment.java:1185`)

✅ Highlights

- `syncCompleted` (value 6) is distinct from `success` (1)/`syncMsg` (3), so the listener's tag-dedup guard does not suppress it during a reconnect sequence.
- `syncCompleted` is dispatched after the conversation refresh callback is queued: in `ConversationManager.saveSyncChat`, `setOnRefreshMsg(...)` is posted before `setConnectionStatus(syncCompleted, ...)`, and both go through `BaseManager.runOnMainThread` → same main-thread `Handler`, so FIFO guarantees `allConversations` is updated before `filterAndDisplay()` runs.
- `filterAndDisplay()` already handles lifecycle deferral (not resumed) and active scrolling safely, and rebuilds both adapters.
- The rebuild re-sorts conversations using current pin and timestamp values via `buildRecentDisplayList()`.
- Release logging is now correctly guarded by `BuildConfig.DEBUG` (`wkuikit/src/main/java/com/chat/uikit/WKUIKitApplication.java:320`).
- No security, error-handling, or blocking performance issues were found.

yujiawei (approved):
# Code Review — PR #127 (octo-android)

Reviewed against head `79e291bc`, which sits directly on top of `origin/main` (`e7d5a29f`). Diff verified as exactly 2 files, +17/-1.

## 1. Spec compliance

**Spec: ✅**

Issue #126 asks for one thing: *"hook `syncCompleted` in the connection-status listener and refresh the list once when it arrives."* That is precisely what the diff does.

- **Missing:** none.
- **Extra:** one disclosed out-of-scope hunk — `WKUIKitApplication.java:320`, gating `Log.e("去连接", "-->")` behind `BuildConfig.DEBUG`. It is unrelated to #126, but it is declared in the PR description, is a one-line log guard, and carries no behavioural risk. Accepted rather than blocked; noted so the scope gate is not silently eroded.
- **Deviation:** none. The change is purely additive — `syncCompleted` previously matched no branch in the `if/else` chain, and the trailing `setTag(i)` / `startConnectTimer()` logic at `ChatFragment.java:1218-1224` behaves identically before and after.

### Root-cause claims verified

I re-derived the PR's central ordering argument rather than taking it on trust:

- `ConversationManager.saveSyncChat` calls `setOnRefreshMsg(uiMsgList, "saveSyncChat")` at `:682`, then `setConnectionStatus(WKConnectStatus.syncCompleted, "")` at `:698`. Order confirmed in source.
- Both dispatch through `BaseManager.runOnMainThread` (`BaseManager.java:18-26`), which either runs inline (already on the main thread) or `Handler.post`s. `ConversationManager` and `ConnectionManager` are separate `BaseManager` instances with separate `Handler`s, but both target the same main `Looper`; `Handler.post` enqueues at `uptimeMillis()` and equal-`when` messages preserve insertion order. Posted from the same worker thread in that order, FIFO holds. **The PR's data-readiness claim is correct.**
- Reconnect emits `success(1) → syncMsg(3) → syncCompleted(6) → success(1)` (`WKConnection.java:664-721`). No two consecutive identical values, so the new branch is genuinely reachable through the dedupe guard at `ChatFragment.java:1149-1157`.
- `filterAndDisplayInternal` (`ChatFragment.java:2795-2819`) is properly guarded: `followAdapter == null || getActivity() == null || !isAdded()` bail out; `!isResumed()` defers via `pendingFilterAndDisplay`, consumed in `onResume` at `:4373`; non-idle scroll defers via `pendingScrollIdleRefresh`, consumed by `scrollIdleWatcher` registered at `:393`/`:400`. No lifecycle crash surface added.
- `buildRecentDisplayList` (`:2932`) is a pure re-projection of `allConversations` with no Space filtering of its own, so the rebuild cannot drop conversations that the incremental path admitted. No data-loss regression.
- `BuildConfig` resolves in both files: `ChatFragment` imports `com.chat.uikit.BuildConfig` at `:80`; `WKUIKitApplication` is in package `com.chat.uikit`; `wkuikit/build.gradle` sets `buildConfig true` and `namespace 'com.chat.uikit'`. Compiles.
- Precedent exists: `ChatActivity.java:1825` already consumes `syncCompleted` for card-frame compensation, so this is a consistent use of the signal.

## 2. Code quality

**Quality: Approved** — no P0/P1. The findings below are all P2 and none of them should block merge.

### P2-1 — `syncCompleted` is an event, but the listener's dedupe guard treats it as a state

`ChatFragment.java:1149-1157` early-returns whenever the incoming status equals the last one stashed in `textSwitcher`'s tag. Two consecutive `syncCompleted` emissions with no intervening status change therefore skip `filterAndDisplay()` entirely.

```java
if (tag instanceof Integer) {
    int tag1 = (int) tag;
    if (tag1 == i) {
        return;   // ← swallows a repeat syncCompleted
    }
}
```

Reachable in practice: `setSyncConversationListener` has multiple callers that all terminate in `saveSyncChat` → `syncCompleted` (`ChatFragment.java:1179`, `:2054`, `:2448`, `WKIMUtils.java:1191`, `WKConnection.java:737`). A Space-switch sync followed by a resync, with no connection transition between them, drops the second refresh.

Not a regression — before this PR *no* `syncCompleted` refreshed anything, so behaviour is strictly better. But the fix is less reliable than it reads. Suggestion: handle `syncCompleted` above the dedupe guard, or exclude it from the tag comparison, so an event-shaped signal is not deduped by state-shaped logic.

### P2-2 — For the new-conversation path, the FIFO argument does not fully hold

The PR's ordering argument covers `setOnRefreshMsg` → `syncCompleted`. But when sync brings *new* conversations, the refresh callback calls `sortMsg(uiList)` (`ChatFragment.java:1141`), and `sortMsg` writes `allConversations` only inside a nested `AndroidUtilities.runOnUIThread` block (`:2615-2626`: `clearAllConversations(); allConversations.addAll(tempList); … filterAndDisplay();`).

That nested runnable is enqueued *behind* the already-queued `syncCompleted` message. So the sequence becomes: refresh callback → `syncCompleted` rebuild **without the new conversations** → nested runnable updates `allConversations` and rebuilds again.

The end state is correct (the second rebuild fixes it) and the reported bug — field updates on *existing* conversations, where `uiList` is empty and `sortMsg` never runs — is unaffected. Net effect is one wasted full rebuild plus a possible one-frame flicker, not a correctness bug. Worth a comment amendment, since the current comment at `:1191-1194` asserts a stronger guarantee than actually holds.

### P2-3 — Unconditional full rebuild on every `syncCompleted`, including main-thread DB reads

`filterAndDisplay()` rebuilds **both** tabs. `buildFollowDisplayList` performs a synchronous `ReminderDBManager.getInstance().queryUndoneChannelIds(...)` on the main thread (`:2848-2850`), and `buildRecentDisplayList` snapshots `getWkChannel()` per row (`:2992-2996`), which falls into `ChannelManager`'s slow path for rows without a cached channel.

`syncCompleted` fires on every reconnect, so network flapping means a rebuild per reconnect. The 150 ms figure in the PR description is plausible for the tested account but scales with conversation count. This codebase already carries scar tissue from exactly this shape — see the comment at `ConversationManager.java:303-306` documenting a main-thread stall against a `saveSyncChat` write transaction. It is pre-existing for the other `filterAndDisplay` call sites and not introduced here, but this PR adds one more trigger at a DB-busy moment. Consider a short coalescing window if reconnect churn shows up in ANR traces.

### P2-4 — `lastFullRefreshTime` side effect on the `onResume` throttle

`filterAndDisplayInternal` sets `lastFullRefreshTime` (`:2816`). `onResume` uses it against `RESUME_THROTTLE_MS = 2000` (`:197`, `:4379`) to choose the light path, which skips `loadCategories()`. A `syncCompleted` rebuild immediately before the user enters and exits a chat will silently suppress the next category reload. Low impact, and arguably the throttle's intent, but it is a behaviour coupling introduced by adding this call site.

### P2-5 — Log gating is incomplete for its own stated rationale

The PR gates `Log.e("去连接", "-->")` at `WKUIKitApplication.java:320`, but `Log.e("接受登录", "-->3")` at `:746` is the same class of bare debug marker and remains ungated in release builds. `:280` and `:299` are legitimate error logs and should stay. Either finish the pass or drop it from this PR.

### Testing

No automated coverage is added for the new branch. Given that `ChatFragment` is not unit-testable without instrumentation, the on-device verification described in the PR is a reasonable substitute — but the listed unit test suites do not exercise this change, so they should not be read as evidence for it.

## 3. Overall verdict

**APPROVE**

The diagnosis is accurate, the ordering argument checks out against the SDK source, the change is genuinely additive, and it is guarded correctly against the fragment lifecycle. The P2 items above are follow-ups, not merge blockers.

## 4. Suggestions

1. Move the `syncCompleted` handling above the tag-dedupe guard (P2-1) — this is the one item with a real chance of the fix silently not firing.
2. Amend the comment at `:1191-1194` to scope the FIFO guarantee to the existing-conversation path, and note that the new-conversation path resolves via `sortMsg`'s own rebuild (P2-2).
3. Either finish the `Log` gating pass in `WKUIKitApplication` or move it to a separate cleanup PR (P2-5).

## 5. Additional observations

- `startConnectTimer()` runs for `syncCompleted` (`:1219`), briefly arming the disconnect-screen timer while the connection is healthy. It is cancelled by the subsequent `success` emission from `markConnected`, and this is pre-existing behaviour untouched by the diff — flagged only because this PR is the first code to deliberately handle `syncCompleted` in this listener, which makes the interaction newly relevant.
- The per-item `notifyRecycler` work done by the refresh callback immediately before the `setList`/`notifyDataSetChanged` rebuild is now redundant on this path. Harmless, but it doubles the bind work at reconnect.

PR #85, opened 2026-07-11
feat(conversation): 1v1 未读消息显示与群聊@我同等的高优先级深红提示

pkuWMH:
## 需求

会话列表里，群聊中「有人@我」会显示醒目的深红色高优先级提示（`reminderColor #FF5353` + `[有人@我]` 文案）。但**一对一（1v1）聊天的未读消息没有这个提示**——而 1v1 中对方发来的每条消息，本质上都等价于「@我」（点对点直达）。

产品口径（方案 A）：只要是 1v1 未读（unread>0），即视同高优先级。

Closes #86

## 设计原则

深红提示不是群聊专属样式，它是一个**注意力分级信号**——「有人在等我响应」。在人机协同场景里，核心是帮人从消息洪流里识别出真正需要关注的信息。1v1 里"对方发消息"即等价 mention，因此**复用同一视觉信号（深红色）**，触发条件在 1v1 退化为 `unread>0`。

但语义上 1v1 未读 ≠ 群聊被 @，所以文案用独立的 `[未读]/[unread]`，不复用 `[有人@我]/[@me]`。

## 改动

**纯前端本地判定，不改后端、不改 schema。**

`wkuikit/.../chat/adapter/ChatConversationAdapter.java`，仅 `showReminders()`（`TYPE_NORMAL` 路径，1v1 行实际渲染处）：

高优先级判定从"仅命中群聊 mention reminder"扩展为"命中 mention reminder **或** 满足全部下列条件的 1v1"：

```java
channelType == PERSONAL
    && !WKSystemAccount.isSystemAccount(channelID)   // 排除系统团队/文件助手
    && !isMuted1v1                                    // 排除免打扰 1v1
    && getRenderUnreadCount(item) > 0                 // 走 Space 过滤路径，不用 raw unreadCount
```

三个关键修正（对齐 review）：
1. **走 `getRenderUnreadCount()` 而非 raw `unreadCount`**：适配器既有契约用 `getRenderUnreadCount()` 把跨 Space 污染的未读清零、`getSpaceFilteredWkMsg()` 隐藏跨 Space 预览。若读 raw 字段，一个 badge 已归零、预览已 null 的跨 Space 私聊仍会点亮深红，夸大紧急度。改后与 `setUnreadCount` 的路由一致。
[...]

PR #129, opened 2026-08-16
fix: Bugly 崩溃批次 —— 联系人搜索 NPE、RecyclerView 状态损坏、关库后残留 DB 任务

### LINKED ISSUE #128: [Android] Bugly 崩溃批次：联系人搜索 NPE、RecyclerView 状态损坏、关库后残留 DB 任务
Bugly 上报的三类崩溃，均已定位根因并完成修复（分支 `fix/crash-batch-01`，实机验证通过）。

## 1. `ChooseContactsActivity.searchUser` NPE（上报量最大）

`allList` 只在异步回调 `buildAndDisplayList()` 里赋值。页面进来自动弹键盘，用户在成员请求返回前打字 → `allList.size()` 空指针。线上日志时间线吻合：建页 → 2s 后弹键盘 → 又 1.3s 后敲字崩溃，请求 3.4s 还没回来。

顺带修一个被崩溃掩盖的 UX 问题：数据到达时无条件 `setList(allList)`，会把用户加载期间已输入的搜索词吞掉。

## 2. RecyclerView 状态损坏（两条崩溃同源）

- `Cannot call this method while RecyclerView is computing a layout`
- `Tmp detached view should be removed from RecyclerView before it can be recycled`

两次上报的日志是同一个三段式：

```
去连接: -->                                     ← 回前台 / 重连
error : The specified child already has a parent...   ← SwipeBackLayout.onLayout 吞掉
FATAL                                           ← 18ms / 151ms 后
```

中间那条异常从 RecyclerView 的 layout 里逃逸后，`onExitLayoutOrScroll()` 不再执行 → 计数器永久 >0；临时 detach 的 holder 停在 `tmpDetached` → 回收时抛。崩溃信息里的 `tmpDetached no parent` 是直接证据。

**根因（那个重复 addView）尚未定位**，已加 `CrashReport.postCatchedException` 捞真实堆栈；本批次先把两条下游路径分别堵死。已排除：消息 provider 的 view 缓存、`RecyclerAnimationScrollHelper`（死代码）、msgeffect overlay、长按菜单、panel tab 切换、pinned 视图。

## 3. `Cannot perform this operation because the connection pool has been closed`

撤回兜底重试延迟 500/1500ms 查库，这中间用户退出登录或 401 关库 → 拿已关闭的连接池。

必崩而非被吞的原因：RxJava 3.1.11 `ScheduledDirectTask.call()` 会 `RxJavaPlugins.onError(ex)` 再 rethrow，本仓库没设 errorHandler，`IllegalStateException` 被判定为 bug 类 → 直送 UncaughtExceptionHandler。即任何在 `WKDbScheduler` 上抛出的异常都必崩，全仓 18 处调用点只有阅后即焚那条链此前有 try/catch。

## 验收

1. 弱网/飞行模式进「选择联系人」，键盘弹出立刻打字
2. 群里改成员备注、撤回消息、快速滑动时收消息、后台切回重连
3. 撤回消息后 2 秒内退出登录
4. 消息列表刷新频繁场景观察有无视觉回归

## 不在本批次

会话同步 OOM（`insertOrReplaceExtra → queryWithMsgIds`，`DispatchQueuePool` 线程）已由 #122 / `df2b938` 修复，该上报为存量老版本。判断依据：崩溃栈里是**单参数**重载，正是那次 diff 删掉的行。

mzanalyticst:
Closes #128

三类线上崩溃，3 笔提交一一对应。实机验证通过，`assembleDebug` 全量通过。

## 1. `fix(contacts)` 联系人选择页搜索 NPE

`allList` 只在异步回调里赋值，页面自动弹键盘 → 用户在成员请求返回前打字就崩。改为声明处初始化空列表，永不为 null，顺带消掉 `rightButtonClick()` 里同样的隐患。

另修一个被崩溃掩盖的 UX 问题：数据到达时无条件 `setList(allList)` 会吞掉用户加载期间已输入的搜索词，现在记 `searchKey` 数据到齐后重新过滤。

## 2. `fix(chat)` RecyclerView 两类崩溃（同源）

两次上报日志是同一个三段式：`去连接` → `already has a parent`（`SwipeBackLayout.onLayout` 吞掉）→ 18ms/151ms 后崩。异常从 RecyclerView 的 layout 里逃逸 → `onExitLayoutOrScroll()` 不执行、holder 停在 `tmpDetached`，此后任何 notify / 回收都崩。

- `safeAdapterAction` 加 FIFO 排队 + 3 次有界重试。原实现 post 一次后仍无条件 `run()`，卡死时照样崩；且新回调可能插到旧回调前面导致同一条消息的刷新乱序。RecyclerView 空闲且无排队时仍是同步 `run()`，与改动前一致。
- 补齐 8 处漏网的 IM/网络回调侧适配器操作（`sendMsgInserted` / `typing` / `receivedMessages` 早已在用这个模式）。
- `MyItemAnimator` 补 `setSupportsChangeAnimations(false)`。change 动画当初是靠**注释掉动画代码**「关」的，flag 一直是 true，不带 payload 的 `notifyItemChanged(i)` 仍走旧/新 holder 双份路径 —— 正是撞上脏 `tmpDetached` 的那条路。关掉后该路径整个消失。
- `SwipeBackLayout` 的 catch 保留，但换成 debug 门控日志 + `CrashReport.postCatchedException`。吞掉换来的存活时间只有 18/151ms，等于零，却把现场埋了。

> ⚠️ 根因那个重复 `addView` 尚未定位（已排除 provider view 缓存、`RecyclerAnimationScrollHelper` 死代码、msgeffect overlay、长按菜单、panel tab 切换、pinned 视图）。本 PR 先堵死两条下游路径 + 加非致命上报捞堆栈，拿到后另开 PR 一击修掉。

## 3. `fix(db)` 关库后残留 DB 任务

撤回兜底重试延迟 500/1500ms 查库，中间用户登出/401 关库 → 拿已关闭的连接池。

必崩原因：RxJava 3.1.11 `ScheduledDirectTask.call()` 是 `catch { RxJavaPlugins.onError(ex); throw ex; }`，本仓库没设 errorHandler，`IllegalStateException` 被判定为 bug 类 → 直送 UncaughtExceptionHandler。**任何**在 `WKDbScheduler` 上抛出的异常都必崩。

`WKDbScheduler` 补 `submit()` 把守卫收进调度器内部，只丢弃 `connection pool has been closed` 这一种，其余原样抛出。守卫必须包在交给 RxJava **之前**的 Runnable 上，包在外层 Executor 上来不及。18 处调用点机械替换，`get()` 标 `@Deprecated`。

## Review 提示

`ChatActivity` 的 diff 显示 +183 行，但**忽略空白后只有 +51**（`?w=1` 看），差额全是代码块包进 lambda 后的整体缩进。

## 测试计划

- [x] 弱网/飞行模式进「选择联系人」，键盘弹出立刻打字
- [x] 群里改成员备注、撤回消息、快速滑动时收消息、后台切回重连
- [x] 撤回消息后 2 秒内退出登录
- [x] 消息列表刷新频繁场景，确认关掉 change 流程后无视觉回归
- [x] `./gradlew assembleDebug`

## 不在本 PR

会话同步 OOM（`insertOrReplaceExtra → queryWithMsgIds`）已由 #122 / `df2b938` 修复，相关上报为存量老版本 —— 崩溃栈里是**单参数**重载，正是那次 diff 删掉的行。

### DIFF
--- a/wkbase/src/main/java/com/chat/base/ui/components/SecretDeleteTimer.java
+++ b/wkbase/src/main/java/com/chat/base/ui/components/SecretDeleteTimer.java
@@ -144,7 +144,7 @@ public class SecretDeleteTimer extends FrameLayout {
             EndpointManager.getInstance().invoke("deleteRemoteMsg", clientMsgNo);
             // DB 删除操作放 IO 线程，onDraw 中做 DB 操作会导致 ANR
             final String msgNo = clientMsgNo;
-            WKDbScheduler.get().scheduleDirect(() ->
+            WKDbScheduler.submit(() ->
                 WKIM.getInstance().getMsgManager().deleteWithClientMsgNO(msgNo)
             );
             return;
--- a/wkbase/src/main/java/com/chat/base/utils/AppExecutors.java
+++ b/wkbase/src/main/java/com/chat/base/utils/AppExecutors.java
@@ -42,7 +42,7 @@ import java.util.concurrent.atomic.AtomicInteger;
  *   <li>{@link #background()} —— CPU 密集（JSON parse、图形运算、差分等）。
  *   池大小 = CPU 核心数，避免过度并发造成 context switch。</li>
  *   <li>{@link #db()} —— 单线程顺序化 DB 任务（非 Rx 场景；Rx 场景请继续走
- *   {@link WKDbScheduler#get()}）。</li>
+ *   {@link WKDbScheduler#submit(Runnable)}）。</li>
  *   <li>{@link #mainThread(Runnable)} / {@link #postDelayed(Runnable, long)} ——
  *   投递到主线程 Looper；等价于 {@code AndroidUtilities.runOnUIThread}，但不依赖它。</li>
  * </ul>
--- a/wkbase/src/main/java/com/chat/base/utils/WKDbScheduler.java
+++ b/wkbase/src/main/java/com/chat/base/utils/WKDbScheduler.java
@@ -16,9 +16,15 @@
 
 package com.chat.base.utils;
 
+import android.util.Log;
+
+import com.chat.base.BuildConfig;
+
 import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 
 import io.reactivex.rxjava3.core.Scheduler;
+import io.reactivex.rxjava3.disposables.Disposable;
 import io.reactivex.rxjava3.schedulers.Schedulers;
 
 /**
@@ -27,6 +33,14 @@ import io.reactivex.rxjava3.schedulers.Schedulers;
  * 避免多线程并发争抢 SQLCipher 连接池导致连接饥饿或 ANR。
  */
 public final class WKDbScheduler {
+
+    private static final String TAG = "WKDbScheduler";
+
+    /**
+     * SQLCipher 连接池关闭后再取连接抛出的固定文案。SQLCipher 抛的是裸
+     * {@link IllegalStateException}，没有专门的异常类型，只能按文案识别。
+     */
+    private static final String POOL_CLOSED_MSG = "connection pool has been closed";
 
     private static final Scheduler INSTANCE = Schedulers.from(
             Executors.newSingleThreadExecutor(r -> {
@@ -39,7 +53,55 @@ public final class WKDbScheduler {
     private WKDbScheduler() {
     }
 
+    /**
+     * 提交一个数据库任务，串行执行。
+     *
+     * <p>登出 / token 失效（401）会关掉数据库，但此前已排队或延迟提交的任务仍会跑到，
+     * 拿已关闭的连接池就抛 {@link IllegalStateException}。RxJava 的 ScheduledDirectTask
+     * 会把它交给 {@code RxJavaPlugins.onError}，默认实现直接调用当前线程的
+     * UncaughtExceptionHandler —— 进程当场崩在 wk-db-worker 上。
+     *
+     * <p>库都关了这些任务本来就没有意义，丢弃即可。只丢这一种，其余异常原样抛出，
+     * 不掩盖真实问题。
+     */
+    public static Disposable submit(Runnable task) {
+        return INSTANCE.scheduleDirect(guard(task));
+    }
+
+    /**
+     * 延迟提交，语义同 {@link #submit(Runnable)}。
+     */
+    public static Disposable submit(Runnable task, long delay, TimeUnit unit) {
+        return INSTANCE.scheduleDirect(guard(task), delay, unit);
+    }
+
+    private static Runnable guard(Runnable task) {
+        return () -> {
+            try {
+                task.run();
+            } catch (IllegalStateException e) {
+                if (!isPoolClosed(e)) throw e;
+                if (BuildConfig.DEBUG) Log.w(TAG, "db closed, drop pending task", e);
+            }
+        };
+    }
+
+    /**
+     * @deprecated 直接拿 Scheduler 会绕过 {@link #submit(Runnable)} 的关库保护，
+     * 新代码一律用 {@link #submit(Runnable)}。
+     */
+    @Deprecated
     public static Scheduler get() {
         return INSTANCE;
+    }
+
+    private static boolean isPoolClosed(Throwable e) {
+        for (Throwable t = e; t != null; t = t.getCause()) {
+            if (t.getMessage() != null && t.getMessage().contains(POOL_CLOSED_MSG)) {
+                return true;
+            }
+            if (t.getCause() == t) break;
+        }
+        return false;
     }
 }
--- a/wkbase/src/main/java/com/chat/base/views/swipeback/SwipeBackLayout.java
+++ b/wkbase/src/main/java/com/chat/base/views/swipeback/SwipeBackLayout.java
@@ -31,7 +31,9 @@ import android.widget.FrameLayout;
 
 import androidx.core.view.ViewCompat;
 
+import com.chat.base.BuildConfig;
 import com.chat.base.R;
+import com.tencent.bugly.crashreport.CrashReport;
 
 import java.util.ArrayList;
 import java.util.List;
@@ -406,7 +408,12 @@ public class SwipeBackLayout extends FrameLayout {
                         mContentLeft + mContentView.getMeasuredWidth(),
                         mContentTop + mContentView.getMeasuredHeight());
             } catch (Exception e) {
-                Log.e("error", "e=" + e);
+                // 吞掉能避免当场崩，但异常从 RecyclerView 的 layout 里逃逸会让它的
+                // mLayoutOrScrollCounter 永久 >0，之后任意 notifyItem* 都抛
+                // "Cannot call this method while RecyclerView is computing a layout"，
+                // 崩在毫不相干的地方。所以这里必须把真正的现场上报出去。
+                if (BuildConfig.DEBUG) Log.e("SwipeBackLayout", "content layout failed", e);
+                CrashReport.postCatchedException(e);
             }
         }
         mInLayout = false;
--- a/wkuikit/src/main/java/com/chat/uikit/chat/ChatActivity.java
+++ b/wkuikit/src/main/java/com/chat/uikit/chat/ChatActivity.java
@@ -3050,36 +3050,38 @@ public class ChatActivity extends SwipeBackActivity implements IConversationCont
         }
     }
 
-    // RecyclerView 正在 layout / scroll 时 notify* 会抛 IllegalStateException，只能延后一帧。
-    // 已有动作在排队时后续动作也必须排队，否则新回调会插到旧回调前面，
-    // 同一条消息的两次刷新会被旧状态覆盖。
-    // 若某次 layout 抛异常被上层吞掉，RecyclerView 会永久停在 computing 态，
-    // 重试多少次都没用——这时宁可丢掉这次刷新也不崩。
+    // RecyclerView 正在 layout / scroll 时 notify* 会抛 IllegalStateException，只能延后。
+    // 排队与重试规则见 SafeAdapterActionQueue —— 关键是队头被挡住时重排的是同一个队头，
+    // 后到的动作不会插队。
     private static final int MAX_SAFE_ADAPTER_RETRY = 3;
-    private int pendingAdapterActions = 0;
 
+    private final SafeAdapterActionQueue adapterActionQueue = new SafeAdapterActionQueue(
+            new SafeAdapterActionQueue.Host() {
+                @Override
+                public boolean isBusy() {
+                    return wkVBinding.recyclerView.isComputingLayout();
+                }
+
+                @Override
+                public void postDrain(Runnable drain) {
+                    wkVBinding.recyclerView.post(drain);
+                }
+
+                @Override
+                public void onDropped(int count) {
+                    if (BuildConfig.DEBUG) {
+                        Log.w("MsgDebug", "[safeAdapterAction] recyclerView stuck in computing state, drop "
+                                + count + " pending refresh");
+                    }
+                }
+            }, MAX_SAFE_ADAPTER_RETRY);
+
+    /** 只能在主线程调用 —— {@link SafeAdapterActionQueue} 的状态全是裸字段，无同步。 */
     private void safeAdapterAction(Runnable action) {
-        if (pendingAdapterActions == 0 && !wkVBinding.recyclerView.isComputingLayout()) {
-            action.run();
-            return;
+        if (BuildConfig.DEBUG && Looper.myLooper() != Looper.getMainLooper()) {
+            Log.e("MsgDebug", "[safeAdapterAction] called off main thread", new IllegalStateException());
         }
-        postAdapterAction(action, 0);
-    }
-
-    private void postAdapterAction(Runnable action, int retryCount) {
-        if (retryCount >= MAX_SAFE_ADAPTER_RETRY) {
-            if (BuildConfig.DEBUG) Log.w("MsgDebug", "[safeAdapterAction] recyclerView stuck in computing state, drop refresh");
-            return;
-        }
-        pendingAdapterActions++;
-        wkVBinding.recyclerView.post(() -> {
-            pendingAdapterActions--;
-            if (wkVBinding.recyclerView.isComputingLayout()) {
-                postAdapterAction(action, retryCount + 1);
-            } else {
-                action.run();
-            }
-        });
+        adapterActionQueue.submit(action);
     }
 
     // 显示一条时间消息
--- a/wkuikit/src/main/java/com/chat/uikit/chat/ChatActivity.java
+++ b/wkuikit/src/main/java/com/chat/uikit/chat/ChatActivity.java
@@ -788,6 +788,7 @@ public class ChatActivity extends SwipeBackActivity implements IConversationCont
         wkVBinding.recyclerView.setLayoutManager(linearLayoutManager);
         wkVBinding.recyclerView.setAdapter(chatAdapter);
         //  perf: MyItemAnimator 需显式关 change 动画，避免 notify 刷新与 fling 叠加掉帧。
+        // 现在由 MyItemAnimator 构造函数里的 setSupportsChangeAnimations(false) 保证。
         wkVBinding.recyclerView.setItemAnimator(new MyItemAnimator());
         chatAdapter.setAnimationFirstOnly(true);
         chatAdapter.setAnimationEnable(false);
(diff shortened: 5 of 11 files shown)


Jerry-Xin (changes requested):
The PR is relevant and addresses valid crash paths, but the new adapter queue does not preserve FIFO ordering as the code comment claims.

## 🔴 Blocking

- 🔴 Critical — Retry reorders queued adapter operations
  `wkuikit/src/main/java/com/chat/uikit/chat/ChatActivity.java` (`postAdapterAction`, ~L3069)

  Once `pendingAdapterActions > 0`, every action is enqueued via `recyclerView.post`, which is FIFO — good. But when an action's posted runnable fires and `isComputingLayout()` is *still* true, it re-posts **itself to the tail** of the queue (`postAdapterAction(action, retryCount + 1)`), letting a later-queued action run ahead of it:

  1. A then B are queued in order → handler queue: [lambdaA, lambdaB].
  2. lambdaA fires, layout still computing → re-posts A to the tail → queue: [lambdaB, lambdaA'].
  3. lambdaB runs first → B executes before A.

  This is exactly the "new callback jumps ahead of the old one" hazard the added comment says it prevents. Reordering a `removeMsg`/`removeAt`/`notifyItemRemoved` ahead of an earlier insert/refresh for the same item can reintroduce inconsistent adapter state (stale index → `notifyItem*` mismatch / IndexOutOfBounds). The retry-on-still-computing path is uncommon (layout usually settles within a frame), but when hit it undermines the invariant the fix is built on.

  Fix: use a single explicit FIFO queue (e.g. `ArrayDeque<Runnable>`); when the head is blocked by `isComputingLayout()`, re-post a drain pass that retries the **same head** without letting later actions run before it. Bounded-retry-then-drop is fine, but the dropped/retried action must stay at the front.

## 💬 Non-blocking

- 🟡 `pendingAdapterActions` is a plain `int` with no synchronization. It is safe only because all IM/network callbacks in this class marshal to the UI thread and `recyclerView.post` runnables also run on the UI thread. Worth a short comment asserting main-thread-only, or it becomes a latent race if any caller is moved off the main thread.
- 🟡 Add focused tests for FIFO ordering, retry exhaustion (drop-not-crash), and actions arriving during a retry. The current manual test plan would not reliably expose this ordering defect.

## ✅ Highlights

- Contact list is now initialized at declaration (never null), closing the search-before-async NPE across `searchUser`, `buildAndDisplayList`, and `rightButtonClick`; entity/channel null-guards and `searchKey` preservation are solid UX bonuses.
- `WKDbScheduler.submit()` is a clean single choke point: the guard wraps the task before it reaches RxJava, narrowly drops only "connection pool has been closed" `IllegalStateException` (cause-chain walk is cycle-safe), and re-throws everything else. All 19 call sites migrated; zero `get().scheduleDirect` remain.
- Disabling change animations via the constructor removes the dual-holder lifecycle path that hit the dirty `tmpDetached` recycle crash; the `SwipeBackLayout` catch now reports the real site via `postCatchedException` instead of silently swallowing.

mzanalyticst:
感谢 review，🔴 Blocking 的那条确实是真缺陷，已在 d785579 修复。

## 🔴 重试乱序

复核确认了你描述的时序完全成立：队头被挡住时旧实现把它 `post` 到**队尾**，后入队的动作因此插队。这正是那段注释声称防住的事 —— 注释写对了，实现没跟上。

按你的建议改成显式 FIFO 队列 `SafeAdapterActionQueue`：队头被挡住时重排的是**同一个队头**，后到的动作只能追加到队尾；重试到上限整队丢弃（此时 RecyclerView 已永久停在 computing 态，留着也跑不了）而不是崩。

## 🟡 主线程约束

队列的线程约束写进类注释（「非线程安全，只能在主线程使用」，并说明依赖的前提），`safeAdapterAction` 里补了 debug 门控的非主线程调用告警。

## 🟡 测试

队列抽成独立类后，`Host` 接口把 `isBusy` / `postDrain` 抽走，可以脱离 Android 用纯 JUnit 测。新增 9 条：

| 用例 | 覆盖 |
|---|---|
| `队头被挡住时重排的是同一个队头_后到的动作不会插队` | **本次缺陷的回归测试** |
| `重试期间新提交的动作排在原有动作之后` | 重试窗口内到达 |
| `重试耗尽_丢弃整个队列且不抛异常` | drop-not-crash |
| `丢弃之后队列仍然可用` | 重试计数复位 |
| `排空过程中动作内部再提交_追加到队尾而不是立刻执行` | 递归提交 |
| `排空途中重新变忙_剩余动作保持顺序留到下一帧` | 动作触发 layout |
| `动作抛异常_剩余队列不会烂在里面` | 异常穿出后仍安排排空 |
| `空闲且无排队时同步执行` / `忙时入队_排空后按提交顺序执行` | 基本语义 |

**回归测试有效性已验证**：把实现临时改回旧行为（队头重排到队尾）后重跑，只有 `队头被挡住时重排的是同一个队头_后到的动作不会插队` 失败，其余 8 条照过；恢复后 9/9 通过。

`assembleDebug` + `:wkuikit:testDebugUnitTest` + `:wkbase:testDebugUnitTest` 全部通过。

Jerry-Xin (approved):
Re-review of the updated head `d7855796448465ff6c33df0733edc6a5184bfb34`. The prior blocking finding (retry re-ordering in `safeAdapterAction`) is now resolved. APPROVE.

## ✅ Blocking finding from prior review — FIXED

The FIFO retry-overtake hazard is properly fixed. The inline `pendingAdapterActions` int + self-reposting runnable was replaced by an explicit single serial queue `SafeAdapterActionQueue` (`wkuikit/src/main/java/com/chat/uikit/chat/SafeAdapterActionQueue.java`):

- **Retry re-runs the head, not the individual runnable.** When `drain()` finds the RecyclerView still busy, it re-schedules the *same* drain (increments `retryCount`) instead of re-posting the head runnable to the tail. Queue order (`ArrayDeque` FIFO) is never disturbed, so a later-enqueued action B can no longer overtake an earlier action A.
- **Drain pulls from the head** (`pollFirst`) in FIFO order, re-checking `isBusy()` before each action.
- **Re-entrant submit guard:** a `draining` flag forces an action that itself calls `submit()` to enqueue at the tail rather than run inline, closing the recursive-overtake gap.
- **Retry exhaustion drops the whole queue** (not one action) and resets state — same "drop rather than crash" intent, now consistent.

## ✅ Non-blocking items from prior review — addressed

- The unsynchronized `pendingAdapterActions` int is gone; the queue is documented main-thread-only and `safeAdapterAction` now asserts off-main-thread calls in debug (`Looper` guard).
- Deterministic tests added (`SafeAdapterActionQueueTest.java`) covering: sync fast-path, FIFO after drain, the exact reorder regression, retry-window ordering, retry exhaustion drop, post-drop reuse, nested re-submit, re-busy mid-drain, and action-throws re-drain.

## ✅ No regression on previously-approved fixes

The NPE fix (contact search `allList` init) and DB-scheduler fix (`WKDbScheduler.submit`) are byte-unchanged from the prior head (delta is only `ChatActivity.java` + the two new queue files; compare is ahead 1 / behind 0, not a rebase).

## 💬 Non-blocking notes (all pre-existing / minor)

- 🟡 `WKDbScheduler` (`wkbase/src/main/java/com/chat/base/utils/WKDbScheduler.java:100`): the pool-closed guard uses `contains()` on the exception message, which could suppress an unrelated `IllegalStateException` whose text happens to include the SQLCipher string. Prefer exact-message matching. Consider unit tests for the guard (direct/nested pool-closed plus unrelated exceptions that must propagate).
- 🔵 `ChooseContactsActivity` (`wkuikit/src/main/java/com/chat/uikit/contacts/ChooseContactsActivity.java:529`): pinyin comparison is `key.contains(entity.pying...)`; typical incremental search is `entity.pying.contains(key)`. Predates this PR — worth a separate fix.

## ✅ Highlights

- `SafeAdapterActionQueue` preserves FIFO ordering, bounds retries, and has strong deterministic coverage.
- Adapter mutations are deferred together with their notifications.
- Contact loading handles null results, destroyed activities, and pre-data searches.
- The DB guard narrowly targets closed-pool failures while letting other exceptions surface.

Note: focused tests could not be executed locally (no Java runtime in the review env); the supplied queue tests plus the reported `assembleDebug` result provide reasonable coverage.

yujiawei (approved):
# Code Review — PR #129 (octo-android)

Reviewed at head `d785579` against merge-base `35c3aac` (= current `main`, so the diff is up to date). This is a re-review after the round-1 ordering fix; I re-derived the verdict from scratch rather than inheriting the earlier one.

---

## 1. Spec compliance — ✅

Mapped against linked issue #128 (three crash classes) plus the round-1 follow-up.

| Required | Status | Evidence |
|---|---|---|
| (1) `ChooseContactsActivity.searchUser` NPE — `allList` never null | ✅ | `ChooseContactsActivity.java:88` initialized at declaration; covers `searchUser`, `buildAndDisplayList`, `rightButtonClick` |
| (1b) Don't swallow the user's in-flight search term | ✅ | `searchKey` tracked at `:516`, replayed at `:507-511` |
| (2a) Stop `notifyItem*` escaping into a computing layout | ✅ | `SafeAdapterActionQueue` + 14 call sites in `ChatActivity` (5 pre-existing, 9 newly covered) |
| (2b) Kill the dual-holder change path that hits dirty `tmpDetached` | ✅ | `MyItemAnimator.java:54-65`; with `mSupportsChangeAnimations=false`, the overridden `canReuseUpdatedViewHolder` (`:677-680`) is now unconditionally `true`, so the hidden-view recycle path is gone |
| (2c) Report the swallowed `onLayout` exception instead of hiding it | ✅ | `SwipeBackLayout.java:411-416` |
| (3) Guard DB tasks against a closed connection pool | ✅ | `WKDbScheduler.submit()`; guard wraps the task **before** it reaches RxJava, which is the part that actually matters given `ScheduledDirectTask` rethrows into the UEH |
| (3b) Migrate all call sites | ✅ | Verified: `grep -rn "WKDbScheduler.get()"` returns **zero** hits repo-wide; `get()` is `@Deprecated` |
| Round-1: FIFO must re-post the *same head*, not the tail | ✅ | `SafeAdapterActionQueue.drain()` re-schedules a drain pass instead of re-posting the action (`:89-113`); regression test `队头被挡住时重排的是同一个队头_后到的动作不会插队` fails against the old behaviour, as the author demonstrated |

**Missing:** none.

**Out of scope / extra:** one item, non-blocking —

- `WKDbScheduler.submit(Runnable, long, TimeUnit)` (`WKDbScheduler.java:74`) has **zero callers**. The delayed revoke retry uses `Handler.postDelayed` + the non-delayed `submit`, so the overload is dead code today. Either drop it or keep it deliberately as API surface.

**Deviation:** none material. One undocumented behaviour change is noted under P2-4 below.

I also confirmed CI runs `./gradlew testDebugUnitTest assembleDevDebug` (`.github/workflows/ci.yml:92`) and it is green on this head, so the 9 new unit tests and the build claim are machine-verified.

---

## 2. Code quality — Approved (no P0/P1)

The core state machine holds up. I walked `submit` / `scheduleDrain` / `drain` for reentrancy and could not construct a reordering, double-execution, or non-draining case:

- `drainScheduled == true` always implies a non-empty queue (it is only set from `submit` after `addLast`, and from `drain` on a non-empty queue), so the synchronous fast path at `:73-76` can never jump ahead of a pending drain.
- `draining` correctly forces re-entrant submits (an action calling `safeAdapterAction`) to the tail.
- The `finally` at `:108-112` reschedules on both the busy-interrupt and the action-threw paths, so the queue cannot rot.
- `animateChange(old == new)` delegating to `animateMove`, which early-returns on zero delta (`MyItemAnimator.java`), matches `DefaultItemAnimator`'s own contract, so no stuck animation bookkeeping.

Everything below is **non-blocking**. The common theme of P2-1..P2-3 is the same root: adapter state is read/mutated *outside* the serialization boundary while the corresponding notify is *inside* it.

### 🟡 P2-1 — `receivedMessages` splits a data mutation from its notification across the queue boundary

`ChatActivity.java:4377` removes the typing row from the backing list **synchronously**, on the IM callback:

```java
chatAdapter.getData().remove(chatAdapter.getItemCount() - 1);
typingRemoved = true;
```

while the matching notify lives inside the queued lambda at `:4450`:

```java
if (finalTypingRemoved) {
    chatAdapter.notifyItemRemoved(chatAdapter.getItemCount());
}
```

Two consequences:

1. Between the two, the adapter's `getItemCount()` and RecyclerView's internal item count disagree. Any layout pass in that window yields `IndexOutOfBoundsException: Inconsistency detected`.
2. Worse, the new bounded-retry **drop** path (`SafeAdapterActionQueue.java:96-101`) discards the queue wholesale. The data mutation already happened; the notify is gone forever, so the desync becomes permanent rather than one-frame.

This split predates the PR (the `safeAdapterAction` wrapper at this site already existed on `main`), and the drop only fires when `isComputingLayout()` is stuck across three drains — a state in which the old code crashed with `IllegalStateException` anyway, so this is not a regression. But the drop path *is* new, and it converts a loud crash into silent state corruption. Suggested fix is one line of motion: move the inspection + `getData().remove(...)` at `:4372-4381` **into** the queued lambda so data and notify are always dropped or applied together.

### 🟡 P2-2 — Same split can strand the typing row permanently

Consequence of the same boundary problem, worth calling out separately because it is user-visible without any crash:

1. A typing event queues `doTyping` (`:4544`).
2. Before the drain, a real message arrives. `receivedMessages` inspects the adapter at `:4374` outside the queue, sees no typing row (it hasn't been added yet), and captures `finalTypingRemoved = false`.
3. FIFO then runs `doTyping` (appends typing) and *then* the message insert, leaving the typing row **not last**.
4. `removeTypingItem` (`:4602-4611`) only removes the last item if it is a typing row, so the row can never be cleared, and `lastMsgIsTyping()`-based index math downstream is off by one.

Note this window is genuinely reachable, not merely theoretical: the SDK's `runOnMainThread` (`wkim/.../manager/BaseManager.java:18-26`) executes **inline** when already on the main thread, so a listener fired from inside a bind/layout pass really does hit `isComputingLayout() == true` and really does enqueue. Same fix as P2-1.

### 🟡 P2-3 — The index-0 synthetic row is mutated both inside and outside the queue

`getChannelState` now routes its spanEmptyView add/remove through the queue (`:2180-2190`, `:2199-2206`), but the two endpoint handlers that mutate the *same* row still do it directly:

- `hide_pinned_view` → `getData().remove(0)` + `notifyItemRemoved(0)` (`:1092-1093`)
- `show_pinned_view` → `addData(0, ...)` (`:1113-1115`)

Both are reachable while a queued index-0 action is pending, so the two can invert. Both sides re-check `type == spanEmptyView` at index 0, so the realistic worst case is a duplicated or missing spacer row rather than a crash — hence P2 — but a partially-adopted serialization boundary is exactly the kind of invariant that decays. Either route these two through `safeAdapterAction` as well, or add a comment stating why they are exempt.

### 🟡 P2-4 — Pinyin matching semantics changed silently, and the underlying comparison is inverted

`ChooseContactsActivity.java:530`:

```java
|| (!TextUtils.isEmpty(entity.pying) && key.contains(entity.pying.toLowerCase(...)))
```

Two things:

- The new `!TextUtils.isEmpty(entity.pying)` guard changes behaviour: previously `content.contains("")` was `true`, so any contact with an empty pinyin matched **every** query. That was almost certainly a bug and removing it is an improvement — but it isn't mentioned in the PR description or the issue, and it is a user-visible search change. Worth one line in the description.
- Pre-existing and untouched: the comparison is backwards. It asks whether the *query* contains the *contact's* pinyin, so typing `zh` will not match `zhangsan`; only typing the full pinyin (or a superstring of it) matches. If pinyin search is meant to work at all, this wants `entity.pying.toLowerCase(...).startsWith(key)`. Out of scope for a crash-fix PR, but you are already editing this exact expression.

### 🟡 P2-5 — Unbounded catched-exception reporting from `onLayout`

`SwipeBackLayout.java:416` calls `CrashReport.postCatchedException(e)` on every failing layout pass, with no dedup or cooldown. Getting the real stack is the whole point of the change and I agree with it — but if the duplicate-`addView` state is persistent rather than one-shot, this fires on every traversal, doing synchronous work on the main thread in exactly the situation where the UI is already struggling. A one-shot `AtomicBoolean` per instance, or a per-signature cooldown, keeps the diagnostic value without the storm risk.

### 🟡 P2-6 — Retry exhaustion is invisible in release builds

`onDropped` (`ChatActivity.java:3068-3074`) is a no-op unless `BuildConfig.DEBUG`. Since the drop only happens in the permanently-stuck state you are trying to characterize, this is precisely the signal you want from the field. Consider a `CrashReport.postCatchedException` here too (rate-limited, per P2-5), so a dropped queue shows up in telemetry rather than as an unexplained frozen list.

### Nits

- `WKDbScheduler.isPoolClosed` (`:98-105`): the `if (t.getCause() == t) break;` guard only catches a self-referential cause. A longer cycle (A → B → A) still loops forever. A visited-set or a bounded hop count (say 16) is cheaper to reason about.
- The guard swallows the pool-closed exception and lets the rest of the `Runnable` be skipped silently. Fine for today's 19 call sites (all are single DB operations), but if a future task interleaves DB work with in-memory state updates, those updates vanish on logout. The class javadoc is a good place to state that only whole-task DB work belongs in `submit`.
- The guard is narrow by design (`IllegalStateException` containing `connection pool has been closed`), which I agree with. Note it does not cover `attempt to re-open an already-closed object`, which some post-close paths raise; if that shows up in Bugly it needs its own decision rather than a widened matcher.
- `safeAdapterAction` (`ChatActivity.java:3080-3084`) only *logs* off-main-thread callers in DEBUG. That is the right call — I verified every listener feeding these sites dispatches on the main thread (`MsgManager.setDeleteMsg`/`setRefreshMsg`/`pushNewMsg` all use `runOnMainThread`; `ChannelManager.setRefreshChannel`; `ChannelMembersManager.setRefreshChannelMember`; network callbacks via `WKBaseModel` `observeOn(AndroidSchedulers.mainThread())`). Since `SafeAdapterActionQueue`'s state is entirely unsynchronized, a `Looper` check that *marshals* rather than logs would make the invariant self-enforcing at negligible cost.

---

## 3. Verdict

**APPROVE.**

Spec is fully met, the round-1 ordering defect is genuinely fixed (and the regression test was shown to fail against the old implementation, which is the part most people skip), the migration is complete and verifiable, and CI is green including the new unit tests. Everything I found is a hardening item with a narrow trigger, and none of it is a regression relative to `main` — the drop path trades a loud crash for a quiet one in a state that was already unrecoverable.

I'd like P2-1/P2-2 picked up soon since they share a one-line-of-motion fix, but they should not hold this batch.

---

## 4. Suggested follow-ups, in priority order

1. Move the typing-row inspection and `getData().remove(...)` at `ChatActivity.java:4372-4381` inside the queued lambda, so every queued action is atomic over (data mutation, notify). Fixes P2-1 and P2-2 together.
2. Route `show_pinned_view` / `hide_pinned_view` index-0 mutations through `safeAdapterAction`, or document the exemption (P2-3).
3. Rate-limit `postCatchedException` in `SwipeBackLayout`, and add release-build reporting to `onDropped` (P2-5, P2-6).
4. Delete or justify the unused `WKDbScheduler.submit(task, delay, unit)` overload.
5. Mention the pinyin empty-match behaviour change in the description; consider fixing the inverted `contains` separately (P2-4).

---

## 5. Additional observations (outside this PR's scope)

- `ChatActivity.removeMsg` (`:2790`): `if (chatAdapter.getData().size() >= timeIndex)` should be `>`, and `getData().get(timeIndex).wkMsg.type` is dereferenced without the null guard this PR adds to sibling loops at `:1668` and `:2181`. Hard to reach today, but it is the same class of defect the PR is fixing elsewhere.
- `ChatActivity.java:1143`: the `reset_channel_all_pinned_msg` loop dereferences `getData().get(i).wkMsg.remoteExtra` with no null guard, unlike its newly-hardened siblings. Worth a consistency pass.
- `refreshMsg` (`:4728-4739`) schedules a `postDelayed(..., 200)` that closes over a captured index. The index can shift within 200 ms; the bounds checks prevent a crash but the wrong rows can be refreshed.
- `WKIMUtils.scheduleRevokeRetry` (`:703-715`): if the user logs out and back in within the 500/1500 ms window, the pool is no longer closed (a *new* DB is open), so the guard correctly does not fire — but the retry then runs against the new account's database with the old account's `channelID`. It only triggers a `syncExtraMsg` request rather than a local write, so the blast radius is small, and this is unchanged by the PR. A login-generation token checked inside the task would close it properly.
- Sibling open PRs (#85, #96, #110, #113, #114, #120, #121) do not touch `ChatActivity`, `MyItemAnimator`, `WKDbScheduler`, or `SwipeBackLayout`, so no post-merge semantic conflict is expected on this batch.

PR #93, opened 2026-07-15
fix(emoji): 补齐内置自定义表情 [尚方宝剑]

mzanalyticst:
## Summary

octo-server 已下发 `[尚方宝剑]` token（web PR Mininglamp-OSS/octo-web#492 已适配），Android 侧 `emoji.xml` 缺 `custom_shangfang` 条目导致：

- 用户收到 `[尚方宝剑]` 消息时看到裸 `[xxx]` 文本
- 表情面板 custom_ 分类里少一条

## 改动

- 复用 octo-web 同款 PNG（512×512, 324 KB）到 `wkbase/src/main/assets/emoji/default/custom_shangfang.png`
- `wkbase/src/main/assets/emoji/emoji.xml` 补一行 `Emoticon` 条目
- `EmojiManager` 走现有 xml → assets 加载路径，**零代码改动**

Refs #81

## 说明

Issue #81 完整验收标准还有 3 条（服务端 manifest 拉取 / 缓存 / URL sanitize），涉及 `WKCommonService` + `EmojiManager` 中型改造。**本 PR 只覆盖"内置显示 [尚方宝剑]"这一条**——真实生产阻塞的部分——用最小改动降风险。剩余 Layer 2（manifest 动态化 + CDN 支持）计划作为独立 PR 处理，故此处用 `Refs #81` 而非 `Closes`。

## Test plan

- [ ] 编译通过（`./gradlew :wkbase:assembleDebug` 已本地验证）
- [ ] 表情面板 → 自定义分类 → 看到 `[尚方宝剑]` 图标（在既有 `[使命必达]` `[崇尚行动]` `[有品位]` 之后）
- [ ] 发送 `[尚方宝剑]` 或收到含该 token 的消息 → 渲染为表情图（非裸文本）
- [ ] 现有其它 custom / Unicode 表情不受影响（回归）

PR #139, opened 2026-08-26
feat(msg-effect): 新增 [尚方宝剑] 全屏视频特效

buyan749:
## What

Adds the missing full-screen video effect for the built-in custom emoji `[尚方宝剑]`.

## Why

`[尚方宝剑]` was added to the emoji picker in #93 (emoji.xml + PNG only). The effect
lookup table in `MessageEffectType.detect()` was last touched in #67, three weeks
earlier, so the emoji never had an effect registered — `detect()` returned null and
nothing played. The other three custom emoji (`[崇尚行动]`, `[有品位]`, `[使命必达]`)
all have effects.

iOS has had this effect since its introduction (`WKShangfangVideoEffect.m`,
mapped in `WKMessageEffectManager.m:72`). This PR brings Android to parity.

## Changes

**Asset** — `wkbase/src/main/assets/effects/shangfang_celebrate.mp4`
Transcoded from the iOS source (HEVC 1764×3840, 2.6 MB) down to HEVC 884×1920,
1.8 MB, matching the existing Android convention (`action_celebrate` and
`classy_celebrate` are both downscaled from their iOS originals).

**`LumaKeyParams` (new)** — per-effect keying parameters.
Android's port had these hardcoded as shared constants, so a new asset could not be
tuned independently. Defaults reproduce the previous constants exactly, so
`action` / `classy` behaviour is unchanged. `LumaKeyParams.SHANGFANG` mirrors
`WKShangfangVideoEffect.m:48-67` field by field.

**`LumaKeyVideoView`** — three capabilities ported from iOS `WKLumaKeyVideoView`:
- configurable center-protect origin (was hardcoded to frame center; iOS uses 0.54/0.41)
- eye-protect circle (second, smaller protect region — did not exist on Android)
- delayed center-disc fade-in (`centerProtectStartTime` / `RampDuration`)

The last one is not cosmetic: iOS documents that without it, the protect disc sits
over still-black background during the first ~600 ms and shows up as a floating
black circle in light mode.

**Coordinate systems differ and are handled**: iOS flips Y for CoreImage's
bottom-left origin; Android `Bitmap` is already top-left, so no flip.

## Incidental fixes

These were pre-existing defects in the shared luma-key path, surfaced while porting:

- `buildCenterMask()`'s return value was discarded and `centerMask` was a `by lazy`
  val, so the mask could never be rebuilt on size change. Harmless while the process
  size was constant; a blocker once it became dynamic.
- The CPU keying buffer was hardcoded to 540×960 (aspect 0.5625) regardless of the
  source aspect. Frames were resampled non-uniformly in and out, turning the circular
  protect mask into a ~1.22:1 vertical ellipse on screen.
- The keyed bitmap was stretched to fill the whole view. Now it is drawn width-filled,
  height by source aspect, vertically centered — matching `WKClassyVideoEffect.m:37-40`.
- Bitmaps are no longer recycled on size change while `onDraw` may still hold them.

## Reviewer note

The last two items change how **`[崇尚行动]` and `[有品位]` render as well** — they are
now aspect-correct rather than stretched, and keyed at a higher resolution. Please
regression-test those two alongside the new effect.

Keying resolution is capped by `LumaKeyParams.processMaxLongSide` (default 1280).
Android keys on the CPU per-pixel, unlike iOS which uses Metal at native resolution,
so this is a sharpness/throughput tradeoff. Dropped frames are skipped by the
existing `processing` latch and do not stall the UI.

## Testing

- `./gradlew :app:assembleDebug` — passes
- `./gradlew :wkbase:testDebugUnitTest` — passes
- Manual: send `[尚方宝剑]` in a chat


Jerry-Xin (dismissed):
## Summary

Adds the missing full-screen luma-key video effect for `[尚方宝剑]`, bringing Android to parity with iOS (`WKShangfangVideoEffect`), plus disclosed fixes to the shared luma-key path. Reviewed at b9e12e4a. No blocking issues found.

## What I verified (byte-level)

- `MessageEffectType.detect()` uses the same exact-match gate as the other two video effects, and the tag matches `wkbase/src/main/assets/emoji/emoji.xml` (`custom_shangfang`, added in #93); the emoji PNG and the new `shangfang_celebrate.mp4` asset are both present.
- All exhaustive `when` branches over the sealed class are updated; the single `LumaKeyVideoEffectPlayer.play()` caller passes the per-effect params.
- Threading: `processFrame` runs on the UI thread via `TextureView.SurfaceTextureListener` (same as the pre-existing `getBitmap` call), so the new `MediaPlayer.currentPosition` read there is safe; per-frame `params` visibility to the worker is established by the handler post.
- `LumaKeyParams.DEFAULT` reproduces the previous hardcoded constants field-for-field, so Action/Classy keying math is unchanged apart from the disclosed aspect/resolution fixes.
- `view.params` is assigned before `startVideo`, satisfying the mask-precompute contract; the 6s watchdog timeout (≈3.5s video + margin) matches the ActionVideo pattern.

💬 Non-blocking

- 🟡 `release()` may recycle bitmaps while a worker frame is still queued; failures are caught and the completion swap degrades to nulls, so it is safe today, but capturing frame-local refs or draining the worker first would be cleaner.
- 🟡 `buildRadialMask` allocates a full-size zero `FloatArray` (~3 MB at 589×1280) for the disabled eye mask on Action/Classy; returning `null` for disabled masks would trim peak memory.
- 🔵 In `LumaKeyVideoEffectPlayer.play`, the local `val params = FrameLayout.LayoutParams(...)` shadows the `params: LumaKeyParams` argument — correct today because the assignment precedes the declaration, but renaming would avoid confusion.
- 🔵 Consider unit tests for the new exact-match detect branch (`[尚方宝剑]`: exact, surrounding whitespace, embedded-token rejection).

✅ Highlights

- Aspect-correct processing buffer + draw rect fix the ellipse-mask and stretch defects in the shared path, explicitly disclosed for Action/Classy regression testing.
- Per-effect `LumaKeyParams` with backward-compatible defaults is a clean refactor.
- Mask rebuild on size change correctly replaces the old discarded-`buildCenterMask()` bug; bitmaps are no longer recycled while `onDraw` may hold them.

Note: the `check-sprint` CI failure is the project-board Sprint-field meta check (also failing on other PRs, not code-related); the code CI workflows are pending first-run approval for this fork PR.

yujiawei (changes requested):
# Code Review — PR #139 (octo-android)

Reviewed at head `b9e12e4a`, against merge-base `3133fe30`. I read the full diff, the whole post-change `msgeffect/video` package, and cross-checked every parity claim against the iOS sources in `octo-ios` (`WKLumaKeyVideoView.m/.h`, `WKShangfangVideoEffect.m`, `WKClassyVideoEffect.m`, `WKActionVideoEffect.m`). I also parsed the mp4 boxes of all three assets to get real dimensions/durations rather than trusting the description.

**What holds up.** Most of this PR is careful, verifiable work:

- `LumaKeyParams.SHANGFANG` (`LumaKeyParams.kt:95-110`) matches `WKShangfangVideoEffect.m:48-67` on all twelve fields — threshold/tolerance, bg floor/ceil, center radius 0.30, softness 0.12, center (0.54, 0.41), eye (0.60, 0.42), eye radius 0.13, eye softness 0.05, start 0.6s, ramp 0.65s.
- `centerStrengthAt()` (`LumaKeyVideoView.kt:261-268`) reproduces `WKLumaKeyVideoView.m:277-286` exactly, including the `rampDuration <= 0` branch; `eyeStrength` (`LumaKeyVideoView.kt:209`) matches `.m:288`.
- The per-pixel alpha math matches the CIColorKernel at `.m:126-144` term for term, including the `max(thr, 0.0001)` / `max(tol, 0.0001)` guards.
- The "no Y flip on Android" claim is right: iOS flips because `destCoord()` is bottom-left origin (`.m:262-271`); `Bitmap` is top-left, so `buildRadialMask` is correct as written.
- The aspect-ratio diagnosis is real. `action_celebrate.mp4` is 1080×2004, `classy_celebrate.mp4` is 1080×2352, `shangfang_celebrate.mp4` is 884×1920 — none are 0.5625, so the old hardcoded 540×960 buffer really was turning the protect disc into an ellipse. iOS keys at native `source.extent` (`.m:293`), so following the source aspect is the right fix.
- The new sealed subclass does not break any exhaustive `when`: `MessageEffectOverlayView.createEffect` is updated and `videoSpecFor` has an `else` branch. Nothing else switches on `MessageEffectType`.
- Trigger semantics match iOS: exact-match on the bracketed tag (`MessageEffectType.kt:41` vs `WKMessageEffectManager.m:105-110`).

One blocking problem, and it is in the shared-path refactor rather than the new effect.

---

## 1. Scope & spec compliance

**Result: ❌**

Nothing is missing — the effect is registered end to end (asset, `detect()`, `videoSpecFor`, overlay branch, params). The problems are on the other two axes.

**Divergence from the reference this change cites**

`LumaKeyVideoView.kt:123-135` comments itself as "aligned with iOS `WKClassyVideoEffect` / `WKShangfangVideoEffect`", and the PR body says the new draw rect matches `WKClassyVideoEffect.m:37-40`. But this `onDraw` is the single shared code path for **all three** effects, and iOS does *not* give `ActionVideo` that layout:

```objc
// WKActionVideoEffect.m:31-36
// 全屏铺满（9:16 视频 aspect-fill 填满整个 effectView）。
videoView.frame = effectView.bounds;
```

versus classy/shangfang, which get an explicitly width-fitted, vertically-centred frame. Inside the view, iOS then always aspect-**fills**:

```objc
// WKLumaKeyVideoView.m:316
CGFloat s = MAX(drawableSize.width / sw, drawableSize.height / sh);
```

So for classy/shangfang the two compose to "width-fill, centred" (the frame already has the source aspect, so `MAX` is a no-op) — the Android port is correct there. For action, iOS gives the view the full bounds and `MAX` scales the video up to cover the screen. Applying the classy layout to action is a behaviour change to a shipped effect that contradicts the reference the change claims to follow. Consequences in §2 P1.

**Out-of-scope default change**

`LumaKeyParams.kt:81` sets `processMaxLongSide = 1280` as the shared default, which changes the keying resolution of the two existing effects (details in §2 P2-1). The PR body describes the new params object as "Defaults reproduce the previous constants exactly, so `action` / `classy` behaviour is unchanged". That is true of the six keying constants (0.10 / 0.12 / 0.05 / 0.45 / 0.30 / 0.14 all match the removed `companion object`), but `processMaxLongSide` is a new knob with no previous equivalent — the old value was effectively 540×960 — and the sentence's conclusion is contradicted by the PR's own "Incidental fixes" and "Reviewer note" sections three paragraphs later. Worth correcting in the description, since a reader deciding how much regression testing this needs will stop at that sentence.

---

## 2. Code quality

**Result: Changes Requested**

### P1 — `[崇尚行动]` now renders letterboxed instead of full-bleed

`wkbase/src/main/java/com/chat/base/msgeffect/video/LumaKeyVideoView.kt:130-135`

```kotlin
val aspect = sourceAspect(bmp)
val dstW = width.toFloat()
val dstH = dstW / aspect
val top = (height - dstH) * 0.5f
dstRect.set(0f, top, dstW, top + dstH)
```

`dstH = width / aspect`. Whenever the container is *taller* than the source aspect implies, this leaves transparent bands rather than covering the view. `action_celebrate.mp4` is 1080×2004 (aspect 0.539) — a much squarer asset than the other two. The container is `android.R.id.content` (`ChatActivity.java:811-816`), essentially full screen.

On a 1080×2400 content area: `dstH = 1080 / 0.539 = 2004`, leaving **198 px uncovered at the top and 198 px at the bottom** — about 8% of screen height at each edge. Even after system-bar insets (say a 2200 px content height) it is still ~98 px per edge. Before this PR the bitmap was stretched to fill, so the effect covered the whole screen.

This matters because the keyed frame is not just the subject — the background is deliberately kept as a luma-driven translucent veil (`backgroundAlphaFloor` 0.05 → `backgroundAlphaCeil` 0.45), and `WKActionVideoEffect.m:32-34` describes the asset as "四角近黑 → 底部蓝光晕". That bottom blue glow is the brightest, most opaque part of the frame, and it will now terminate at a hard horizontal line ~200 px above the bottom of the screen with untinted chat content below it. That reads as a rendering bug, not as a design.

Classy and shangfang are unaffected — their assets are ~0.459 aspect, so on a 0.45 screen the bands are ~24 px, and iOS produces the same ~24 px gap. This is specifically an action-effect regression.

The `Reviewer note` asks for `[崇尚行动]` / `[有品位]` to be regression-tested, but `Testing` lists only "Manual: send `[尚方宝剑]` in a chat", so it looks like that test did not happen.

**Suggested fix** — make the layout per-effect rather than global. Either add a mode to `LumaKeyParams` (e.g. `FILL` vs `WIDTH_FIT`, default `FILL` so action keeps its coverage and classy/shangfang opt in), or implement iOS's actual `MAX(w/sw, h/sh)` aspect-fill in `onDraw` and let the caller size the view. The first is closer to how iOS structures it. Either way, please verify all three on a 20:9 device before/after.

### P2-1 — Per-frame cost of the two existing effects goes up 45–70%

`LumaKeyParams.kt:81`, `LumaKeyVideoView.kt:199`, `LumaKeyVideoView.kt:295`

The old path always keyed 540×960 = 518,400 px. With `processMaxLongSide = 1280` and the source aspect now honoured:

| effect | source | new buffer | pixels/frame | vs before |
|---|---|---|---|---|
| action | 1080×2004 | 689×1280 | 881,920 | **+70%** |
| classy | 1080×2352 | 587×1280 | 751,360 | **+45%** |
| shangfang | 884×1920 | 589×1280 | 753,920 | n/a (new) |

Each frame pays that on both the main-thread `TextureView.getBitmap` readback (`:199`) and the scalar Kotlin keying loop (`:295`), plus `getPixels`/`setPixels`. The `processing` latch means this degrades to a lower effect frame rate rather than UI jank, which is the right design — but it is a silent quality/battery regression on two shipped effects for a PR that adds a third. Consider keeping the shared default at the previous effective resolution (long side ~960) and setting `processMaxLongSide = 1280` only on `SHANGFANG`, where the sharper asset actually motivates it.

Minor, same area: `processSize()` allocates a `Pair<Int, Int>` per frame (`:189`, `:247`), and `mediaPlayer.currentPosition` is read per frame (`:203-207`) even for action/classy where the time gate is disabled (`centerProtectStartTimeMs == 0`).

### P2-2 — A disabled protect mask still allocates a full-size zero array

`LumaKeyVideoView.kt:328-332` and `LumaKeyVideoView.kt:350-351`

```kotlin
val mask = FloatArray(w * h)
if (radiusRatio <= 0f) return mask
```

`buildMasks` unconditionally builds both masks, so action/classy — which have `eyeProtectRadius == 0f` and never read `eMask` (`:309` is gated on `es > 0f`) — retain an all-zero `FloatArray(w*h)` for the whole playback. For action that is 882K floats ≈ 3.5 MiB of pure waste.

Total live frame state for action: 2 × ARGB_8888 bitmaps (3.5 MiB each) + `pixelBuffer` (3.5 MiB) + two masks (3.5 MiB each) ≈ **17.6 MiB**, up from ~8.3 MiB before this PR. And because `processFrame` deliberately abandons rather than recycles the fallback buffers on the size transition (`:190-198` — the reasoning for not recycling is sound), peak can transiently be higher while GC catches up. Cheap fix: return `null` from `buildRadialMask` when `radiusRatio <= 0f` and make `eyeMask` genuinely absent.

### P2-3 — Teardown does not fence the in-flight worker frame

`LumaKeyVideoView.kt:211-213`, `LumaKeyVideoView.kt:380-386`

```kotlin
workerThread?.quitSafely()
workerThread = null
workerHandler = null
captureBitmap?.recycle()
```

`quitSafely()` does not join — already-queued work still runs. So `release()` can recycle and null `captureBitmap` while the worker is inside `getPixels`/`setPixels`, or before it dereferences `captureBitmap!!` (the lambda at `:213` captures the *field*, not a local). The resulting `IllegalStateException` / NPE is swallowed by `catch (_: Exception) {}` at `:214`, and the worker then still posts the buffer swap and can fire `onFirstFrame` on an already-released view.

In today's call order this appears not to crash — `LumaKeyVideoEffectPlayer.removeView`/`cleanup` call `release()` before detaching the view, so the trailing `invalidate()` is a no-op — but it is correctness by accident across two files. Capturing the bitmap into a local before `post` and guarding the main-thread continuation with a `released` flag would make it explicit. Not blocking; worth doing while this file is open.

### P2-4 — `pendingAfd` can leak the asset file descriptor (pre-existing)

`LumaKeyVideoView.kt:141-145` and `LumaKeyVideoView.kt:179-182`

`startVideo` stores the `AssetFileDescriptor` and only opens the player if the `SurfaceTexture` already exists. It does not, at the point `LumaKeyVideoEffectPlayer.play` calls it (the view was just added, so the texture is not yet available), so the fd sits in `pendingAfd` until `onSurfaceTextureAvailable`. If the effect is cancelled first (`destroy()` → `cancel()` → `cleanup()` → `release()`), nothing closes it — `release()` never touches `pendingAfd`. The `catch` in `initMediaPlayer` (`:180-182`) also skips `afd.close()`. Not introduced here, but this PR is the natural place to add `try { pendingAfd?.close() } catch (_: Exception) {}` to `release()`.

### P2-5 — `params` is shadowed inside `play()`

`wkbase/src/main/java/com/chat/base/msgeffect/video/LumaKeyVideoEffectPlayer.kt:46` / `:54` / `:58`

```kotlin
fun play(..., params: LumaKeyParams = LumaKeyParams.DEFAULT) {
    ...
    view.params = params                                  // :54 — keying params (correct today)
    ...
    val params = FrameLayout.LayoutParams(...)            // :58 — shadows the parameter
```

This is correct only because `:54` precedes the local declaration. It compiles with a "Name shadowed" warning and is a landmine for anyone reordering these lines. Rename the local to `lp`.

### P2-6 — small accuracy nits

- `MessageEffectManager.kt:148-149`: the comment says action ≈ 3.2s and classy ≈ 5.07s. The actual assets are 2.234s and 2.400s (mvhd). The classy timeout of `8000L` (`:157`) is also well above iOS's `scheduleRemovalAfterDelay:4.5` (`WKClassyVideoEffect.m:77`), while the comment claims the timeouts are "与 iOS ... 一致". Pre-existing, but the block was edited here.
- `MessageEffectType.kt:28`: `ShangfangVideo(4000L)` is shorter than the real 3.5s video + 0.45s fade-in + 0.6s fade-out = 4.55s. `durationMs` is unused for video types (`createEffect` returns `null`, so `BaseEffect.isFinished` never runs), so this is inert — but it is misleading next to a 3.5s asset.
- No unit tests were added. `detect()`, `centerStrengthAt()` and `processSize()` are pure and JVM-testable, and `processSize()` in particular has real logic (cap, aspect preservation, truncation). There is currently zero coverage in this package, so this is a suggestion rather than a requirement — but `centerStrengthAt` boundaries and the `processSize` cap are exactly the kind of thing that silently rots.

---

## 3. Verdict

**CHANGES_REQUESTED** — on P1 alone. The new `[尚方宝剑]` effect itself is well-executed and faithfully ported; the blocker is that the shared-path layout change regresses `[崇尚行动]` from full-bleed to letterboxed, contradicting the iOS reference the change cites. Everything in §2 P2 is advisory and can be handled at your discretion.

---

## 4. Additional observations (outside this PR's scope)

- `MessageEffectType.detect` uses `text.contains("[使命必达]")` for the rocket effect (`MessageEffectType.kt:43`), whereas iOS requires an exact trimmed match for *every* bracketed key (`WKMessageEffectManager.m:105-110`, with an explicit comment that inline tags must not trigger). The three video effects were fixed to exact-match; rocket was not. Pre-existing platform divergence, worth a separate issue.
- `shangfang_celebrate.mp4` is 884×1920, i.e. 884 rather than 882 (half of the iOS 1764). The 0.2% aspect delta is irrelevant since the protect centres are normalised, but 884 is not a multiple of 8 or 16, which relies on the HEVC conformance window being honoured. If any edge artifacts show up on older MediaTek/Qualcomm decoders, an aligned width is the first thing to try.
- `LumaKeyVideoEffectPlayer.play` silently drops a second effect while one is playing (`:48`, `if (isPlaying) return`). Pre-existing and probably intended.

---

## 5. What I could not verify

- **No build or test run.** There is no Android SDK in this environment, so the claimed `:app:assembleDebug` and `:wkbase:testDebugUnitTest` results are taken on trust. My "no broken `when`" conclusion is from reading every `MessageEffectType` reference, not from a compile.
- **No device rendering.** The whole premise of the eye-protect circle and the delayed centre disc is light-mode appearance; I could not confirm the black-circle artifact is actually gone, nor measure real frame rates or heap on a low-end device. The P1 numbers above are computed from asset dimensions and the draw math, not observed.
- **HEVC decode availability** across the device fleet, and vendor differences in `TextureView.getBitmap` scaling behaviour.
- **Rotation / large-screen behaviour** — with a landscape or unfolded container the width-fit formula crops very aggressively (only ~20% of the shangfang frame would be visible). iOS uses the same formula for classy/shangfang, so this may be equally true there; I did not confirm whether chat is orientation-locked.
- **Back-to-back effects** and the collect-phase interaction were not exercised.

Review legs actually run: this write-up, plus two independent advisory passes (one line-level adversarial, one whole-repo). Both independently flagged the action-layout regression as blocking, which is why it is P1. One advisory finding — that the keying loop iterates `pixels.indices` beyond `w * h` on a shrinking buffer — I checked and **rejected**: `LumaKeyVideoView.kt:274-276` reallocates on `pixelBuffer!!.size != size`, an inequality in both directions, so `pixels.size` always equals `w * h`.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #139 (octo-android)
Head: `b9e12e4ac308b3054fba602fd5b8a62c339ecf66` · Base: `main` @ `3133fe30`

## Summary

新增 `[尚方宝剑]` 全屏 luma-key 视频特效，并把原先写死在 `LumaKeyVideoView` 里的抠像常量提成 `LumaKeyParams` 值对象（`DEFAULT` 等价旧常量、`SHANGFANG` 逐项对齐 iOS），同时重写了抠像缓冲尺寸（固定 540×960 → 按素材比例、长边封顶）与绘制布局（拉伸铺满 → 等比宽铺满+垂直居中）。触发链路、sealed-class 穷举、资源与超时配置均完整。未发现 P0/P1 阻塞；三个 P2 均为本次重构对既有 action/classy 特效的回归风险面与一处 pre-existing 竞态，见下。

## Verification

Static analysis only at head `b9e12e4a`; build and tests not executed in this environment.

- ✅ **新特效端到端连线** — `MessageEffectType.detect()` 精确匹配 `[尚方宝剑]`（`MessageEffectType.kt:41`）→ `shouldTrigger` → 收集路径 `fireBestCandidate` 与去抖路径 `playEffect` 均已加分支（`MessageEffectManager.kt:109-111`）→ `videoSpecFor` 返回 `SHANGFANG` 参数（`:158`）→ `player.play(..., params)` 且 `view.params = params` 在 `startVideo` 之前（`LumaKeyVideoEffectPlayer.kt:53-54`）。
- ✅ **sealed-class 穷举完整** — 全部三处 `when`/路由点（manager 路由、`videoSpecFor`、overlay 工厂 `MessageEffectOverlayView.kt:113`）已覆盖 `ShangfangVideo`；全仓 grep 无遗漏消费点，无对已删常量（`PROCESS_WIDTH`/`THRESHOLD`/`drawMatrix` 等）的残留引用。
- ✅ **DEFAULT 参数与旧行为逐位等价** — threshold 0.10 / tolerance 0.12 / floor 0.05 / ceil 0.45 / 中心盘 r=0.30 soft=0.14 @ (0.5,0.5) 与旧常量一致；eye 圈默认关闭（radius 0 → `es=0` 分支跳过）、`centerProtectStartTimeMs=0` → `centerStrengthAt` 恒 1，action/classy 的 luma 数学不变（`LumaKeyParams.kt:30-67`）。
- ✅ **线程模型** — `processFrame` 在 TextureView 帧回调执行，`processing`（`@Volatile`）门闩保证同一时刻只有一帧在途，双缓冲交换只在主线程；`currentPosition` 读取有 try/catch 兜底。
- ✅ **资源与兜底** — `shangfang_celebrate.mp4` 存在（1.8MB）且为合法 MP4（ftyp mp42 + moov）；6s 看门狗超时与 ≈3.5s 视频匹配、与 action 的 3.2s/6s 模式一致；`openFd`/MediaPlayer 出错均走 `cleanup()`，无 view 残留。

## Findings

无 P0/P1；三个 P2 如下（其中一个为 pre-existing）。

### P2 — 默认 `processMaxLongSide=1280` 使既有 action/classy 每帧 CPU 负载上升约 45% (`wkbase/src/main/java/com/chat/base/msgeffect/video/LumaKeyParams.kt:81`)

改造前抠像缓冲固定 540×960（≈0.52M px/帧）；改造后 `processSize()`（`LumaKeyVideoView.kt:239`）按素材比例、长边上限 1280 计算，884×1920 素材得到 589×1280（≈0.75M px/帧），且该默认值经 `LumaKeyParams.DEFAULT` 同样作用于 action/classy（`videoSpecFor` 未给它们单独传参）。`getPixels`/逐像素 luma 循环/`setPixels`/`getBitmap` 开销随像素数线性增长，低端机上既有特效会掉帧。diff-scope：new（本 PR 引入）。`processing` 门闩丢帧保 UI、不卡主线程，且新素材确实需要更高分辨率，故不阻塞；建议低端机实测，必要时给 action/classy 显式传较小的 `processMaxLongSide`（如 960）维持旧负载。

### P2 — 绘制布局重构同时改变既有 action/classy 的显示契约 (`wkbase/src/main/java/com/chat/base/msgeffect/video/LumaKeyVideoView.kt:134`)

绘制从"拉伸铺满"改为"等比宽铺满+垂直居中+溢出裁剪"（对齐 iOS）。这同样作用于既有特效：屏幕宽高比大于素材时裁掉素材上下边缘，小于时上下出现无半透明黑纱覆盖的区域。注释称被裁部分本是抠透明的黑边，但 `bgA` 并非 0（floor 0.05 / ceil 0.45 的纱），裁剪会改变既有特效在部分机型上的可视范围。diff-scope：new（本 PR 有意为之的 iOS parity 修正，特效仍可正常播放）。不阻塞；建议对 action/classy 做一次真机回归（尤其 16:9 与折叠屏内屏）确认主体不被裁切。

### P2 — release() 与在途帧的位图回收竞态（pre-existing） (`wkbase/src/main/java/com/chat/base/msgeffect/video/LumaKeyVideoView.kt:190`)

`release()`（`:383-386`）先 `recycle` capture/display bitmap 并把 `pixelBuffer` 置空，而 worker 的 `applyLumaKey` 可能正在使用它们；外层 try/catch 会吞掉异常，随后已入队的交换 runnable 仍会 `renderView.invalidate()`，若交换发生在 detach 之前、绘制遍历随后执行，则 `onDraw` 对已 recycle 的 `displayBitmap` 绘制会抛 "trying to use a recycled bitmap"。diff-scope：pre-existing——base 分支 recycle/交换结构完全一致，本 PR 仅新增 mask 置空（有 `!= null` 守卫，安全），按定级规则不阻塞本次合并。触发需 `destroy()`/`cancel()` 恰好落在单帧窗口（毫秒级）；建议后续加 `released` 标志：交换 runnable 先检查再 `invalidate`，或 `release()` 先 `workerHandler.removeCallbacksAndMessages(null)`。

- 遮罩重建条件含 `centerMask == null || eyeMask == null`，尺寸变化/release 后都会重建；`buildRadialMask` 对 `radius<=0` 返回全 0、`soft>=0.0001f`，smoothstep 无除零。
- `processSize()` 兜底：`coerceAtLeast(1)` + `maxOf(1,…)`，素材尺寸未知时回退 540×960；`sourceAspect` 同兜底。
- `ShangfangVideo.durationMs=4000L` 仅粒子特效消费（`BaseEffect.isFinished`），视频路径用 `timeoutMs`，与 ActionVideo 一致、无害。
- `currentPosition` 的主线程假设：即使个别 ROM 帧回调不在主线程，try/catch 兜底为 0 仅导致该帧保护盘强度归零（瞬时、不崩溃），可接受。
- `ChatActivity.java` 两处 `onMessageVisible` 调用点与 `destroy()` 生命周期未变。

## Human-verify

1. `SHANGFANG` 的圆心/眼部圈坐标（0.54/0.41、0.60/0.42）按 PR 注释为 iOS 侧按帧实测值，本环境无法核对 iOS 源码与素材画面，需真机视觉确认。**非本 PR 合并阻塞项。**
2. 既有 action/classy 在新绘制布局下的显示回归（尤其非 20:9 机型）。**非本 PR 合并阻塞项。**

## 数据流回溯

- `[尚方宝剑]` 文本 → `msg.baseContentMsgModel.getDisplayContent()` → `detect()` → `ShangfangVideo` → 双路径（收集/去抖）→ `VideoSpec(SHANGFANG)` → `view.params` → `applyLumaKey` 每帧消费 `params.lumaThreshold/lumaTolerance/backgroundAlpha*/` 几何字段：每个消费点上游均有真实赋值，无空值/短路。
- `srcWidth/srcHeight` ← `onPrepared`/`onVideoSizeChanged`（`@Volatile`）→ `processSize()`/`sourceAspect()`：到达前走兜底尺寸，到达后下一帧自动重建缓冲与遮罩，数据真实流到消费点。
- `positionMs` ← `mediaPlayer?.currentPosition`（主线程、try/catch）→ `centerStrengthAt` → 逐帧标量 `cs`：600ms 前为 0、650ms 内线性到 1，与门控设计一致。
- 帧数据 → `TextureView.getBitmap(captureBitmap)` → worker `applyLumaKey` → 交换为 `displayBitmap` → `onDraw(dstRect)`：双缓冲 + `processing` 门闩下无并发读写同一 bitmap 的路径。
- 遮罩几何 ← `buildMasks(params.…)` 于处理尺寸首帧预计算：`params` 在 `startVideo` 前设置（`LumaKeyVideoEffectPlayer.kt:53-54`），happens-before 经 Handler post 建立，worker 可见。

## 盲点 checklist（C1–C6）

- **C1 双路径 parity — hit/clear**：新特效类型在 4 个消费点对称添加（detect、fireBestCandidate 路由、videoSpecFor、overlay 工厂），grep 证实无遗漏；`LumaKeyVideoView` create↔release 对称（mask 建于 `buildMasks`、release 置空并验证重建条件覆盖）。
- **C2 control-flow ordering / 复用 — clear**：共享的 `play()` 路径中 `params` 先于 `startVideo` 设置；DEFAULT 逐位等价证明（见 Verification）表明 action/classy 无二次作用/顺序错位；所有参数为代码常量，无用户可控输入需非 canonical 试穿。
- **C3 授权边界 — N/A**：无权限/凭证/endpoint 改动。
- **C4 授权生命周期/级联 — N/A**：同上。
- **C5 build/运行期差异 — hit**：未在本环境构建/运行（Android 工程需 Gradle+SDK）；已做运行期推演（线程模型、位图生命周期、资源打包、错误路径降级到 `cleanup()`），视觉调参值列入 Human-verify。
- **C6 治理/文档自洽 — N/A**：非治理/策略类改动。

## 跨轮 blocker 复检（R6）

N/A —— 本 issue 无既往审查评论，为该 PR 首轮审查。

## 额外发现

- APK 体积 +1.8MB（`shangfang_celebrate.mp4`），与既有 action（1.4MB）/classy（0.5MB）同模式，仅提示。
- 一条引用本 PR 未变更文件（`internal/service/plugin/service.go`，属其他仓库上下文）的外部审查线索已按路径有效性校验剔除，未计入上述 findings。

## Verdict: APPROVED

新特效触发链路完整、sealed-class 穷举无缺口、`DEFAULT` 参数保证既有 action/classy 抠像数学逐位不变，未发现让能工作的路径不可用/产生错误数据/越权的 P0/P1 缺陷。三个 P2 中两个是本 PR 重构对既有特效的回归风险面（CPU 负载、绘制布局），建议以真机回归闭环；一个为 pre-existing 竞态，建议后续单独修。

[Octo-Q] verdict: APPROVE — 无 P0/P1 阻塞；仅 3×P2（2 个为本 PR 对既有特效的回归风险面 + 1 个 pre-existing 竞态）。建议终审在合并前要求作者对 action/classy 做低端机帧率与 16:9/折叠屏显示的真机回归确认。

yujiawei (approved):
# Code Review — PR #139 (octo-android)

Re-review at head `9260269c`, against merge-base `3133fe30`. The previous round was reviewed at `b9e12e4a`; this write-up re-verifies the whole change from the merge-base rather than inheriting that verdict, and additionally checks the incremental `b9e12e4a..9260269c` fix commit line by line. I re-parsed the mp4 boxes of all three assets myself rather than trusting the description.

**Measured asset facts** (from `mvhd` / `tkhd` / `stsd`, not from the PR body):

| asset | dimensions | duration | codec |
|---|---|---|---|
| `action_celebrate.mp4` | 1080×2004 | 2.234 s | hvc1 |
| `classy_celebrate.mp4` | 1080×2352 | 2.400 s | hvc1 |
| `shangfang_celebrate.mp4` | 884×1920 | 3.500 s | hvc1 |

---

## 1. Scope & spec compliance

**Result: ✅**

**Nothing missing.** The effect is registered end to end and I walked every link in the chain:

- asset present and a real 1.8 MB file — `wkbase/src/main/assets/effects/shangfang_celebrate.mp4`
- tag exists in the picker — `wkbase/src/main/assets/emoji/emoji.xml:7` (`Tag="[尚方宝剑]"`, `custom_shangfang.png` present)
- detection — `MessageEffectType.kt:41`, exact-match on the trimmed body, same gate as the other two video effects
- type — `MessageEffectType.kt:28`
- dispatch — `MessageEffectManager.kt:109-112` (the collect-window branch) and `MessageEffectManager.kt:158-160` (`videoSpecFor`)
- particle-system opt-out — `MessageEffectOverlayView.kt:113`
- keying params — `LumaKeyParams.kt:95-113`

**No unrequested functionality.** The only additions beyond the two review findings are comment rewrites, which the commit message discloses as comment-only. I diffed `b9e12e4a..9260269c` and confirmed the comment hunks carry no code change.

**No divergence.** The single blocking item from the previous round — the shared `onDraw` regressing `[崇尚行动]` from full-bleed to letterboxed — is genuinely fixed, and I verified the new math rather than taking the commit message's word for it.

`wkbase/src/main/java/com/chat/base/msgeffect/video/LumaKeyVideoView.kt:144-155`

```kotlin
if (width.toFloat() / height > aspect) {
    dstW = width.toFloat()
    dstH = dstW / aspect
} else {
    dstH = height.toFloat()
    dstW = dstH * aspect
}
```

This is exactly `scale = MAX(viewW/srcW, viewH/srcH)`: when `viewW/viewH > srcW/srcH` then `viewW/srcW > viewH/srcH`, so scaling by width is the covering scale, and the complementary branch handles the other case. Both branches are reachable with the real assets, and both are correct. Worked example on a 1080×2340 container with `action` (aspect 0.53892): the old code produced `dstH = 1080/0.53892 = 2004`, leaving 168 px transparent at each edge; the new code takes the `else` branch, gives `dstH = 2340`, `dstW = 1261`, and crops ~90 px per side instead. The hard horizontal termination of the translucent veil is gone.

Guards are sound: `width <= 0 || height <= 0` returns early at `:130`, and `sourceAspect()` (`:272-276`) cannot return 0 — `srcWidth/srcHeight` are gated on `> 0`, and the `Bitmap` fallback is always ≥ 1×1. No division by zero, no degenerate rect.

The second finding — the shared keying-resolution default — is also fixed, and the numbers in `LumaKeyParams.kt:74-77` check out. Computing `processSize()` (`:260-269`) in real `Float` precision:

| effect | cap | buffer | px/frame | vs. pre-PR 540×960 (518,400) |
|---|---|---|---|---|
| action | 960 | 517×960 | 496,320 | **0.96×** |
| classy | 960 | 440×960 | 422,400 | **0.81×** |
| shangfang | 1280 | 589×1280 | 753,920 | 1.45× (new effect only) |

Neither shipped effect regresses; `classy` actually gets ~19% cheaper. I also diffed `LumaKeyParams.DEFAULT` against the removed `companion object` at `3133fe30` — all six keying constants (0.10 / 0.12 / 0.05 / 0.45 / 0.30 / 0.14) match byte for byte, so the "action / classy keying math unchanged" claim now holds without the caveat the previous round attached to it.

Sealed-class exhaustiveness holds: the only two `when` expressions over `MessageEffectType` in non-test source are `MessageEffectOverlayView.createEffect` (all three video branches present) and `videoSpecFor` (has an `else`).

---

## 2. Code quality

**Result: Approved**

No P0 or P1. Everything below is advisory. Two items were raised at P1 by one of the advisory passes and I am ruling them down to P2 — reasoning given inline, since disagreement is more useful to you than a smoothed-over consensus.

### P2-1 — Aspect-fill center-crops hard on landscape / tablet / foldable / split panes

`wkbase/src/main/java/com/chat/base/msgeffect/video/LumaKeyVideoView.kt:144-155`

Aspect-fill covers the view unconditionally, so a container much wider than the source aspect crops the video vertically by a large factor. In a 16:9 landscape window only ~30% of `action`'s frame height and ~26% of `classy`/`shangfang`'s survives; at 19.5:9 it is ~25% / ~21%. This is reachable — chat is not orientation-locked:

`wkuikit/src/main/AndroidManifest.xml:109-112`

```xml
<activity android:name="com.chat.uikit.chat.ChatActivity"
    android:resizeableActivity="true"
    android:configChanges="screenSize|screenLayout|smallestScreenSize|orientation|keyboardHidden"
    android:windowSoftInputMode="stateAlwaysHidden|adjustResize" />
```

There is no `setRequestedOrientation` call anywhere in the repo, and the manifest also documents Activity Embedding split panes, so wide containers are a supported configuration rather than a theoretical one.

**Why P2 and not P1.** This is not a regression this PR introduces. On `main` the same path was `drawMatrix.setScale(width/bmp.width, height/bmp.height)` — a non-uniform stretch, which in a 16:9 landscape window stretched the frame horizontally by ~4.8×. Both states are wrong for that configuration; the change swaps "grotesquely distorted but complete" for "undistorted but heavily cropped", and for a full-screen translucent veil that is the more conventional trade. The previous round's aspect-fit code behaved identically here too (when the container is wider than the source, fit and fill select the same branch), so this is not something the fix commit caused. It is worth fixing, but blocking a third round on a pre-existing multi-form-factor gap would be whack-a-mole.

**Suggested fix.** Either clamp the fill scale (fall back to fit + a separate full-view scrim once the container/source aspect ratio diverges past some threshold), or size the video view to the source aspect and let a sibling dim layer cover the rest. Please spot-check a foldable inner screen and a split pane before this ships to those form factors.

**Related, same root cause, worth checking on-device:** `contentRoot` is `android.R.id.content` (`wkuikit/src/main/java/com/chat/uikit/chat/ChatActivity.java:811-816`), and the activity is `adjustResize`. With the IME up — the common case immediately after sending an emoji — the container shortens to roughly 1080×1310 on a 1080×2400 device, which pushes `shangfang`'s draw height to `1080/0.4604 = 2346` and crops ~44% of the frame. I could not confirm on-device whether the effect actually fires while the keyboard is up, so treat this as something to observe rather than a stated defect.

### P2-2 — Teardown does not fence the in-flight worker frame

`wkbase/src/main/java/com/chat/base/msgeffect/video/LumaKeyVideoView.kt:232-247` and `:400-406`

```kotlin
workerThread?.quitSafely()
workerThread = null
workerHandler = null
captureBitmap?.recycle()
captureBitmap = null
```

`quitSafely()` does not join, so `release()` can recycle and null `captureBitmap` while the worker is inside `getPixels`/`setPixels`, or before it dereferences the `captureBitmap!!` field captured by the lambda at `:234`.

**Why P2 and not P1.** One advisory pass rated this P1 and a second flagged the possibility of `onDraw` painting an already-recycled `displayBitmap`. I traced every path and no reachable crash exists today:

- `getPixels`/`setPixels` on a recycled bitmap throws `IllegalStateException`, and `captureBitmap!!` on a nulled field throws `KotlinNullPointerException` — both are `Exception` subclasses, both swallowed by the `catch (_: Exception) {}` at `:235`.
- The continuation then swaps three fields that `release()` has already set to `null`, so `onDraw` hits `val bmp = displayBitmap ?: return` at `:129` and returns.
- The recycled-`displayBitmap` draw is not reachable either: `release()` recycles and nulls `displayBitmap` in the same synchronous main-thread block (`:405-406`), so no draw traversal can observe the recycled-but-non-null state.
- Both `release()` call sites (`LumaKeyVideoEffectPlayer.kt:98` and `:112`) detach the view immediately afterwards, so the trailing `invalidate()` is a no-op.

So this is a latent hazard held together by a broad `catch`, not a live bug — correctness by accident across two files, and the accident currently holds. Still worth doing while the file is open: snapshot the bitmap into a local before `post`, and gate the main-thread continuation on a `released` flag so `onFirstFrame` cannot fire on a torn-down view.

### P2-3 — Disabled protect mask still allocates a full-size zero array

`wkbase/src/main/java/com/chat/base/msgeffect/video/LumaKeyVideoView.kt:341-354` and `:369-371`

```kotlin
val mask = FloatArray(w * h)
if (radiusRatio <= 0f) return mask
```

`buildMasks` builds both masks unconditionally, so `action` and `classy` — which have `eyeProtectRadius == 0f` (`LumaKeyParams.kt:53`) and never read `eMask` (`:330` is gated on `es > 0f`) — hold an all-zero `FloatArray(w*h)` for the whole playback: ~1.9 MiB for action, ~1.6 MiB for classy. Return `null` for a disabled radius and let `eyeMask` be genuinely absent. (Note this also drives the rebuild condition at `:301`, which currently requires `eyeMask != null` — that check needs to move to a separate flag if you make the mask nullable.)

### P2-4 — `pendingAfd` can leak the asset file descriptor (pre-existing)

`wkbase/src/main/java/com/chat/base/msgeffect/video/LumaKeyVideoView.kt:162-166` and `:170-204`

`startVideo` stores the `AssetFileDescriptor` and only opens the player if the `SurfaceTexture` already exists — which it does not at the point `LumaKeyVideoEffectPlayer.play` calls it (`LumaKeyVideoEffectPlayer.kt:61-76` adds the view and immediately calls `startVideo`, before any layout pass). So the fd always sits in `pendingAfd` for at least one frame. If the effect is cancelled in that window, nothing closes it: `release()` never touches `pendingAfd`, and the `catch` at `:201-203` skips `afd.close()` too. Add `try { pendingAfd?.close() } catch (_: Exception) {}` to `release()` and move the `afd.close()` at `:200` into a `finally`.

### P2-5 — `params` is shadowed inside `play()`

`wkbase/src/main/java/com/chat/base/msgeffect/video/LumaKeyVideoEffectPlayer.kt:45` / `:53` / `:57`

```kotlin
fun play(..., params: LumaKeyParams = LumaKeyParams.DEFAULT) {
    ...
    view.params = params                                  // :53 — keying params (correct today)
    ...
    val params = FrameLayout.LayoutParams(...)            // :57 — shadows the parameter
```

Correct only because `:53` precedes the local declaration. It compiles with a "Name shadowed" warning and is a landmine for anyone reordering these lines. Rename the local to `lp`. (Flagged last round, still open.)

### P2-6 — `SHANGFANG` keys 1.45× more pixels per frame than any shipped effect, on the main thread

`LumaKeyParams.kt:110-112`, `LumaKeyVideoView.kt:220`, `:316`

589×1280 = 753,920 px/frame vs. the 518,400 the shipped effects have run at. `decoderView.getBitmap(captureBitmap!!)` at `:220` is a GPU→CPU readback on the **UI thread**, and the `processing` latch does not protect it — the latch only skips *starting* new frames, so every frame that does start still pays the readback inline. The per-pixel Kotlin loop at `:316` then pays it again on the worker. This is a deliberate, documented sharpness trade for a genuinely more detailed asset, and it does not regress anything existing, so it is not blocking — but it is untested at 1.45× the known-good budget. Please measure frame time on a low-end device before shipping; if it janks, 1024 (471×1024 ≈ 482k px) is the obvious middle setting.

### P2-7 — small accuracy nits

- `MessageEffectManager.kt:147-152`: the comment still says action ≈ 3.2 s and classy ≈ 5.07 s. Measured `mvhd` durations are **2.234 s** and **2.400 s**. The fix commit rewrote the surrounding prose but kept the wrong numbers. The `8000L` timeout for classy (`:157`) is ~3.3× the real asset length. Harmless (it is only a watchdog) but misleading.
- `MessageEffectType.kt:28`: `ShangfangVideo(4000L)` is shorter than 3.5 s video + 0.45 s fade-in + 0.6 s fade-out = 4.55 s. `durationMs` is unused for video types (`createEffect` returns `null`, so `BaseEffect.isFinished` never runs), so this is inert — but it reads as a real budget sitting next to a 3.5 s asset.
- `LumaKeyParams.kt:88-92`: the rewrite dropped the provenance of the twelve `SHANGFANG` magic numbers. "Measured frame by frame" tells a future maintainer *how* they were derived but not *from what*. A one-line pointer to the source material or the tuning ticket would keep them auditable.
- `LumaKeyVideoView.kt:261-262` / `:273-274`: `srcWidth` and `srcHeight` are two independent `@Volatile` reads used to compute one aspect ratio. Safe today (both writers are `MediaPlayer` callbacks on the main thread, and `onPrepared` sets them before `start()`, so the fallback 540×960 path is essentially never taken), but a single `@Volatile` `Size`/packed-`Long` would make the pairing structural rather than incidental.
- Still no unit tests in this package. `detect()`, `centerStrengthAt()` and `processSize()` are pure, JVM-testable, and `processSize()` in particular has real logic (cap, aspect preservation, `toInt()` truncation). Zero coverage here today, so this is a suggestion rather than a requirement.

### Rejected

One advisory pass reported that the keying loop iterates `pixels.indices` and therefore churns stale pixels beyond `w * h` when the buffer shrinks. **This is wrong.** `LumaKeyVideoView.kt:295-297` reallocates on `pixelBuffer!!.size != size` — an inequality in *both* directions, not a grow-only check — so `pixels.size` always equals `w * h` exactly. No stale tail exists. (The same claim was raised and rejected last round.)

---

## 3. Verdict

**APPROVE**

The one blocking finding from the previous round is fixed, and I verified the fix from the math rather than from the commit message: `onDraw` now implements true aspect-fill, and the shared keying default at 960 leaves both shipped effects at or below their pre-PR per-frame cost (0.96× and 0.81×). The new effect itself is carefully built and correctly wired. Everything remaining is advisory and can be handled at your discretion or in follow-ups.

---

## 4. Suggested follow-ups

1. Clamp or fall back the fill scale for wide containers (P2-1), and spot-check a foldable inner screen / split pane.
2. Fence teardown with a `released` flag and a local bitmap snapshot (P2-2) — cheap, and removes the reliance on a blanket `catch`.
3. Return `null` for disabled protect masks (P2-3) and fix the `:301` rebuild condition accordingly.
4. Close `pendingAfd` in `release()`; move `afd.close()` into a `finally` (P2-4).
5. Rename the shadowed `params` local to `lp` (P2-5).
6. Measure `shangfang` frame time on a low-end device at `processMaxLongSide = 1280` (P2-6).
7. Correct the asset durations in the `MessageEffectManager` comment to 2.234 s / 2.400 s (P2-7).

---

## 5. Additional observations (outside this PR's scope)

- `MessageEffectType.detect` uses `text.contains("[使命必达]")` for the rocket effect (`MessageEffectType.kt:43`), whereas the three video effects require an exact trimmed match. An inline `[使命必达]` inside a long message still fires a full-screen effect. Pre-existing inconsistency, worth a separate issue.
- `shangfang_celebrate.mp4` is 884 px wide, which is not a multiple of 8 or 16. This relies on the HEVC conformance window being honoured. If edge artifacts appear on older MediaTek/Qualcomm decoders, an aligned width is the first thing to try.
- `LumaKeyVideoEffectPlayer.play` silently drops a second effect while one is playing (`:47`, `if (isPlaying) return`). Pre-existing and probably intended, but worth stating since a third effect makes collisions more likely.
- `classy`'s horizontal keying resolution drops from 540 to 440 samples with this change (the buffer now follows the source aspect). It is keyed correctly-shaped now instead of squashed, which is the net win, but it is ~19% fewer horizontal samples than the shipped build — include it in the regression pass.

---

## 6. What I could not verify

- **No build or test run.** No Android SDK in this environment, so the claimed `:app:assembleDebug` and `:wkbase:testDebugUnitTest` results are taken on trust. The exhaustiveness conclusion is from reading every `MessageEffectType` reference in non-test source, not from a compile.
- **No device rendering.** The eye-protect circle, the delayed centre disc, and the light-mode black-circle artifact are all appearance claims I could not confirm. All crop and pixel-count figures above are computed from asset dimensions and the draw math, not observed.
- **No frame-time or heap measurement**, on any device class.
- **HEVC decode availability** across the device fleet, and vendor differences in `TextureView.getBitmap` scaling behaviour.
- **Whether the effect fires with the IME visible**, which determines how much the `adjustResize` container shrink in P2-1 actually matters.
- **Back-to-back effects** and the collect-window interaction were not exercised.

Review legs actually run: this write-up, plus two independent advisory passes (one line-level adversarial, one whole-repo long-context). Both completed. Their disagreements with this verdict are recorded above rather than averaged away: one rated the wide-container crop and the teardown race as blocking, and I ruled both down to P2 with the evidence given in each section; one reported a stale-pixel loop bug that I checked and rejected outright.

Jerry-Xin (approved):
# Re-review — PR #139 at head `9260269c`

Re-review of the single fix commit `9260269c` ("修复 action 特效黑边与已有特效抠像分辨率回退") on top of `b9e12e4a`, covering the full PR state. I byte-verified every blocker raised by @yujiawei at the previous head against the new head, re-checked the merge-base `3133fe30` baseline, independently parsed the asset mp4 boxes with ffprobe (action 1080×2004 / 2.233s, classy 1080×2352 / 2.400s, shangfang 884×1920 / 3.500s), and re-verified the points from my earlier approval.

## Verdict: APPROVE — all three blocking items are fixed

### 1. ActionVideo full-screen layout — FIXED

`LumaKeyVideoView.kt:131-158`: `onDraw` now implements true aspect-fill:

```kotlin
if (width.toFloat() / height > aspect) {
    dstW = width.toFloat(); dstH = dstW / aspect
} else {
    dstH = height.toFloat(); dstW = dstH * aspect
}
val left = (width - dstW) * 0.5f
val top = (height - dstH) * 0.5f
dstRect.set(left, top, left + dstW, top + dstH)
```

This is exactly the `MAX(drawableSize.w/sw, drawableSize.h/sh)` scaling of iOS `WKLumaKeyVideoView.m:316`, applied over the full container bounds — which is precisely how iOS renders the action effect (`WKActionVideoEffect.m:31-36` gives the view the full bounds, then MAX fills). The previous width-fit letterbox is gone: on a 1080×2400 container, action (aspect 0.5389) now draws 1293×2400 centered (≈107 px cropped per side), covering the whole screen including the bottom blue-glow region that previously terminated in a hard horizontal line ~198 px above the bottom. On a 19.5:9 screen (1080×2340) it is 1261×2340. No device aspect can produce bands: the branch selection guarantees both dimensions are ≥ the view. `renderView` is a MATCH_PARENT child, so the Canvas clips the overflow naturally — no extra clip needed, and none is missing.

### 2. Default `processMaxLongSide` 1280 → 960 — FIXED

`LumaKeyParams.kt:81` now defaults to `processMaxLongSide = 960`; `LumaKeyParams.kt:112` sets `processMaxLongSide = 1280` only on `SHANGFANG`. This is the exact remedy suggested in the previous review. Call-site check: `MessageEffectManager.videoSpecFor` (`MessageEffectManager.kt:155-163`) routes action/classy through `VideoSpec`'s default `LumaKeyParams.DEFAULT` and shangfang through `LumaKeyParams.SHANGFANG`; there are no other `processMaxLongSide` references in the repo.

Per-frame buffer arithmetic at the new default (aspect-following `processSize()`):

| effect | source | buffer @960 | px/frame | vs pre-PR hardcoded 540×960 (518,400) |
|---|---|---|---|---|
| action | 1080×2004 | 517×960 | 496,320 | −4% |
| classy | 1080×2352 | 440×960 | 422,400 | −19% |
| shangfang | 884×1920 | 589×1280 (own knob) | 753,920 | n/a (new effect) |

The two shipped effects are now at or below their pre-PR per-frame pixel budget, so the +45–70% CPU regression is fully reverted; only the new effect pays for its sharper buffer.

### 3. Shared-path refactor vs iOS reference — fixed as a blocker, one residual micro-divergence noted

The two concrete divergences that made this blocking (action layout, resolution default) are both resolved as above. The remaining shared-path changes are all toward iOS parity, not away from it: aspect-following buffers (iOS keys at native `source.extent`, so following the source aspect is correct — the pre-PR 540×960 squash is what distorted the protect discs), the six keying constants in `LumaKeyParams.DEFAULT` byte-match the removed constants at merge-base `3133fe30` (`THRESHOLD 0.10 / TOLERANCE 0.12 / BG_FLOOR 0.05 / BG_CEIL 0.45 / CENTER_PROTECT_RADIUS 0.30 / CENTER_PROTECT_SOFTNESS 0.14`), and direct-alpha output preserves the original shipped behavior.

💬 One residual, non-blocking: on containers strictly taller than the asset ratio (>2.178:1, e.g. 20:9), classy/shangfang now aspect-fill (≈11–12 px cropped per side) where iOS's width-fitted frame leaves ≈24–27 px transparent bands top/bottom. On ≤19.5:9 screens both platforms clip the same few pixels. The cropped region is the keyed near-black frame edge, coverage of the veil is preserved, and this is within the fix options proposed in the previous review; noting it so it isn't a surprise in a side-by-side iOS/Android comparison. If pixel-exact parity is ever required, a per-effect layout mode (`FILL` vs `WIDTH_FIT`) is the way.

## Re-verified from my earlier approval (still hold)

- `MessageEffectType.detect()` exact-match gate for `[尚方宝剑]` unchanged (`MessageEffectType.kt:41`); only the iOS-reference comment was edited. Tag wiring via `custom_shangfang` from emoji.xml (#93) untouched.
- `when` exhaustiveness: `MessageEffectOverlayView.createEffect` covers all eight sealed subtypes; `videoSpecFor` keeps its `else -> null`.
- `LumaKeyParams.SHANGFANG` — all twelve keying fields unchanged from the reviewed head; the only addition is `processMaxLongSide = 1280`.
- Threading: `processing` latch intact; buffer/mask rebuild on size transition intact; `quitSafely` teardown unchanged. Watchdog `postDelayed(timeoutRunnable)` intact (`LumaKeyVideoEffectPlayer.kt:70`).

## Carried over (explicitly non-blocking per @yujiawei's review, still present)

- P2-2 disabled eye mask still allocates a full-size zero `FloatArray`.
- P2-3 teardown does not fence the in-flight worker frame (pre-existing).
- P2-4 `pendingAfd` fd leak on early cancel (pre-existing).
- P2-5 `params` shadowing in `play()`.
- P2-6 comment accuracy: the edited `MessageEffectManager.kt:148-150` comment now drops the "与 iOS 一致" claim but keeps action ≈ 3.2s / classy ≈ 5.07s while the assets are actually 2.233s / 2.400s (ffprobe); `ShangfangVideo(4000L)` is still shorter than the real 3.5s + fades ≈ 4.55s (inert). Worth a small follow-up.
- mochashanyao's pre-existing teardown race note folds into P2-3 above.

## Caveats

No device rendering or build in this environment; layout conclusions are computed from asset dimensions (ffprobe) + draw math, and CI is the compile gate. Recommend the 20:9 before/after pass on all three effects that the previous review asked for — the math says coverage is restored, and a visual pass would close it out.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #139 (octo-android)
Head `9260269c54aa479cec42a0d56fbb6aa5e5239526` · merge-base `3133fe30` · 7 files, +363/−54

## Summary

Adds the `[尚方宝剑]` full-screen luma-key video effect and refactors the luma-key pipeline from hardcoded constants to a per-effect `LumaKeyParams` value object. Three behavior-critical upgrades ride along: the keying buffer's aspect now follows the source asset instead of a fixed 540×960, drawing switches from stretch to aspect-fill, and the protection masks are rebuilt explicitly per effect (replacing a latent `by lazy` single-build that could not adapt to per-effect params). The second commit self-corrects two regressions the first commit introduced (default long-side cap 1280→960 to preserve the existing perf profile; aspect-fit→aspect-fill to kill black gaps on the shorter action asset). I reviewed the net diff. Existing action/classy keying behavior is provably unchanged — `LumaKeyParams.DEFAULT` reproduces every deleted constant exactly — and the new effect's wiring is complete at every dispatch site.

## Verification

Static analysis only at head `9260269c…`; the full Android build and instrumentation tests were not executed in this environment.

- ✅ **DEFAULT parity with pre-PR constants** — every removed constant in `wkbase/src/main/java/com/chat/base/msgeffect/video/LumaKeyVideoView.kt` (`THRESHOLD=0.10f`, `TOLERANCE=0.12f`, `BG_FLOOR=0.05f`, `BG_CEIL=0.45f`, `CENTER_PROTECT_RADIUS=0.30f`, `CENTER_PROTECT_SOFTNESS=0.14f`, center (0.5,0.5), eye off, gate off) matches the `LumaKeyParams()` defaults exactly (`wkbase/src/main/java/com/chat/base/msgeffect/video/LumaKeyParams.kt:30`); action/classy rendering is bit-identical.
- ✅ **Perf envelope preserved** — default `processMaxLongSide=960` yields action(1080×2004)→517×960≈496k px and classy(1080×2352)→440×960≈422k px per frame, both under the old fixed 540×960=518k px; SHANGFANG's 1280 cap (≈754k px) applies only to the new effect, and the `processing` latch drops frames instead of queueing.
- ✅ **Asset/geometry consistency** — parsed `wkbase/src/main/assets/effects/shangfang_celebrate.mp4` box headers: 884×1920, 3.500s, matching the kdoc figures; the 6000ms watchdog (`MessageEffectManager.kt:159`) comfortably covers the 3.5s playback.
- ✅ **Exhaustive dispatch coverage** — all four video-type sites handle `ShangfangVideo`: detect (`wkbase/src/main/java/com/chat/base/msgeffect/MessageEffectType.kt:41`), fireBestCandidate (`wkbase/src/main/java/com/chat/base/msgeffect/MessageEffectManager.kt:111`), videoSpecFor (`wkbase/src/main/java/com/chat/base/msgeffect/MessageEffectManager.kt:158`), createEffect→null (`wkbase/src/main/java/com/chat/base/msgeffect/MessageEffectOverlayView.kt:114`, enforced by the exhaustive `when`). Repo-wide grep (Kotlin + Java) finds no other exhaustive switch over this sealed type.
- ✅ **Wiring present** — manager constructed with `android.R.id.content` as contentRoot (`wkuikit/src/main/java/com/chat/uikit/chat/ChatActivity.java:816`); trigger asset `custom_shangfang.png` and the `[尚方宝剑]` emoji entry already exist at base.
- ✅ **Compile-check of the one suspect construct** — `play()`'s local `val params` shadowing the `params` parameter: compiled a faithful repro with kotlin-compiler-embeddable 2.2.0 (the exact version pinned by `build.gradle ext.kotlin_version`); it compiles clean, and a same-scope redeclaration control case errors as expected. Legal, deterministic, type-safe.
- ✅ **Threading** — `processFrame` runs on the main thread behind the `processing` latch; `MediaPlayer.currentPosition` is sampled on the main thread (MediaPlayer is not thread-safe) and travels with the frame to the worker; `view.params` is set before `startVideo` and only read afterwards, with Handler post establishing happens-before; release/cleanup races are benign (references captured at post, exceptions swallowed).

## Findings

No P0/P1/P2 issues; one readability nit below.

### Nit — local `params` shadows the same-named parameter (`wkbase/src/main/java/com/chat/base/msgeffect/video/LumaKeyVideoEffectPlayer.kt:53`)

`play()` declares `params: LumaKeyParams = LumaKeyParams.DEFAULT` (:45), consumes it at `view.params = params` (:53), then declares `val params = FrameLayout.LayoutParams(...)` (:57) in the same function body, so `container.addView(view, params)` (:61) binds the local. Verified legal on this repo's pinned Kotlin 2.2.0 (compile test above); resolution is deterministic — a local's scope starts at its declaration — and the two distinct types prevent any cross-use. Pure readability cost: rename the local to `layoutParams` so no future reader has to check whether this compiles. (Diff-scope: new in this PR; zero behavior impact.)

## Dataflow trace

- `LumaKeyParams` → view: `videoSpecFor` (`MessageEffectManager.kt:155`) → `play(root, assetPath, timeoutMs, lumaParams)` (`MessageEffectManager.kt:133`) → `view.params = params` before `startVideo` (`LumaKeyVideoEffectPlayer.kt:53`) — the ordering the mask-precompute comment requires; the `= LumaKeyParams.DEFAULT` default keeps action/classy call sites byte-identical.
- `srcWidth/srcHeight` → `processSize()`/`sourceAspect()`: filled by `onVideoSizeChanged`/`onPrepared` (`LumaKeyVideoView.kt:180-191`) before the first rendered frame in the normal `prepareAsync` sequence; until set, buffer (540×960 fallback) and draw aspect (`sourceAspect` falls back to the buffer's own ratio) stay self-consistent, so the one pre-size frame cannot distort — and it renders at alpha≈0 anyway (fade-in starts only on `onFirstFrame`).
- `MediaPlayer.currentPosition` → `centerStrengthAt` (`LumaKeyVideoView.kt:280`): sampled per frame on the main thread, drives SHANGFANG's timed center-disc gate (0 until 600ms, linear ramp over 650ms, then ≡1); reduces to constant 1 when `startTime<=0`, i.e. exactly the old action/classy behavior.
- Mask rebuild: the removed `centerMask by lazy` (single build regardless of params) is replaced by explicit rebuild — required now that masks depend on per-effect params; no lazy cache keyed on first use remains.

## Blind-spot checklist (C1–C6)

- **C1 dual-path parity** — hit, cleared: four symmetric dispatch sites for video types (detect/fireBestCandidate/videoSpecFor/createEffect), all updated for `ShangfangVideo`; grep confirms no other exhaustive dispatch site repo-wide.
- **C2 control-flow ordering / reuse** — cleared: `play()`/`processFrame`/`applyLumaKey` are shared by all three video effects; DEFAULT-parity proof above shows existing callers keep identical ordering and math, and the new gate degenerates to ≡1 for `startTime<=0`.
- **C3 authorization boundary** — N/A: local UI effect; no permission/tool/credential surface touched.
- **C4 authorization lifecycle** — N/A: same reason.
- **C5 build/note ≠ runtime** — partially applicable: no Android build executed here (environment constraint). Mitigated by a targeted compile-check of the only suspect construct plus runtime-path tracing of buffer sizing, aspect math, threading, and watchdog; stated as static analysis in Verification rather than asserted away.
- **C6 governance/doc self-consistency** — N/A: no policy/governance docs touched.

## Cross-round blocker recheck (R6)

N/A — first review round for this PR.

## Additional observations

- Commit 2 fixes the perf cap (1280→960 default) and the aspect-fit black gaps that commit 1 introduced; the net diff is what this review covers.
- `MessageEffectType.kt:28` gives `ShangfangVideo` `durationMs=4000`, same as ActionVideo; the video path actually consumes `VideoSpec.timeoutMs=6000` (`MessageEffectManager.kt:133`), mirroring the existing specs.
- No unit tests added for the msgeffect package — consistent with the repo (no tests exist for this area at base). Not counted against this PR, but `centerStrengthAt`/`processSize` are pure-function-shaped and would be cheap to pin.

## Verdict: APPROVED

The refactor is behavior-preserving for the two existing effects (constant-by-constant parity proven against the removed values), the new effect's dispatch is exhaustive and wired end to end, buffer/aspect math fixes a real shape distortion without exceeding the old per-frame pixel budget, and the only suspect construct (parameter shadowing) was proven legal on the project's exact Kotlin version. One optional readability rename; nothing blocking.

[Octo-Q] verdict: APPROVE — no P0/P1/P2; DEFAULT parity proven, dispatch exhaustive at all four sites, perf envelope preserved, shadowing verified legal on pinned Kotlin 2.2.0; only a readability nit remains.

PR #142, opened 2026-08-26
refactor(chat): 移除敏感词提示功能

### LINKED ISSUE #141: Sensitive-word tip is fabricated client-side and persisted as a real message
### Expected behavior

The chat transcript and the conversation list contain only messages that actually
came from the server.

### Actual behavior

When an inbound text message contains a word from the locally cached
`sensitive_words` list, the Android client fabricates a message of
`WKContentType.sensitiveWordsTips` (`-10`) and writes it into the local database
via `saveAndUpdateConversationMsg`, with `orderSeq = inboundMsg.orderSeq + 1`.

Consequences:

1. **A message that never existed is persisted.** It lives in the `message` table
   forever and reappears on every relaunch. It is not deduplicated against any
   server state because the server never knew about it.
2. **It hijacks the conversation list.** Because it is saved through
   `saveAndUpdateConversationMsg`, it becomes the conversation's last message, so
   the list preview shows the warning text instead of what the other party
   actually said.
3. **The matching has no authority behind it.** It is a plain
   `displayContent.contains(word)` scan over a client-cached word list. There is
   no server-side enforcement, so the tip neither blocks nor flags anything — and
   substring matching fires on innocent words that merely contain a listed token.
4. **It is platform-inconsistent.** The same conversation, viewed on another
   client, shows no such message — so the two transcripts disagree.
5. **It costs startup work on the critical path.** The word list is parsed from
   SharedPreferences in the synchronous startup phase (it must run before
   `initIMListener`, otherwise the filter sees a null list), plus a network sync
   afterwards — all to power a tip that provides no actual protection.

### Steps to reproduce

1. Have a non-empty `sensitive_words` list cached (synced from
   `GET message/sync/sensitivewords`).
2. Receive a text message from another user containing one of those words.
3. After ~2s, a grey system-style tip bubble appears below the message in the chat.
4. Back out to the conversation list — the preview for that conversation now shows
   the tip text, not the received message.
5. Kill and relaunch the app — the fabricated message is still there.

### Proposal

Remove the feature end to end (UI provider, layout, entity, network sync,
inbound-message filter, startup wiring) and add a DB migration that deletes the
rows it already wrote. Keep the `-10` content-type constant reserved so the
protocol number is never reused.

buyan749:
Fixes #141

## What

Removes the client-side sensitive-word tip end to end, and cleans up the rows it
already wrote into local databases.

Deleted:

| Path | Role |
|---|---|
| `chat/provider/WKSensitiveWordsProvider.kt` | rendered the tip bubble |
| `res/layout/chat_item_sensitive_words_layout.xml` | its layout |
| `enity/SensitiveWords.java` | `{tips, version, list}` payload |
| `MsgService.syncSensitiveWords()` | `GET message/sync/sensitivewords` binding |
| `MsgModel.syncSensitiveWords()` | version-gated sync + SharedPreferences cache |
| `WKUIKitApplication.parseLocalSensitiveWords()` | startup cache parse |
| `WKUIKitApplication.sensitiveWords` field | in-memory word list |
| the `WK_TEXT` branch in `WKIMUtils.initIMListener()` | matched words and built the fake message |
| the provider registration + Phase-C sync call | wiring |

Kept on purpose:

- `WKContentType.sensitiveWordsTips = -10`, now `@Deprecated`. The value was in
  use on the wire; deleting it invites a future reuse that would make historic
  rows on not-yet-upgraded clients render as some unrelated type.

Added:

- `wkim/src/main/assets/wk_sql/202608191100.sql` — `DELETE FROM message WHERE type = -10;`

## Why

The tip was fabricated locally and then persisted as if it were a real message:

- `WKIMUtils` scanned every inbound `WK_TEXT` with `displayContent.contains(word)`
  against a client-cached list, then built a `WKMsg` with
  `orderSeq = inbound.orderSeq + 1` and saved it through
  `saveAndUpdateConversationMsg`.
- Because it went through `saveAndUpdateConversationMsg`, it also became the
  conversation's last message, so the conversation list preview showed the warning
  text instead of the message the user actually received.
- Nothing enforced anything. There is no server-side counterpart to the match, so
  the tip blocked nothing and reported nothing; substring matching also fired on
  ordinary words that happened to contain a listed token.
- The same conversation on another client shows no such message, so the two
  transcripts disagree about what is in the history.
- It occupied the synchronous startup phase: the cached list had to be parsed
  before `initIMListener` (otherwise the inbound filter would see `null`), plus a
  network sync in the idle phase — for a feature that provided no protection.

Given all of that, the tip is not worth fixing in place; the right move is to stop
producing it and clean up what it already produced.

## How

- **Inbound path** — dropped the `WK_TEXT` matching block and the delayed
  `saveAndUpdateConversationMsg`. The surrounding loop is otherwise untouched:
  the same messages are processed in the same order; only the extra fabricated
  message is gone.
- **Startup** — `parseLocalSensitiveWords()` is deleted from Phase-B, so Phase-B is
  now just `initIM()` + `initIMListener()`. Phase-C keeps `ProhibitWordModel.sync()`
  and `deleteFlameMsg()` and loses `syncSensitiveWords()` only. Phase ordering and
  the reasons for it are unchanged; the comments were updated to match.
- **Migration** — the SQL file is picked up by `WKDBUpgrade`, which sorts asset
  filenames as numeric indices and runs anything above the stored index exactly
  once, then advances the index whether or not the statement succeeded. `202608191100`
  is the highest index in `wk_sql/`, so it runs once on upgrade and never again.
  The deletion is safe to make unconditional: type `-10` rows were only ever
  produced by the code being removed here, are local-only, and were never synced
  from or to the server — so no user-authored content is touched.

## Scope

- Client-only. The `message/sync/sensitivewords` endpoint is left alone; other
  clients are unaffected by this PR.
- No behaviour change for any other message type. `ProhibitWord` — a separate
  feature with a similar-sounding name — is untouched.

## Testing

- `./gradlew :wkuikit:compileDebugJavaWithJavac` passes.
- Manual, on a device that already had type `-10` rows in its database:
  - [ ] After upgrading, the previously visible tip bubbles are gone from the
        affected conversations.
  - [ ] The conversation list preview for those conversations now shows the real
        last message.
  - [ ] Receiving a message that would previously have matched produces no tip and
        no extra row.
  - [ ] Relaunching does not bring anything back (migration index advanced).
  - [ ] Normal text / image / system messages render unchanged.
  - [ ] Cold start still lands on a populated conversation list (Phase-B ordering
        unchanged).


Jerry-Xin (changes requested):
Deletion refactor reviewed at head 0a26e9be. The feature removal itself is complete and clean — but the companion migration deletes the fabricated message rows without repairing the conversation metadata they polluted, leaving a user-visible dangling-pointer state. One 🔴 below.

## 🔴 Blocking

- 🔴 **Migration leaves `conversation.last_client_msg_no` dangling** — `wkim/src/main/assets/wk_sql/202608191100.sql` does `DELETE FROM message WHERE type = -10`, but the removed feature persisted its tip through `MsgManager.saveAndUpdateConversationMsg` → `ConversationDbManager.insertOrUpdateWithMsg` (wkim/src/main/java/com/xinbida/wukongim/db/ConversationDbManager.java:406), which wrote the fabricated row's `clientMsgNO` / `messageSeq` / `timestamp` into the conversation row (`last_client_msg_no`, `last_msg_seq`, `last_msg_timestamp`). `WKMsg()` auto-generates `clientMsgNO` (wkim/src/main/java/com/xinbida/wukongim/entity/WKMsg.java:100), so those pointers are real and now reference deleted rows. Verified downstream effect: `WKUIConversationMsg.getWkMsg()` returns null for the stale pointer, `ConversationPreviewSelector.hydrateSystemBotEmpty` (wkbase/src/main/java/com/chat/base/space/ConversationPreviewSelector.java:233) only falls back to the newest remaining message for SystemBot channels, and for every regular DM/group the list preview renders **empty** (`getContent(null)` → "") until the next message arrives — possibly indefinitely for quiet conversations. This contradicts the PR's own test-checklist item "conversation list preview ... now shows the real last message" (true only for SystemBot channels). No crash paths — all consumers null-guard (ChatConversationAdapter.java:2000, WKIMUtils.buildStartChatIntent, MsgManager.java:877) — but this is a partial removal of the feature's persistence footprint, which is exactly what this refactor set out to eliminate. Suggested fix in the same migration: for conversations whose `last_client_msg_no` points at a `type = -10` row, re-point to the channel's newest remaining message row (update `last_client_msg_no` / `last_msg_seq` / `last_msg_timestamp`) before deleting, so the preview restores the real last message.

## 💬 Non-blocking

- Commit 0a26e9be ("清理敏感词移除后的残留未使用 import") removed 3 orphaned imports but missed 3 more that this same deletion orphaned: `org.json.JSONException` and `com.xinbida.wukongim.msgmodel.WKTextContent` in `wkuikit/src/main/java/com/chat/uikit/chat/manager/WKIMUtils.java` (both now referenced only from comments, lines 518/532) and `com.alibaba.fastjson.JSON` in `wkuikit/src/main/java/com/chat/uikit/message/MsgModel.java` (its only use was `JSON.toJSONString` inside the removed `syncSensitiveWords`). No compile impact (Java warnings only); worth finishing the sweep. (The duplicate `import org.json.JSONObject;` in WKIMUtils is pre-existing, not from this PR.)
- The SP keys `wk_sensitive_words` / `wk_sensitive_words_version` have zero references at this head — existing on-device entries are orphaned forever. Harmless, but a one-time SP cleanup would complete the removal.
- Consider a migration test for a conversation whose last message is `type = -10` (including the "no earlier message remains" case), and adjust the PR test-checklist wording for the regular-chat preview behavior once fixed.

## ✅ Highlights

- Full-chain removal verified byte-clean: repo-wide grep at head shows zero remaining references to `WKSensitiveWordsProvider`, `SensitiveWords`, `syncSensitiveWords`, `parseLocalSensitiveWords`, the `message/sync/sensitivewords` binding, the SP keys, or `chat_item_sensitive_words_layout`; `sensitiveWordsTips` survives only as the deprecated constant definition.
- Keeping `WKContentType.sensitiveWordsTips = -10` as `@Deprecated` with a do-not-reuse rationale is the right protocol-hygiene call.
- Migration mechanics are sound: `202608191100` is the highest index in `wk_sql/`, `WKDBUpgrade` runs it exactly once at DB open (before any read), and `message(type)` is indexed so the DELETE is cheap.
- The inbound-message loop in `WKIMUtils.initIMListener()` is otherwise untouched; no orphaned resources (`colorSystemBg` / `SystemMsgBackgroundColorSpan` remain shared), no scope creep — the +15 lines are the javadoc and the SQL.

yujiawei (approved):
# Code Review — PR #142 (octo-android)

Reviewed at head SHA `0a26e9beeb825ec5b8aa94b84bd492a5bd8f3c83` against `main`. Full checkout, plus two independent adversarial passes over the same diff; every claim below was re-verified against the code before being kept or dropped.

---

## 1. Spec compliance

**Spec: ✅**

Issue #141's *Proposal* section enumerates eight deliverables. All eight are present, and nothing beyond them:

| #141 asks for | In this diff |
|---|---|
| UI provider | `WKSensitiveWordsProvider.kt` deleted |
| layout | `res/layout/chat_item_sensitive_words_layout.xml` deleted |
| entity | `enity/SensitiveWords.java` deleted |
| network sync | `MsgService.syncSensitiveWords()` + `MsgModel.syncSensitiveWords()` deleted |
| inbound-message filter | `WK_TEXT` branch removed from `WKIMUtils.initIMListener()` |
| startup wiring | `parseLocalSensitiveWords()`, the `sensitiveWords` field, the provider registration, and the Phase-C sync call removed from `WKUIKitApplication` |
| DB migration for existing rows | `wkim/src/main/assets/wk_sql/202608191100.sql` added |
| keep `-10` reserved | `WKContentType.sensitiveWordsTips = -10` kept, now `@Deprecated` with a do-not-reuse note (`WKContentType.java:44-55`) |

- **Missing:** none.
- **Extra / out of scope:** none. `ProhibitWord` — the similarly-named but genuinely separate feature (own table via `ProhibitWordDB`, own `syncProhibitWord` endpoint) — is untouched, as claimed.
- **Deviations:** none. A repo-wide grep for `sensitiveWords|SensitiveWords|sensitive_words|sensitivewords` returns only the two surviving lines in `WKContentType.java`; no dangling references in any module.

---

## 2. Code quality

**Quality: Approved** — no P0 or P1. Four P2s below.

### Verified safe (things that looked risky and are not)

- **The unconditional `DELETE` cannot destroy server-authored content.** The removed block built the tip purely locally and persisted it via `saveAndUpdateConversationMsg` (`MsgManager.java:691`), which does a local insert plus a conversation-row update and no network send. Type `-10` rows therefore have no `message_id` and were never uploaded. Deleting all of them is safe.
- **Multi-account is not affected.** The migration cursor is keyed per user — `setting.getLong(getUid() + "_db_upgrade_index", 0)` (`WKIMApplication.java:154`) — so a second account on the same device still runs the cleanup against its own DB.
- **Leftover `-10` rows cannot crash the UI** if the migration ever fails to run. `ChatAdapter.getItemType()` falls through to `return WKContentType.unknown_msg;` (`wkbase/.../msg/ChatAdapter.java:212`) for any unregistered type, and `unknown_msg` is unconditionally registered with `WKUnknownProvider` (`WKMsgItemViewManager.java:49`). The conversation list has the matching fallback at `ChatConversationAdapter.java:947` (`R.string.unknow_msg_type`). Worst case is a degraded "unknown message type" row, not an inflation NPE.
- **No unread-count residue.** The tip was saved with `addRedDots = false`, so it never incremented `unread_count`; nothing to clean up there.
- **The single-line SQL survives the loader.** `WKDBUpgrade.getExecSQL()` concatenates asset lines with `stringBuilder.append(line)` and **no separator** (`WKDBUpgrade.java:86`) before splitting on `;`. One statement on one line is fine — see the P2-4 note before adding a second.
- **Phase ordering is intact.** Phase-B is now `initIM()` + `initIMListener()`; the removed `parseLocalSensitiveWords()` was the only reason Phase-B had a pre-listener step, and its removal cannot reintroduce the empty-conversation-list race the surrounding comment describes.

### P2-1 — The migration leaves `conversation.last_client_msg_no` pointing at a deleted row

`wkim/src/main/assets/wk_sql/202608191100.sql:1`

```sql
DELETE FROM message WHERE type = -10;
```

Because the tip went through `saveAndUpdateConversationMsg`, it became the conversation's last message: `wkConversationMsg.lastClientMsgNO = msg.clientMsgNO` (`ConversationDbManager.java:420`). The migration deletes the `message` row but does not touch the `conversation` row.

The preview then resolves to nothing:

```java
// WKUIConversationMsg.java:44-48
public WKMsg getWkMsg() {
    if (wkMsg == null) {
        wkMsg = MsgDbManager.getInstance().queryWithClientMsgNo(clientMsgNo);
```

and for an ordinary DM or group there is no fallback — `ConversationPreviewSelector.java:113` is explicit that `raw == null` for a non-SystemBot channel means "no message, null is correct".

**Why this is P2 and not P1:** it self-heals. Cold start forces a *full* conversation sync — `final long version = coldStartSyncDone ? dbVersion : 0;` (`ConversationManager.java:489`) — and the response rewrites the pointer from server state at `ConversationManager.java:627`, persisted with `CONFLICT_REPLACE` (`WKDBHelper.java:454`). Since the tip was never known to the server, the server's `last_client_msg_no` is the real last message. Every consumer null-checks (e.g. `ChatConversationAdapter.java:982`), so there is no crash path.

**Residual impact:** a blank preview between DB open and the sync response landing on the upgrade launch, and a blank preview that persists for the whole session if the device is offline on first launch after upgrade.

**This matters for the PR description**, which currently lists as a migration outcome:

> - [ ] The conversation list preview for those conversations now shows the real last message.

That is true, but it is the conversation sync doing it, not the migration. Either:

1. repair the pointer in the migration — statement order matters, and it must go **before** the `DELETE`, each statement on its own single line:

   ```sql
   UPDATE conversation SET last_client_msg_no = COALESCE((SELECT m.client_msg_no FROM message m WHERE m.channel_id = conversation.channel_id AND m.channel_type = conversation.channel_type AND m.type <> -10 AND m.is_deleted = 0 AND m.type <> 0 AND m.type <> 99 ORDER BY m.order_seq DESC LIMIT 1), last_client_msg_no) WHERE last_client_msg_no IN (SELECT client_msg_no FROM message WHERE type = -10);
   DELETE FROM message WHERE type = -10;
   ```

   (this still leaves `last_msg_timestamp` / `last_msg_seq` stale until the next sync, so it is a partial fix); **or**

2. keep the migration as-is and reword that checklist item to say the preview is restored by the cold-start conversation sync — then add the test case that actually exercises the gap: **upgrade while offline, confirm the preview is acceptable before any sync completes.**

Option 2 is defensible. Option 1 without option 2's test is not.

### P2-2 — Retired SharedPreferences keys are never cleared

`wkuikit/src/main/java/com/chat/uikit/message/MsgModel.java` (deleted `syncSensitiveWords`, previously ~L1163-1181)

The removed code wrote two keys that nothing now reads or erases:

```java
WKSharedPreferencesUtil.getInstance().putLong("wk_sensitive_words_version", result.version);
WKSharedPreferencesUtil.getInstance().putSP("wk_sensitive_words", json);
```

Every upgrading install keeps the full downloaded word list on disk indefinitely. `WKSharedPreferencesUtil` normally uses `EncryptedSharedPreferences` (AES256-GCM), which limits the exposure — but it has a plaintext `context.getSharedPreferences(mTAG, Context.MODE_PRIVATE)` fallback path, so on any device that took the fallback the list is sitting there in cleartext. A cleanup pass is cheap and belongs in the same PR that removes the writers; the `## What` table lists every other artifact this change reclaims, so the omission reads as an oversight rather than a decision.

Suggested: a one-time removal of both keys, guarded by its own SP flag so it does not run on every launch.

### P2-3 — Three imports are now unused

The removal orphaned imports whose only live uses were inside the deleted blocks (the remaining textual occurrences are commented-out code):

- `wkuikit/src/main/java/com/chat/uikit/message/MsgModel.java:26` — `import com.alibaba.fastjson.JSON;` (sole use was `JSON.toJSONString(result)` in the deleted method; `JSONObject` / `JSONArray` on lines 27-28 are still used, `JSON` is not)
- `wkuikit/src/main/java/com/chat/uikit/chat/manager/WKIMUtils.java:85` — `import com.xinbida.wukongim.msgmodel.WKTextContent;` (only remaining occurrence is the comment at L518)
- `wkuikit/src/main/java/com/chat/uikit/chat/manager/WKIMUtils.java:87` — `import org.json.JSONException;` (only remaining occurrence is the comment at L532)

Cosmetic — javac warns, it does not fail — but the diff already removed `WKSendMsgResult` and `java.util.Objects` from the same file, so the cleanup was clearly intended to be complete.

### P2-4 — Forward-looking note on this migration file

Two properties of the loader are worth knowing before anyone edits `202608191100.sql`:

1. **No newline separator.** `WKDBUpgrade.java:86` does `stringBuilder.append(line)` with nothing between lines, then splits on `;`. A statement wrapped across two lines gets its tokens glued together into invalid SQL. Keep one statement per line.
2. **The index advances even when the statement throws** (`WKDBUpgrade.java:57-64`): the `catch` logs, sets `success = false`, and the code then calls `setDBUpgradeIndex(dbSql.index)` unconditionally. Every prior file in `wk_sql/` is idempotent DDL (`ALTER TABLE` / `CREATE INDEX IF NOT EXISTS`), where that is the right trade-off. This is the first **data-mutating** migration in the directory, so it is the first one where a single transient failure means the cleanup never happens for that user — permanently. No change requested here (it is pre-existing infrastructure), but it is the reason P2-1's "leftover rows can't crash the UI" verification above was worth doing.

---

## 3. Overall verdict

**APPROVE**

Spec ✅ and no P0/P1. This is a well-scoped, well-argued deletion: the reasoning in `## Why` matches what the code actually did, the `-10` constant is correctly retained rather than freed for reuse, and the removal is surgical — the inbound-message loop is otherwise byte-identical and no other message type changes behaviour. The four P2s are follow-ups, not blockers.

---

## 4. Suggested follow-ups

1. Pick one of the two options in P2-1 — either the pointer repair, or an honest rewording plus an **offline-upgrade** test case. Do not ship the current checklist wording unchanged.
2. Add the one-time `wk_sensitive_words` / `wk_sensitive_words_version` cleanup (P2-2).
3. Drop the three orphaned imports (P2-3).
4. The manual test checklist in the PR description is entirely unchecked. Given this PR deletes rows from the message table on upgrade, please run it on a device that actually has type `-10` rows and check the boxes before merging.

---

## 5. What still needs human eyes

This PR carries `needs-human-review`. Static analysis cannot settle these; none of them are code defects, and none of them changed my verdict:

1. **Compliance sign-off.** The change is technically sound — the tip enforced nothing, it fired *after* the message was already delivered and rendered, and it blocked no send or receipt. But "a visible content warning existed and now does not" is a product and regulatory question, not an engineering one. Someone who owns that call should confirm no obligation depended on this bubble being on screen.
2. **Server-side and other clients.** `GET message/sync/sensitivewords` is intentionally left live. Worth confirming (a) whether any other consumer still depends on it and when it can be retired, and (b) whether iOS/Web still fabricate the same tip — if they do, issue #141's complaint #4 ("the two transcripts disagree") is only half fixed, and the disagreement simply flips direction.
3. **Migration on real data.** Nobody has run this against a populated production-shaped database, only against source.

---

## 6. Coverage note

What this review did **not** do: run Gradle or any test locally (the repo's own `build` check is green at this SHA and is the only compile evidence here); instrument an upgraded SQLite database; exercise the offline-then-online startup sequence on a device; inspect `octo-server` for enforcement or for other consumers of the sensitive-words endpoint; or check the iOS and Web clients for parity. Areas outside the diff that were deliberately checked and found clean: repo-wide symbol references across `wkbase` / `wkuikit` / `wkim` / `app`, the migration loader and index semantics, conversation hydration and the conversation-sync repair path, the unregistered-message-type render fallback, and multi-account migration keying.

yujiawei (changes requested):
# Code Review — PR #142 (octo-android) — correction to my earlier approval

**This review supersedes my previous `APPROVED` on the same head SHA `0a26e9beeb825ec5b8aa94b84bd492a5bd8f3c83`.** I graded the dangling-conversation-pointer issue as non-blocking on the strength of a self-healing argument. I have since checked the server, and that argument is wrong. Re-grading to blocking, and retracting the approval.

## What I got wrong

In my earlier review I argued the stale `conversation.last_client_msg_no` repairs itself on the next launch, reasoning from the client:

```java
// wkim/src/main/java/com/xinbida/wukongim/manager/ConversationManager.java:489
final long version = coldStartSyncDone ? dbVersion : 0;
```

I read `version = 0` as "cold start requests a full conversation sync", which would rewrite every conversation row from server state (`ConversationManager.java:627`, persisted with `CONFLICT_REPLACE` at `WKDBHelper.java:454`) and restore the correct preview within the same launch.

**The server does not honour `version = 0`.** In `octo-server` (`4442db49`), `syncUserConversation` overwrites the client's value with a per-device cached cursor:

```go
// modules/message/api_conversation.go:314
version := req.Version
...
// :319
if !co.ctx.GetConfig().MessageSaveAcrossDevice {
    // :324
    cacheVersion, err := co.getDeviceConversationMaxVersion(loginUID, req.DeviceUUID)
    ...
    // :347
    if cacheVersion > version {
        version = cacheVersion
    }
```

`messageSaveAcrossDevice` is commented out in `configs/tsdd.yaml:8`, so it defaults to Go's zero value `false` and this branch is live on the default deployment. The client's "full sync" is therefore silently downgraded to an incremental one: only conversations whose version advanced past the device cursor come back. **A quiet conversation gets nothing, so its pointer is never rewritten.**

There is no self-heal. The blank preview persists until a new message actually arrives in that conversation — indefinitely for a dormant chat.

## Re-graded finding

### P1 (blocking) — Migration leaves `conversation.last_client_msg_no` pointing at a deleted row

`wkim/src/main/assets/wk_sql/202608191100.sql:1`

```sql
DELETE FROM message WHERE type = -10;
```

The tip was persisted through `saveAndUpdateConversationMsg`, which wrote its `clientMsgNO` into the conversation row (`ConversationDbManager.java:420`). Deleting the message row without repairing the conversation row leaves the pointer dangling:

```java
// WKUIConversationMsg.java:44-48
public WKMsg getWkMsg() {
    if (wkMsg == null) {
        wkMsg = MsgDbManager.getInstance().queryWithClientMsgNo(clientMsgNo);
```

returns `null`, and for a regular DM or group `ConversationPreviewSelector` has no fallback — `ConversationPreviewSelector.java:113` states plainly that `raw == null` on a non-SystemBot channel means "no message, null is correct". The list preview renders empty.

No crash: every consumer null-guards (`ChatConversationAdapter.java:982`, `ChatAdapter.getItemType()` falls back to `WKContentType.unknown_msg` at `ChatAdapter.java:212`). But this directly contradicts the PR's own checklist item —

> - [ ] The conversation list preview for those conversations now shows the real last message.

— which holds only for SystemBot channels. And it is a partial removal of exactly the persistence footprint this refactor set out to eliminate: the feature's fingerprint survives in the conversation table after the migration claims to have cleaned up.

**Suggested fix**, in the same migration, before the `DELETE`. Note `WKDBUpgrade.getExecSQL()` concatenates asset lines with no separator (`WKDBUpgrade.java:86`) and splits on `;`, so each statement must stay on one line:

```sql
UPDATE conversation SET last_client_msg_no = COALESCE((SELECT m.client_msg_no FROM message m WHERE m.channel_id = conversation.channel_id AND m.channel_type = conversation.channel_type AND m.type <> -10 AND m.is_deleted = 0 AND m.type <> 0 AND m.type <> 99 ORDER BY m.order_seq DESC LIMIT 1), last_client_msg_no) WHERE last_client_msg_no IN (SELECT client_msg_no FROM message WHERE type = -10);
DELETE FROM message WHERE type = -10;
```

`last_msg_seq` and `last_msg_timestamp` want the same treatment, and the "no earlier message remains in the channel" case needs deciding explicitly rather than falling out of a `COALESCE`.

This converges with @Jerry-Xin's independent finding on the same PR; their read of the persistence footprint was right and mine was not.

## Unchanged from my earlier review

Spec compliance is still **✅** — all eight deliverables in issue #141's *Proposal* are present, nothing extra, `ProhibitWord` correctly untouched, and a repo-wide grep leaves only the deprecated `-10` constant. These verifications also stand:

- The unconditional `DELETE` cannot destroy server-authored content — the tip was local-only, never uploaded, and has no `message_id`.
- Migration keying is per-user (`WKIMApplication.java:154`), so multi-account devices each run the cleanup.
- Leftover `-10` rows cannot crash the UI if the migration fails — `unknown_msg` is unconditionally registered (`WKMsgItemViewManager.java:49`).
- No unread-count residue (`addRedDots = false`).
- Phase-B/Phase-C ordering is intact.

Still open at P2, unchanged:

- **Retired SP keys never cleared.** `wk_sensitive_words` / `wk_sensitive_words_version` have zero readers at this head but are never erased. `WKSharedPreferencesUtil` normally uses `EncryptedSharedPreferences`, but it has a plaintext `MODE_PRIVATE` fallback, so on any device that took the fallback the word list stays on disk in cleartext.
- **Three imports orphaned by this removal:** `MsgModel.java:26` (`com.alibaba.fastjson.JSON`), `WKIMUtils.java:85` (`WKTextContent`), `WKIMUtils.java:87` (`JSONException`) — each now referenced only from commented-out code at `WKIMUtils.java:518` / `:532`.
- **The manual test checklist is entirely unchecked.** For a PR that deletes rows on upgrade, please run it against a device that actually has `-10` rows — including the offline-upgrade path, which is where the P1 above is most visible.

## Overall verdict

**REQUEST_CHANGES**

The feature removal itself remains clean, well-argued work, and the `@Deprecated` treatment of the `-10` constant is the right protocol-hygiene call. The blocker is narrow and fixable in the migration file.

## Still needs human eyes

`needs-human-review` is on this PR and these are not code questions:

1. **Compliance sign-off.** The tip enforced nothing — it fired after delivery and blocked no send or receipt — so no technical control is lost. Whether a visible content warning was required for any other reason is a product/regulatory call.
2. **Server-side and other clients.** `GET message/sync/sensitivewords` stays live; worth confirming who else consumes it, and whether iOS/Web still fabricate the same tip. If they do, issue #141's complaint #4 ("the two transcripts disagree") only changes direction.

## Coverage note

Did not run Gradle or any test locally (the repo's `build` check is green at this SHA and is the only compile evidence), instrument an upgraded SQLite database, or exercise startup on a device. On the server I read only `syncUserConversation` in `modules/message/api_conversation.go` and the `messageSaveAcrossDevice` default; I did not verify how any specific production deployment sets that flag — if it is `true` somewhere, the version-override branch is skipped there and the self-heal I originally described would apply on that deployment. That does not change the verdict, since the default path is the one users upgrade on. Not checked: iOS/Web parity, and other consumers of the sensitive-words endpoint.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #142 (octo-android)

## Summary

This PR removes the client-side sensitive-word tip feature end to end. That feature matched incoming `WK_TEXT` content against a locally synced word list and fabricated a fake `type=-10` tip message, inserting it into the local message DB and overwriting the conversation's last-message pointer (`WKIMUtils.java`). The PR deletes the provider, layout, entity, the `message/sync/sensitivewords` call, the SP parse step, and the fake-message construction; it deprecates the `sensitiveWordsTips = -10` constant (deliberately kept so the protocol number is never reused) and adds migration `202608191100.sql` to delete historical `type=-10` rows. The removal is reference-clean; two non-blocking P2 cleanup-parity items remain.

## Verification

Static analysis only at head `0a26e9be`; build and tests not executed in this environment.

- ✅ **No dangling references** — repo-wide grep for `sensitiveWordsTips` / `sensitiveWords` / `SensitiveWords` / `syncSensitiveWords` / `wk_sensitive_words` / `chat_item_sensitive_words_layout`: only the intentionally deprecated constant remains (`wkbase/src/main/java/com/chat/base/msgitem/WKContentType.java:55`).
- ✅ **Migration wiring** — `WKDBHelper`'s constructor runs `WKDBUpgrade.onUpgrade` on every DB open, before any read (`wkim/src/main/java/com/xinbida/wukongim/db/WKDBHelper.java:147`); execution is gated by a per-user index in SP, and the new file is the highest index, so it runs last. `message.type` exists in the baseline schema and is covered by `type_index` (`wkim/src/main/assets/wk_sql/202008292051.sql:27`) — the DELETE is index-backed, no full-scan stall on first open.
- ✅ **Sole producer removed** — the deleted block in `WKIMUtils` was the only code creating `type=-10` messages; they were local-only (`saveAndUpdateConversationMsg`, never uploaded), so no server or cross-device source can reintroduce them.
- ✅ **Unknown-type safety** — `ChatAdapter.getItemType` downgrades unregistered types to `unknown_msg(-3)` (`wkbase/src/main/java/com/chat/base/msg/ChatAdapter.java:201-212`), so even a surviving row (e.g. if the migration statement failed) degrades gracefully instead of crashing.
- ✅ **Null-safe preview path** — `getWkMsg()` lazy lookup may return null; `ChatConversationAdapter.getContent(null)` returns `""` (`wkuikit/src/main/java/com/chat/uikit/chat/adapter/ChatConversationAdapter.java:894-896`) and `resetMsgProhibitWord(null)` early-returns (`WKIMUtils.java:569-571`).
- ✅ **Unread counts unaffected** — fake messages were inserted with `addRedDots=false`, so the conversation unread delta was always 0.
- ✅ **Commit-2 import cleanup verified** — no remaining users of `WKSendMsgResult` / `Objects` / fastjson `JSON` in the touched files.

## Findings

No P0/P1 issues; two P2 items below.

### P2 — Migration leaves a dangling conversation pointer; preview blank until the next message (`wkim/src/main/assets/wk_sql/202608191100.sql:1`)

The DELETE removes only `message` rows. For a conversation whose last message was a sensitive-word tip, the conversation row's `last_client_msg_no` still points at a deleted row — the fake message did get a generated `clientMsgNO` from the `WKMsg()` constructor (`wkim/src/main/java/com/xinbida/wukongim/entity/WKMsg.java:100`), and `insertOrUpdateWithMsg` stored it as the conversation pointer (`ConversationDbManager.java:420`). After upgrade, `WKUIConversationMsg.getWkMsg()` resolves null (`WKUIConversationMsg.java:45`), so that conversation shows an empty preview in the list until its next real message. Scope is limited to conversations where a tip was the last message, and it self-heals — hence P2, not P1. Fix direction: in the same migration file, *before* the DELETE, repair the pointer, e.g. `UPDATE conversation SET last_client_msg_no = '' WHERE last_client_msg_no IN (SELECT client_msg_no FROM message WHERE type = -10);` — or explicitly accept the transient blank in the PR description.

### P2 — Orphaned SharedPreferences keys: plaintext sensitive-word list stays on disk (`wkuikit/src/main/java/com/chat/uikit/WKUIKitApplication.java:249`)

The add side wrote `wk_sensitive_words` (plaintext JSON of word list + tip text) and `wk_sensitive_words_version` via `MsgModel.syncSensitiveWords`. This PR removes the sync/parse/field (Phase-C no longer calls it), but never clears the keys already written on existing devices — after upgrade nothing reads them, so the plaintext word list stays on disk as dead data indefinitely. The DB rows got a cleanup migration while the SP side got none; the add↔remove parity is asymmetric. Fix direction: one-shot `remove()` of both keys in the startup chain or Phase-C.

## Human-verify

1. The server endpoint `message/sync/sensitivewords` is no longer called by Android. Whether to deprecate it server-side, and whether iOS/web still surface sensitive-word tips (cross-client consistency of this removal), is outside this diff — **not** a merge blocker for this PR.
2. Confirm the transient blank preview in the first P2 is acceptable for affected conversations, or adopt the migration fix-up — **not** a merge blocker either way.

- Residual new-message listener loop after the removal is coherent: `isAlertMsg` / `needsMembershipRefresh` / Space-filter / notification logic unchanged.
- `orderSeq` gaps left by deleted fake messages are harmless for range pagination (`idx_message_channel_order_seq`).
- `lastMsgSeq=0` on affected conversations predates this PR (fake messages had `messageSeq=0`); sync-cursor semantics are unchanged by the migration.
- Fresh installs run all migrations from index 0; the DELETE against an empty `message` table is a no-op.

## Verdict: COMMENT

The feature removal is complete and reference-clean; the migration is correctly wired, index-backed, and runs before any read; every consumer of the removed data path is null-safe. The two P2 items are cleanup parity (SP keys) and a transient, self-healing preview blank — non-blocking. automated review recommendation: approve.

yujiawei:
Re: the remediation suggested in the automated review above —

> `UPDATE conversation SET last_client_msg_no = '' WHERE last_client_msg_no IN (SELECT client_msg_no FROM message WHERE type = -10);`

**Please don't apply that one.** It doesn't fix the symptom, and it introduces a failure mode the current state cannot have.

**It doesn't fix anything.** `MsgDbManager.queryWithClientMsgNo` early-returns on an empty pointer:

```java
// wkim/src/main/java/com/xinbida/wukongim/db/MsgDbManager.java:880-881
public WKMsg queryWithClientMsgNo(String clientMsgNo) {
    if (TextUtils.isEmpty(clientMsgNo)) return null;
```

`getWkMsg()` still resolves to `null`, so the preview is still blank. It converts a dangling pointer into an empty pointer with no user-visible difference.

**It can cross-link conversations to unrelated messages.** The conversation-list batch path does *not* have that empty-string guard:

```java
// wkim/src/main/java/com/xinbida/wukongim/db/MsgDbManager.java:786
try (Cursor cursor = helper.select(message, "client_msg_no in (" + WKCursor.getPlaceholders(clientMsgNos.size()) + ")", ...
```

and the result is matched back by plain string equality:

```java
// wkim/src/main/java/com/xinbida/wukongim/db/ConversationDbManager.java:108
if (uiMsg.clientMsgNo.equals(msg.clientMsgNO)) {
    uiMsg.setWkMsg(msg);
```

Message rows with an empty `client_msg_no` are reachable — `WKMsg()` generates one in its constructor (`WKMsg.java:100`), but the sync path overwrites it with the server value unguarded:

```java
// wkim/src/main/java/com/xinbida/wukongim/manager/MsgManager.java:1223
msg.clientMsgNO = wkSyncRecent.client_msg_no;
```

So `'' == ''` matches, and the affected conversation renders some unrelated message as its preview. A random UUID pointing at a deleted row can never collide with anything; `''` can. That is strictly worse than the bug being fixed.

The repoint statement in my review above (re-target `last_client_msg_no` to the channel's newest remaining row before the `DELETE`) is the direction I'd suggest instead.

---

Two other notes on that review, since they bear on the verdict:

**The "self-heals — hence P2" grading.** The review's own text says the preview is empty *"until its next real message"* — that is an unbounded wait, not a self-heal. For a dormant conversation it is indefinite. The natural rebuttal is "the conversation sync will repair the pointer" — I originally made exactly that argument myself and had to retract it: `octo-server` overrides the client's `version = 0` with a per-device cursor (`modules/message/api_conversation.go:347`, and `messageSaveAcrossDevice` is commented out in `configs/tsdd.yaml:8` so it defaults `false`), which downgrades the cold-start "full sync" to an incremental one. Quiet conversations are never returned, so the pointer is never rewritten.

**The import-cleanup ✅ is checking the wrong direction.** "No remaining *users* of ... fastjson `JSON`" is correct — there are zero. The defect is the import that remains *without* users:

- `wkuikit/src/main/java/com/chat/uikit/message/MsgModel.java:26` — `import com.alibaba.fastjson.JSON;`, 0 usages
- `wkuikit/src/main/java/com/chat/uikit/chat/manager/WKIMUtils.java:85` — `WKTextContent`, now only in the comment at `:518`
- `wkuikit/src/main/java/com/chat/uikit/chat/manager/WKIMUtils.java:87` — `JSONException`, now only in the comment at `:532`

Warnings only, not a blocker — but the sweep isn't finished.

The rest of that review's verification I checked and agree with, including two things I had missed: `type_index` on `message(type)` (`wk_sql/202008292051.sql:27`) making the DELETE index-backed, and `getContent(null)` returning `""` (`ChatConversationAdapter.java:894-896`) confirming the blank preview is a graceful degradation rather than a crash.

yujiawei (approved):
# Code Review — PR #142 (octo-android)

Re-review at head `294289542c4882fc21943c128dcd47321214dbf9`. **This supersedes the earlier `CHANGES_REQUESTED` on `0a26e9be`.** That review's single blocker — the migration deleting the fabricated rows without repairing the conversation pointers they had hijacked — is addressed by commit `294289542`, and I verified the fix by executing the migration rather than only reading it.

---

## 1. Spec compliance

**Spec: ✅**

Issue #141's *Proposal* lists the removal surface; all of it is here and nothing else is.

| #141 asks for | In this diff |
|---|---|
| UI provider | `WKSensitiveWordsProvider.kt` deleted |
| layout | `res/layout/chat_item_sensitive_words_layout.xml` deleted |
| entity | `enity/SensitiveWords.java` deleted |
| network sync | `MsgService.syncSensitiveWords()` + `MsgModel.syncSensitiveWords()` deleted |
| inbound-message filter | `WK_TEXT` branch removed from `WKIMUtils.initIMListener()` |
| startup wiring | `parseLocalSensitiveWords()`, the `sensitiveWords` field, provider registration, Phase-C sync call all removed |
| DB migration for existing rows | `wkim/src/main/assets/wk_sql/202608191100.sql` added |
| keep `-10` reserved | `WKContentType.sensitiveWordsTips = -10` kept, `@Deprecated` with a do-not-reuse rationale (`WKContentType.java:45-57`) |

- **Missing:** none.
- **Extra / out of scope:** none. A repo-wide grep for `sensitiveWords|SensitiveWords|sensitive_words|sensitivewords` across `wkbase` / `wkuikit` / `wkim` / `app` returns exactly one line — the deprecated constant at `WKContentType.java:57`. `ProhibitWord` (separate feature, similar name) is untouched, as claimed.
- **Deviations:** none. The pointer-repair statement added in `294289542` is remediation of the migration's own collateral damage, squarely within "add a DB migration that deletes the rows it already wrote".

---

## 2. Code quality

**Quality: Approved** — no P0, no P1. Four P2s.

### The blocker is genuinely fixed — verified by execution

I rebuilt the schema by replaying every file in `wk_sql/` through the loader's own semantics (concatenate asset lines with **no** separator, split on `;` — `WKDBUpgrade.java:80-92`), loaded fixtures, and ran the migration:

| Fixture | Before | After | Result |
|---|---|---|---|
| A — tip on top of a real message | `A-tip` | `A-real`, seq+timestamp repointed | ✅ repaired |
| C — conversation that never had a tip | `C-real` | `C-real` unchanged | ✅ no collateral damage |
| D — real message newer than the tip | `D-real2` | `D-real2` unchanged | ✅ untouched |
| E — newest remaining row is `type = 99` | `E-tip` | `E-real` (99 correctly skipped) | ✅ filter correct |
| B — tip is the *only* remaining row | `B-tip` | `B-tip`, row deleted | ⚠️ pointer left stale (see P2-1) |

Concretely confirmed along the way:

- **The file parses into exactly 2 statements in the right order.** Each statement is on its own single line, which is mandatory here — the loader glues lines together with no separator, so a wrapped statement would produce invalid SQL.
- **Every column exists.** `last_msg_seq` is not in the base schema (`202008292051.sql:31-40`); it is added by `202204271423.sql:1`. This matters because `WKDBUpgrade` advances the stored index even when a statement throws (`WKDBUpgrade.java:57-64`), so a bad column name would have meant a silent, permanent, unretryable failure. It executes clean.
- **The subquery is idiomatic, not ad-hoc.** `channel_id=? and channel_type=? and is_deleted=0 and type<>0 and type<>99 order by order_seq desc limit 1` is byte-for-byte the app's own `MsgDbManager.queryMaxOrderSeqMsgWithChannel` (`MsgDbManager.java:918`) — the query the client already uses for exactly this "repoint a conversation after its last message went away" job (`ConversationManager.java:224`, `MsgManager.java:428`, `MsgManager.java:479`).
- **The `WHERE EXISTS` guard is precise.** `client_msg_no` carries a `UNIQUE` index (`202008292051.sql:25`) and is a generated ID (`WKMsg.java:100`), so the un-scoped first `EXISTS` cannot cross-match another channel. Fixtures C and D confirm untouched conversations are not rewritten.
- **The `DELETE` cannot destroy server-authored content.** The tip was built locally and persisted through `saveAndUpdateConversationMsg` (`MsgManager.java:691`) — local insert plus conversation-row update, no network send — so `type = -10` rows have no `message_id` and were never uploaded.
- **No unread-count residue.** The removed call passed `addRedDots = false`, and `MsgManager.java:707` maps that to `unreadCount = 0`, so these rows never incremented `unread_count`.
- **Multi-account is safe.** The migration cursor is per-user: `setting.getLong(getUid() + "_db_upgrade_index", 0)` (`WKIMApplication.java:154`).
- **Leftover `-10` rows cannot crash the UI** if the migration ever fails: `ChatAdapter.getItemType()` falls through to `WKContentType.unknown_msg` (`ChatAdapter.java:212`), which is unconditionally registered (`WKMsgItemViewManager.java:49`).
- **The inbound loop is undisturbed.** The removed `WK_TEXT` block was a standalone leaf at the tail of the loop body; the `isAlertMsg` chain, the two suppression guards, and the `needsMembershipRefresh` batch flag are all intact (`WKIMUtils.java:243-310`).
- **Startup ordering is intact.** Phase-B is now `initIM()` + `initIMListener()`; the deleted parse step was the only reason Phase-B had a pre-listener stage, and removing it cannot reintroduce the empty-conversation-list race the surrounding comment describes.

### P2-1 — A tip-only conversation keeps a stale pointer, and the Javadoc overstates the repair

`wkim/src/main/assets/wk_sql/202608191100.sql:1`

```sql
... WHERE EXISTS (... t.type = -10) AND EXISTS (SELECT 1 FROM message m ... AND m.type <> -10);
```

The second `EXISTS` skips any conversation with no other eligible message, but line 2 deletes its `-10` row regardless. Fixture B above reproduces it: `last_client_msg_no` / `last_msg_seq` / `last_msg_timestamp` go on describing a row that no longer exists, and the list preview renders blank.

That contradicts the Javadoc added in this PR, which states the repair without qualification (`WKContentType.java:49-52`): *"先把 last_client_msg_no 还指着这类行的会话重新指向该频道最新一条真实消息，再删行 —— 否则会话列表预览会因为指针悬空而渲染成空白."*

**Why I grade this P2 and not a blocker** — three independent reasons, each verified:

1. **The obvious fix changes nothing the user sees.** Clearing `last_client_msg_no` to `NULL`/`''` routes into `MsgDbManager.queryWithClientMsgNo`, whose first line is `if (TextUtils.isEmpty(clientMsgNo)) return null;` (`MsgDbManager.java:881`). A stale pointer already resolves to `null` via the same method. Both paths produce the identical blank preview. Only deleting the conversation row would look different, and silently dropping a conversation from the list is a worse trade than a blank preview.
2. **It matches the codebase's existing convention.** `MsgManager.deleteWithClientMsgNO` re-points only `if (tempMsg != null)` (`MsgManager.java:473-486`), and `ConversationManager.updateWithMsg` does the same (`ConversationManager.java:224-228`). Delete the last message of any conversation today and you get this exact state. This PR does not introduce a new defect class.
3. **It self-heals on the next message.** `ConversationDbManager.insertOrUpdateWithMsg` overwrites `lastClientMsgNO` unconditionally (`ConversationDbManager.java:420`), so the first inbound message repairs it.

Reachability is also narrow: the tip is only ever created on receipt of a real `WK_TEXT`, so this requires every real message in that channel to have been deleted while the tip survived. And for such a channel, "blank" is arguably the honest rendering — there is genuinely nothing left to preview.

**Suggested (non-blocking):** either extend the migration to also null the three fields when no replacement exists, or — cheaper and equally honest — soften the Javadoc at `WKContentType.java:49-52` to say the repair applies to channels that still have a real message. Please don't ship the current unqualified wording.

### P2-2 — The pointer repair and the delete are not atomic, and this is the first data migration in `wk_sql/`

`WKDBUpgrade.java:50-65` runs each statement in its own `try`/`catch`; a failure logs, sets `success = false`, and the loop **continues to the next statement**, after which `setDBUpgradeIndex(dbSql.index)` runs unconditionally. So if the `UPDATE` throws, the `DELETE` still executes and the migration is permanently recorded as done — every affected conversation left dangling with no retry.

This is pre-existing infrastructure and **I am not asking for it to be changed in this PR** — every prior file in `wk_sql/` is idempotent DDL (`ALTER TABLE` / `CREATE INDEX IF NOT EXISTS`) where advance-on-failure is the right call. But this is the directory's first *data-mutating* migration, so it is the first time that flaw has real blast radius, and it is worth someone's attention as a follow-up. The statement itself executes clean against the real schema, so the realistic trigger is an environmental failure (locked DB, disk full, corruption).

Related: `setDBUpgradeIndex` writes to SharedPreferences via `editor.apply()` (`WKIMApplication.java:157-163`), outside any DB transaction — so index bookkeeping and DB state can diverge.

### P2-3 — Three imports orphaned by this removal are still present

Carried over from the previous round, unaddressed at this head. Each has exactly one remaining non-comment occurrence: the import line itself.

- `wkuikit/src/main/java/com/chat/uikit/chat/manager/WKIMUtils.java:85` — `com.xinbida.wukongim.msgmodel.WKTextContent` (only other occurrence is the commented-out line 518)
- `wkuikit/src/main/java/com/chat/uikit/chat/manager/WKIMUtils.java:87` — `org.json.JSONException` (only other occurrence is the commented-out line 532)
- `wkuikit/src/main/java/com/chat/uikit/message/MsgModel.java:26` — `com.alibaba.fastjson.JSON` (its sole use was `JSON.toJSONString` inside the deleted `syncSensitiveWords`; `JSONArray`/`JSONObject` on lines 27-28 are still live)

Javac warnings only. Worth finishing since the diff already correctly dropped `WKSendMsgResult` and `java.util.Objects` from the same file, and correctly kept `Handler`/`Looper`/`TextUtils`/`Log`/`WKSharedPreferencesUtil`, which are still used — the sweep was clearly meant to be complete.

Also vestigial: `assert msgList != null;` (`WKIMUtils.java:366`) immediately preceded the removed delayed-save block and is now the last statement in the lambda, guarding nothing — `msgList` is dereferenced ~120 lines earlier at line 243.

*(The duplicate `import org.json.JSONObject;` at `WKIMUtils.java:83` and `:88` is pre-existing, not from this PR.)*

### P2-4 — Retired SharedPreferences keys are never cleared

The deleted `syncSensitiveWords` wrote `wk_sensitive_words` and `wk_sensitive_words_version`; at this head both have zero readers and zero writers, so every upgrading install keeps the downloaded word list on disk forever. `WKSharedPreferencesUtil` normally uses `EncryptedSharedPreferences` (AES256-GCM), which bounds the exposure — but it has a plaintext `MODE_PRIVATE` fallback path, so on any device that took the fallback the list sits there in cleartext. A one-time removal of both keys, guarded by its own flag, would complete the reclamation the `## What` table otherwise documents in full.

---

## 3. Overall verdict

**APPROVE**

Spec ✅ and no P0/P1. The previous blocker is properly fixed: the repair statement is correctly ordered before the delete, uses the app's own canonical latest-message query, is precise enough not to touch conversations that never had a tip, and executes cleanly against the real replayed schema. The four P2s are follow-ups, and P2-1's Javadoc wording is the one I'd most like tightened before merge.

---

## 4. Suggested follow-ups

1. Tighten the `WKContentType.java:49-52` Javadoc, or extend the migration to the tip-only case (P2-1).
2. Drop the three orphaned imports and the vestigial `assert` (P2-3).
3. Add the one-time `wk_sensitive_words` / `wk_sensitive_words_version` cleanup (P2-4).
4. **The manual test checklist in the PR description is entirely unchecked.** This PR deletes rows and rewrites conversation metadata on upgrade — please run it against a device that actually has `type = -10` rows and tick the boxes. Worth adding one case the checklist misses: a conversation whose only remaining message is the tip.

---

## 5. What still needs human eyes

This PR carries `needs-human-review`. None of the following are code defects and none changed my verdict:

1. **Compliance sign-off.** The tip enforced nothing — it fired *after* delivery, blocked no send or receipt, and had no server-side counterpart — so no technical control is lost. But "a visible content warning existed and now does not" is a product and regulatory question. Whoever owns that call should confirm no obligation depended on this bubble being on screen.
2. **Server-side and other clients.** `GET message/sync/sensitivewords` is intentionally left live. Worth confirming (a) who else still consumes it and when it can be retired, and (b) whether iOS/Web still fabricate the same tip — if they do, issue #141's complaint #4 ("the two transcripts disagree") only changes direction rather than being resolved.
3. **Migration against production-shaped data.** My verification ran against a schema replayed from `wk_sql/` with synthetic fixtures, not a real populated device database. The upgrade path deserves one real-device run, ideally including the offline-first-launch case.

---

## 6. Coverage note

**Done:** full checkout at this head; repo-wide symbol grep across `wkbase`/`wkuikit`/`wkim`/`app`; schema and index verification including the `last_msg_seq` ALTER; execution of the migration through the loader's exact concat/split semantics with five fixture conversations; loader index-advance and transaction-boundary analysis; conversation-hydration and preview-resolution path; multi-account migration keying; unread-count residue; unregistered-message-type render fallback; inbound-listener control-flow integrity; per-symbol import liveness in both modified Java files.

**Not done:** no Gradle build or test run locally (the repo's own `build` check is green at this SHA and is the only compile evidence here); no on-device upgrade against a real populated database; no exercise of the offline-then-online startup sequence; no inspection of `octo-server` for other consumers of the sensitive-words endpoint; no iOS/Web parity check. Also not assessed: whether any deployment sets `messageSaveAcrossDevice`, which the previous round's analysis turned on — it is moot for the main case now that the migration repairs the pointer locally, but it still governs how quickly the P2-1 edge case would be papered over by a server sync.

Jerry-Xin (approved):
Re-review at head 29428954 (supersedes my REQUEST_CHANGES at 0a26e9be, review 5027494748). The previous blocker is fixed and byte-verified; approving.

## Previous blocker — FIXED

- **Dangling `conversation.last_client_msg_no` after hint-row deletion** (my 🔴 at 0a26e9be): the migration now re-points `last_client_msg_no` / `last_msg_seq` / `last_msg_timestamp` of every conversation whose pointer still targets a `type = -10` row to the channel's newest remaining real message, *before* deleting the rows. Verified end-to-end:
  - **UPDATE-before-DELETE order** is correct in the file and preserved by `WKDBUpgrade.getExecSQL()` (lines joined, split on `;`, executed in order); each statement is one physical line with no embedded semicolons/comments, matching the loader's constraints.
  - **Repoint filter** (`is_deleted = 0 AND type <> 0 AND type <> 99 AND type <> -10 ORDER BY order_seq DESC LIMIT 1`) is the same primitive the runtime already uses for post-deletion pointer repair (`MsgDbManager.queryMaxOrderSeqMsgWithChannel`, wkim/src/main/java/com/xinbida/wukongim/db/MsgDbManager.java:918, consumed by `MsgManager.deleteWithClientMsgNO`). All three columns are selected from the same row, so pointer/seq/timestamp stay mutually consistent.
  - **Scope**: the double EXISTS restricts the UPDATE to conversations whose pointer actually sits on a hint row AND that have an eligible replacement. I replayed the exact runner semantics against the production schema in SQLite: an affected conversation re-points to the newest real message (correctly skipping newer type=0 / type=99 / soft-deleted rows), unaffected conversations stay byte-identical, and non-hint message rows are untouched. `ConversationPreviewSelector` (wkbase/src/main/java/com/chat/base/space/ConversationPreviewSelector.java) therefore renders the real last message for regular DM/group previews again — no blank preview, no crash.
  - **Single-run / idempotent**: `202608191100` remains the highest `wk_sql/` index and is gated by the stored upgrade index in `WKDBUpgrade.onUpgrade`; even if replayed it no-ops (once hint rows are gone, the first EXISTS can never match).

## Edge case — conversation whose only eligible row was a hint: deliberate skip, non-blocking

The second EXISTS guard intentionally leaves the pointer unchanged (never NULL) when no real message remains, then the DELETE removes the hint. Result: preview stays blank until the next message arrives. This is acceptable because: (a) it is byte-identical to the SDK's existing behavior for any conversation whose messages are all deleted — `MsgManager.deleteWithClientMsgNO` likewise skips re-pointing when `queryMaxOrderSeqMsgWithChannel` returns null; (b) no crash path exists — `WKUIConversationMsg.getWkMsg()` returns null for the stale pointer and all consumers null-guard; (c) reachability is negligible — the removed fabrication code only ever created a tip alongside a triggering inbound WK_TEXT message (real sibling always saved with `orderSeq + 1`), so this branch requires every triggering real message to be separately deleted while the tip survived; (d) the choice is documented in the commit message ("频道内没有可指的真实消息时跳过，不把指针写成 NULL"). Clearing the pointer to NULL/0 or deleting the conversation row would diverge from existing semantics and produce worse outcomes (conversation sinking to the bottom / losing mute-top-unread state) without fixing anything the SDK doesn't already exhibit.

## 💬 Non-blocking

- The three orphaned imports flagged at 0a26e9be are still present: `org.json.JSONException` and `com.xinbida.wukongim.msgmodel.WKTextContent` in wkuikit/src/main/java/com/chat/uikit/chat/manager/WKIMUtils.java (referenced only from comments at lines 518/532) and `com.alibaba.fastjson.JSON` in wkuikit/src/main/java/com/chat/uikit/message/MsgModel.java. No compile impact; worth finishing the sweep at some point.
- SP keys `wk_sensitive_words` / `wk_sensitive_words_version` remain orphaned on device — unchanged from last round, still harmless.
- Unit tests for the migration (normal re-point + skip branch + unaffected conversation) would lock in the verified behavior; currently covered by review + simulation only.

## ✅ Highlights

- yujiwei's sync concern is resolved by design: the repair is fully local and deterministic at DB-upgrade time, so no pointer depends on server full-sync self-healing (which the server does not support).
- Repoint semantics deliberately mirror the SDK's existing max-order-seq query, and the commit message documents both the statement-ordering rationale and the skip-branch choice.
- The rest of the deletion refactor remains byte-clean from the previous round: provider / model / API binding / layout / registration / inbound fabrication all removed; `-10` retained as a deprecated, do-not-reuse constant.

mochashanyao (approved):
[Octo-Q · automated review]

---

Head: `294289542c4882fc21943c128dcd47321214dbf9` · merge-base: `3133fe30` · range: `pr-base...HEAD` · 9 files, +18/−232 · round 2 (re-review after new commit)

# Code Review — PR #142 (octo-android)

## Summary

Second-round review. This PR removes the client-fabricated sensitive-word tip feature end to end: the `WKIMUtils` inbound-listener block that matched `WK_TEXT` against a locally synced word list and injected a fake `type=-10` message, plus its provider/layout/entity, the `message/sync/sensitivewords` call, SP parse, and the registration in `WKUIKitApplication`; the `sensitiveWordsTips = -10` constant is deliberately kept `@Deprecated` (protocol number must not be reused). The new round-2 commit `29428954` addresses the prior round's migration finding: `wk_sql/202608191100.sql` now redirects conversation pointers away from doomed `-10` rows before deleting them. The fix is correct and the removal remains complete; no new defects found.

## Verification

Static analysis only at head `29428954`; build and tests not executed in this environment.

- ✅ **Migration executes multi-statement file correctly** — `WKDBUpgrade.getExecSQL()` concatenates asset lines with no separator, then splits on `;` and runs each part via `db.execSQL` (`wkim/src/main/java/com/xinbida/wukongim/db/WKDBUpgrade.java:75`). Byte-level check of `wk_sql/202608191100.sql`: line 1 (1104 chars) ends with `;`, line 2 is `DELETE FROM message WHERE type = -10;`, exactly one `;` per statement, file ends `\n` — no token-merge corruption, no embedded semicolons.
- ✅ **UPDATE runs before DELETE and only where needed** — guarded by `WHERE EXISTS (… last_client_msg_no → type=-10 row) AND EXISTS (… real message in channel)`; subquery picks latest message by `order_seq` excluding deleted/type 0/99/−10, matching what the conversation preview should show.
- ✅ **All referenced columns and indexes exist** — `conversation(last_client_msg_no, last_msg_seq, last_msg_timestamp, channel_id, channel_type)` and `message(client_msg_no, channel_id, channel_type, is_deleted, type, order_seq, message_seq, timestamp)`; `msg_client_msg_no_index`, `msg_channel_index`, `type_index` cover the subqueries (`wk_sql/202008292051.sql:24-27`).
- ✅ **Migration wiring at runtime** — `WKDBHelper` ctor runs `WKDBUpgrade.onUpgrade(mDb)` on every per-user DB open before any read (`wkim/src/main/java/com/xinbida/wukongim/db/WKDBHelper.java:147`); `202608191100` is the highest index in `wk_sql/`, so nothing shadows it.
- ✅ **Removal is complete** — repo-wide grep finds zero remaining references to `WKSensitiveWordsProvider`, `SensitiveWords`, `syncSensitiveWords`, `wk_sensitive_words`, `chat_item_sensitive_words_layout`, or `sensitiveWords` outside the deprecated constant and its javadoc.
- ✅ **Fallbacks hold if migration partially fails** — runner advances the index per-file even on per-statement error; surviving `-10` rows render as `unknown_msg(-3)` via `ChatAdapter.getItemType` (`wkbase/src/main/java/com/chat/base/msg/ChatAdapter.java:195`), and dangling conversation pointers render blank via `getWkMsg()`→null→`getContent(null)`→`""` (`wkuikit/src/main/java/com/chat/uikit/chat/adapter/ChatConversationAdapter.java:894`, `wkim/src/main/java/com/xinbida/wukongim/entity/WKUIConversationMsg.java:44`). Both are no-crash, self-healing on next real message.

## Findings

No P0/P1 issues; one carried-over P2 and one carried-over Nit below.

### P2 — Orphaned SP keys `wk_sensitive_words` / `wk_sensitive_words_version` never cleaned (`wkuikit/src/main/java/com/chat/uikit/WKUIKitApplication.java:249`)

All readers/writers of these SharedPreferences keys were removed, but existing installs keep the plaintext sensitive-word list JSON forever — an add↔remove parity gap for persisted state. Diff-scope: pre-existing data, newly orphaned by this PR (the removal creates the stale state), non-blocking. Fix direction: one-shot `removeSP` for both keys (guarded by a one-time flag), e.g. in the `postPhaseC("post-wkim-idle")` block next to `deleteFlameMsg()`. Carried over from round 1, still open.

### Nit — Unused import in `WKIMUtils.java` (`wkuikit/src/main/java/com/chat/uikit/chat/manager/WKIMUtils.java:85`)

`import com.xinbida.wukongim.msgmodel.WKTextContent;` is unused after the removal; the only remaining reference is commented-out code at `:518`. Carried over from round 1.

## Human-verify

1. Removing the client-side sensitive-word tip is a product/policy decision: server-side moderation behavior is untouched by this diff (Android only *displayed* a local tip), but confirm with the product owner that dropping the visible warning is intended. **Not a merge blocker for this PR.**
2. The server endpoint `message/sync/sensitivewords` loses its Android caller; confirm iOS/web clients no longer need it before the server side is retired, if retirement is planned. **Not a merge blocker for this PR.**

- `last_msg_seq` cannot regress: conversations matched by the UPDATE have `last_msg_seq=0` (fake tips were saved with `messageSeq=0`, `WKMsg.java:96` ctor default + auto `clientMsgNO` at `:100`), so the redirect only advances the sync cursor; re-fetch of already-present messages is idempotent upsert.
- `unread_count` untouched by the migration — tips were saved with `addRedDots=false`, so no unread state ever existed to clean.
- Migration target correctness: a tip was only ever created immediately after a `WK_TEXT` message with `orderSeq=real+1`, so for matched conversations the redirect target is exactly the real message the tip followed (or a newer real message), never a type-0/99 row in practice.
- Compile safety of the excision: removed imports (`JSON`, `WKSendMsgResult`, `Objects`, `SensitiveWords`) were the only consumers; `TextUtils`/`WKReader`/`WKChannel`/`Handler`/`Looper`/`AtomicLong` still have other users; no tests reference the removed classes.
- Search state: message search is a direct `LIKE` over `message.searchable_word` (`MsgDbManager.java:1164,1199`), so deleting the rows removes them from search — no separate index table to clean.
- UPDATE+DELETE not wrapped in a transaction is a pre-existing property of the runner; both partial-failure orders land in acceptable, self-healing states (see fallbacks above).

## Data-flow trace (consumed data → upstream source → does it actually arrive?)

| Consumed data | Upstream source | Arrives at runtime? |
|---|---|---|
| `conversation.last_client_msg_no` matched by UPDATE | fake tip's auto-generated `clientMsgNO` written via `saveAndUpdateConversationMsg(msg,false)`→`insertOrUpdateWithMsg` (`ConversationDbManager.java:406`) | Yes — tip rows had real unique `client_msg_no`; conversations whose tip was superseded by a newer real msg don't match (correct) |
| UPDATE subquery result rows | `message` rows with `is_deleted=0`, type∉{0,99,−10} in same channel | Yes for any channel that had a real message; empty-channel case excluded by second EXISTS → falls back to round-1 blank-preview behavior (null-safe) |
| migration file content | asset `wk_sql/202608191100.sql` read by `getExecSQL()` | Yes — parsed into exactly 2 statements (verified at byte level) |
| `last_msg_seq` consumers (`syncConversationChat`) | conversation table after UPDATE | Consistent — value only advances; sync cursor semantics preserved |

## Blind-spot checklist (security_sensitive — all items)

- **C1 dual-path parity** — hit (P2): removal audited pair-by-pair (register↔unregister done; sync-call↔endpoint-reader removed; SP-write↔SP-read removed; row-producer↔row-cleanup done), but SP-key cleanup is the one missing remove-leg → P2 above.
- **C2 control-flow ordering / nested reuse** — clear: removed listener block had a single call site; migration statement order (UPDATE→DELETE) is required and correct; runner's line-concatenation quirk tested against this exact file (no non-canonical form slips through: no embedded `;`, each statement line-terminated).
- **C3 authorization boundary** — N/A: no new endpoint/tool/permission exposed; change removes a client-local display feature; no auth surface change.
- **C4 authorization lifecycle / container-member cascade** — N/A: no auth/membership-status logic touched.
- **C5 build ≠ runtime path** — clear: runtime path proven, not assumed — `WKDBHelper` ctor → `onUpgrade` → per-statement `execSQL` on every per-user DB open (`WKDBHelper.java:147`); partial-failure states analyzed.
- **C6 governance/docs self-consistency** — clear: no repo docs/string resources reference the sensitive-word feature (grep over `*.md` and res values returned nothing); the `@Deprecated` javadoc on `sensitiveWordsTips` correctly documents the protocol-number freeze rationale.

## Cross-round blocker recheck (round 1: [automated review](mention://issue/16715c1d-d96f-491b-9ae9-a50156c10f7c), head `0a26e9be`, verdict APPROVED)

Round 1 had no P0/P1 blockers. Carried items:

1. **P2 — dangling `conversation.last_client_msg_no` after DELETE** → **已修 (fixed)**: commit `29428954` adds the pointer-redirect UPDATE before the DELETE in the same migration file; fix verified against runner parsing, schema, indexes, and guard semantics (evidence in Verification above).
2. **P2 — orphaned SP keys** → **未修 (still open)**: zero remaining readers/writers, but no erase added; carried forward as P2.
3. **Nit — unused `WKTextContent` import** → **未修 (still open)**: `WKIMUtils.java:85`.

## Verdict: APPROVED

The round-2 commit correctly resolves the only substantive migration concern from round 1; the removal is complete, compile-safe, and every residual state (including partial-migration failure) degrades to a null-safe, self-healing UI. Remaining items are one P2 cleanup and one nit, both carried over and non-blocking.

[Octo-Q] verdict: APPROVE — 无 P0/P1；上轮唯一实质问题（迁移后指针悬空）已在 29428954 正确修复并经运行期路径验证；遗留 P2（SP 键未擦除）与 Nit（未用 import）不阻塞。

